PR #11, opened 2015-09-22
Fix link to FOSUserBundle's User class doc

kix:
FOSUserBundle doc uses ReStructuredText instead of Markdown. The link to the User entity docs was broken because of this (used `.md` instead).


theofidry:
Good catch!

PR #17, opened 2015-10-06
Add info on using dynamic validation groups with API bundle

teohhanhui:
Requires dunglas/DunglasApiBundle#254


dunglas:
Thank you @teohhanhui

PR #21, opened 2015-10-31
Add doc to custom normalizer

yelmontaser:
Reference: https://github.com/dunglas/DunglasApiBundle/issues/316

### DIFF
--- a/api-bundle/serialization-groups-and-relations.md
+++ b/api-bundle/serialization-groups-and-relations.md
@@ -237,8 +237,8 @@ class CustomItemNormalizer implements NormalizerInterface, DenormalizerInterface
      */
     public function __construct(NormalizerInterface $normalizer)
     {
-        if (!$normalizer instanceof DenormalizerInterface) {
-            throw new \InvalidArgumentException('The normalizer must implement the DenormalizerInterface');
+        if (!$normalizer instanceof NormalizerInterface) {
+            throw new \InvalidArgumentException('The normalizer must implement the NormalizerInterface');
         }
 
         $this->normalizer = $normalizer;
--- a/api-bundle/serialization-groups-and-relations.md
+++ b/api-bundle/serialization-groups-and-relations.md
@@ -198,6 +198,101 @@ the data provider and any changes in the embedded relation will be applied to th
 
 You can create as relation embedding levels as you want.
 
+### Creating a custom Normalizer
+The normalizer must implements the ```NormalizerInterface``` and ```DenormalizerInterface```.
+If you add simple modification eg: add a new serialization group for each item, you can inject the Normalizer provided with the API.
+
+Define the service:
+```yml
+api.serializer.normalizer.item:
+    public: false
+    class: AppBundle\Serializer\CustomItemNormalizer
+    arguments: [ @api.json_ld.normalizer.item ]
+    tags: [ { name: serializer.normalizer } ]
+```
+
+The custom normalizer:
+
+```php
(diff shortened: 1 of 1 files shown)


api-bundle/serialization-groups-and-relations.md
```diff
@@ -198,6 +198,101 @@ the data provider and any changes in the embedded relation will be applied to th
...
+    private $normalizer;
+
+    /**
+     * CustomItemNormalizer constructor.
+     *
+     * @param $normalizer
+     */
+    public function __construct(NormalizerInterface $normalizer)
+    {
+        if (!$normalizer instanceof DenormalizerInterface) {
```
dupuchba: shouldn't it be `NormalizerInterface`?

yelmontaser: Well seen, since normalize and denormalize can check both.

api-bundle/serialization-groups-and-relations.md
```diff
@@ -198,6 +198,101 @@ the data provider and any changes in the embedded relation will be applied to th
 
 You can create as relation embedding levels as you want.
 
+### Creating a custom Normalizer
+The normalizer must implements the ```NormalizerInterface``` and ```DenormalizerInterface```.
```
teohhanhui: "must implements" -> "must implement"

Also, there's an `AbstractNormalizer` which makes things easier...

api-bundle/serialization-groups-and-relations.md
```diff
@@ -198,6 +198,101 @@ the data provider and any changes in the embedded relation will be applied to th
...
+
+        $this->normalizer = $normalizer;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function denormalize($data, $class, $format = null, array $context = array())
+    {
+        return $this->normalizer->denormalize($data, $class, $format, $context);
```
teohhanhui: You should implement `SerializerAwareInterface` and call `$this->serializer->denormalize(...)` instead. Using a hardcoded normalizer is a bad idea (it bypasses the serializer's mechanism for determining the normalizer to use.

p/s: Which is also why the easiest way is just to extend `AbstractNormalizer`.

teohhanhui:
It's unclear what you're trying to achieve here...

dunglas:
@yelmontaser any update on this PR?

yelmontaser:
@teohhanhui as shown in the following post https://github.com/dunglas/DunglasApiBundle/issues/316, the Dunglas API does not permit to specify a serialization group in some contexts, such as having diffirents following groups that is in the resource or in the collection or simply when a user is authenticated etc.

The problem of the method ```initNormalizationContext``` is that it applies the group in general context of the resource and collection.
The objective is not to create a new serializer but to simply add a class that helps control the context of application groups.

Have you another good easy way to do it?

yelmontaser:
I understand that it is a bit ambiguous, but I think that the proposal @dunglas is cleaner than doing it through a controller or event.

sroze:
@yelmontaser so we should close this PR?

yelmontaser:
@sroze close if not necessary

dunglas:
I'm 👍 to merge this PR when comments of @teohhanhui will be fixed. It would be nice to also add a link to the related Symfony doc.

Simperfit:
@yelmonster, Do you have some time to finish this PR ?

yelmontaser:
@Simperfit sorry I do not have time right now and the API-Platform has changed significantly since.

PR #25, opened 2015-11-24
Error when following tutorial

totaltrash:
I encountered an error while following the tutorial. I'd attempt to generate-types for the provided schema, but would get this error:

[error] The type "SocialMediaPosting" (parent of "BlogPosting") doesn't exist

As a workaround, I added the SocialMediaPosting to the types and no error, but i'm not sure if any properties need to be exposed?


dunglas:
Good catch. The Schema.org vocab has changed to introduce this new type. It's why there is an error.
Thanks @totaltrash

PR #29, opened 2016-01-07
Escape backslashes within double quotes

phc:


dunglas:
Thank you @phc

teohhanhui:
@dunglas I thought we decided to go with single quotes all the way?

dunglas:
@teohhanhui yes but a fixed doc is better than a broken one. I'll convert it to simple quotes ASAP.

PR #32, opened 2016-01-25
Add selflink restangular support

mRoca:
This PR adds the [selfLink](https://github.com/mgonto/restangular#using-self-reference-resources) Restangular option in order to do : 

```javascript
Restangular.all('/images').getOne(...).then(image => {
    image.title = 'The new title';
    image.save();
});
```


sroze:
Good catch. Thank you!

PR #33, opened 2016-01-27
Fix schema-generator getting-started

toofff:
Obligation to put the path to the configuration file.


sroze:
Well... that's just depends where you put your `address-book.yml` file (that I wouldn't put in `app/config` btw)

PR #38, opened 2016-03-12
Filters V2

Simperfit:
In the test I was doing to make the `BooleanFilter` work using api-platform v2 and my core, I decided to updated the documentation on filters.

Didn't know if I needed to base from master or V1.

This is what I've done in my entity to make the filter work.


dunglas:
Thank you very much @Simperfit

Simperfit:
NP @dunglas

PR #41, opened 2016-04-07
Remove impolite language

teohhanhui:


theofidry:
`dead simple`?

teohhanhui:
http://english.stackexchange.com/questions/51410/dead-simple-vs-really-simple

theofidry:
Yeah it's not formal either, but I don't feel like our target audience would feel offended by it. But if it's bothering you let's change it.

dunglas:
Thank you @teohhanhui

PR #45, opened 2016-05-04
Remove invalid link

aledeg:
The link returns a 404. I tried to find the page on the blog to find the new link but couldn't find it.
I assume that the post was deleted.


dunglas:
Thanks for the report @aledeg. I ask to the Les-Tilleuls.coop webmastering team to put the link back online.

dunglas:
Thanks @aledeg

PR #52, opened 2016-06-07
Fix and improve the doc of filters

dunglas:


core/filters.md
```diff
@@ -27,42 +27,32 @@ Prepend the letter `i` to the filter if you want it to be case insensitive. For
...
 ```yaml
-
 # app/config/services.yml
 
 services:
     offer.search_filter:
-        parent:    "api_platform.doctrine.orm.search_filter"
-        arguments: [ { id: "exact", price: "exact", name: "partial"  } ]
+        parent:    'api_platform.doctrine.orm.search_filter'
+        arguments: [ { id: 'exact', price: 'exact', name: 'partial' } ]
```
teohhanhui: What do you think about a CS rule which forbids inlining of maps and lists? Or at least for service `arguments`...

teohhanhui:
:+1:

core/filters.md
soyuka: Hmm, I'd be +1 for an injection list, however we're here declaring a single argument which are the filter parameters. IMO it's fine like this and it's easy to read.

teohhanhui: I do not consider this easy to read :smile:

This is what it looks like for me:

```yaml
app.resource.aggregate_offer.search_filter:
    parent: 'api_platform.doctrine.orm.search_filter'
    arguments:
        - 'availability': 'exact'
          'itemOffered': 'exact'
          'offers': 'exact'
          'offers.priceSpecifications.priceType': 'exact'
          'seller': 'exact'
    tags:
        - name: 'api_platform.filter'
          id: 'aggregate_offer.search'
```

dunglas: @teohhanhui I really dislike this notation (I always fail to write it without reading the doc, it looks weir to me). I feel uncomfortable with this dash then indentation. But it's probably a matter of taste.

For short lines like this one, I prefer to keep the current syntax (your syntax can be useful when the line grows up).

soyuka: I don't find this notation clear enough and also more error-prone, newbies might add `-` everywhere and then it'll be broken \o/. I'd do this if I had to avoid inline arguments:

```yml
app.resource.aggregate_offer.search_filter:
    parent: 'api_platform.doctrine.orm.search_filter'
    arguments:
        - { id: 'exact', price: 'exact', name: 'partial' }
```

Also, I think that symfony docs is using the inline notation if there is only one argument.

teohhanhui: > I feel uncomfortable with the indentation after the dash.

Same here. Hahaha... But there's always a compromise.

teohhanhui: Anyway, can we quote the keys too? e.g. `'id'`

For me the rule of thumb is to quote it if it's an arbitrary user-defined key.

dunglas: I prefer to quote keys only when the contain special characters because it makes documents closer to JSON (and as YAML is a superset of JSON, all JSON docs are YAML).
Symfony use this rule too: always quote values, but quote keys only if they contain a special char like `%` or `@`.

theofidry: keys are strings, so unless there special characters why quoting them? Most of Sf doc doesn't quote YAML keys and it looks weird to me to quote `id`, `availabilities` and so and not quoting `name` for tag. I get that you kind of see them differently, but I fear it will only make it more confusing

theofidry:
👍

PR #65, opened 2016-06-19
Improve resource-path-generators.md

dunglas:


core/resource-path-generators.md
```diff
@@ -1,24 +1,26 @@
 # Resource path generators
```
teohhanhui: From the perspective of a reader I don't think the chapter title is very relatable...

core/resource-path-generators.md
```diff
@@ -1,24 +1,26 @@
 # Resource path generators
 
-With API Platform, you can configure the path of your exposed resources. 
+With API Platform, you can configure globally the path of your exposed resources.
```
teohhanhui: When I read this sentence, what comes to mind is the root prefix (e.g. `/api`) :laughing:

core/resource-path-generators.md
dunglas: WDYT of my last commit?

teohhanhui: Better :+1:

dunglas: Maybe should we also rename those services in the code?

teohhanhui: It was @sroze's idea: https://github.com/api-platform/core/pull/547#discussion_r64135705

core/path-naming-strategy.md
```diff
@@ -1,38 +1,37 @@
-# Resource path generators
+# Path naming strategy
 
-With API Platform, you can configure the path of your exposed resources. 
+With API Platform, you can configure globally the naming strategy of URL paths exposing your resources.
```
soyuka: `globally` should be at the end of the sentence

PR #68, opened 2016-06-20
Update security.md

dunglas:


core/security.md
```diff
@@ -1,11 +1,12 @@
 # Security
 
-To completely disable some operations from your application, refers to [Disabling operations](operations.md#disabling-operations).
+To completely disable some operations from your application, refers to the [disabling operations](operations.md#disabling-operations)
```
soyuka: refer to

teohhanhui: We should also be consistent with the casing. There was another place where "Getting started" was used.

soyuka: Indeed, we can maybe review the whole docs once published to see leftovers. IMO, if it's not a project name we should keep it lowercased (as it has been done in this commit).

PR #72, opened 2016-06-21
Add pagination.md. Rework the TOC.

dunglas:


core/pagination.md
```diff
@@ -0,0 +1,198 @@
+# Pagination
+
+API Platform Core has native support for paged collections. Pagination is enabled by default for all collections. Each collections
+contains 30 items per page.
+The activation of the pagination and the number of elements per page can be configured server-side (globally and per resource)
+and client-side (disabled by default).
+
+When issuing a `GET` request on a collection containing more than 1 page (here `/books`), a [Hydra collection](http://www.hydra-cg.com/spec/latest/core/#collections)
+is returned. It's a perfectly valid JSON(-LD) document containing items of the requested page and metadata.
```
soyuka: perfectly isn't really needed

core/pagination.md
```diff
@@ -0,0 +1,198 @@
+# Pagination
+
+API Platform Core has native support for paged collections. Pagination is enabled by default for all collections. Each collections
+contains 30 items per page.
+The activation of the pagination and the number of elements per page can be configured server-side (globally and per resource)
+and client-side (disabled by default).
```
soyuka: can be configured from:
- the server-side (globally or per resource)
- the client-side, via a GET parameter (disabled by default)

core/pagination.md
```diff
@@ -0,0 +1,198 @@
...
+    "@type": "hydra:PartialCollectionView",
+    "hydra:first": "/books?page=1",
+    "hydra:last": "/books?page=2",
+    "hydra:next": "/books?page=2"
+  }
+}
+```
+
+Hypermedia links to the first, the last, previous and the next page in the collection are displayed as well as the number
+items in the collection.
```
soyuka: as the number of total items in the collection.

core/pagination.md
```diff
@@ -0,0 +1,198 @@
...
+# app/config/config.yml
+
+api_platform:
+    collection:
+        page_parameter_name: _page
+```
+
+## Disabling the pagination
+
+Paginating collection is generally accepted as a good practice. It allows to browse large collections and prevent [DOS and
```
soyuka: > Paginating collection is generally accepted as a good practice. 

What are you trying to say? I find this confusing.

It also allows browsing large collections without to much overhead as well as preventing DDOS attacks.

core/pagination.md
```diff
@@ -0,0 +1,198 @@
...
+{
+   // ...
+}
+```
+
+### Client-side
+
+#### Globally
+
+You can configure API Platform Core to let the client enabling or disabling the pagination. To activate this feature globally,
```
soyuka: You can configure API Platform Core to let the client enable or disable the pagination.

core/pagination.md
```diff
@@ -0,0 +1,198 @@
...
+            enabled_parameter_name: pagination # optional
+```
+
+The pagination can now be enabled or disabled by adding a query parameter named `pagination`:
+
+* `GET /books?pagination=false`: disabled
+* `GET /books?pagination=true`: enabled
+
+Any value accepted by the [`FILTER_VALIDATE_BOOLEAN`](http://php.net/manual/en/filter.filters.validate.php) filter can be
+used as value.
```
soyuka: as the value 

(not sure about this but I think it sounds better 😋 )

core/pagination.md
```diff
@@ -0,0 +1,198 @@
...
+
+#### For a specific resource
+
+The ability of the client to disable the pagination can also be set in the resource configuration:
```
soyuka: The client ability to disable the pagination can...

core/pagination.md
```diff
@@ -0,0 +1,198 @@
...
+ */
+class Book
+{
+   // ...
+}
+```
+
+## Changing the number of items per page
+
+Similarly, the number of items per page is configurable and can be set client-side.
```
soyuka: In the same manner,

core/pagination.md
Simperfit: I understand as is, that it's better to paginate a collection to prevent any attack but if you want to, you can disable it.

core/pagination.md
Simperfit: used as a value ?

soyuka: we're talking about the value (true/false) of the pagination, not any value.

core/pagination.md
dunglas: Exactly.

dunglas:
Thank you for the review @soyuka!

PR #75, opened 2016-06-22
Incorrect call to operation reference route

AlexandreHagen:
After testing in my app I fount that:
```"special"={"special"={"route_name"="book_special"}``` Do not work
```"special"={"route_name"="book_special"}``` Work


teohhanhui:
Thanks @AlexandreHagen

PR #76, opened 2016-06-26
Update configuration.md

JDruery:
Describe new configuration options: guessFrom and additional cardinalities for many to one relationships


dunglas:
Thanks @JDruery

PR #80, opened 2016-06-30
hotfix: take in account normalization on operations

Simperfit:
- Take in account itemOperation normalization 
- Take in account collectionOperation normalization


dunglas:
Thanks @Simperfit 

I'm working on a PR on core to ease support of other formats than JSON-LD.

PR #89, opened 2016-07-11
Fix tutorial api

toofff:


tutorial/api.md
```diff
@@ -89,10 +91,16 @@ Run the built-in web server:
 You can open `http://localhost:8000` with you preferred REST client.
 We recommend [Postman](https://www.getpostman.com), and you will see later that API Platform is nicely integrated with it.
 
+### With docker
```
dunglas: Docker instead of docker.

tutorial/api.md
```diff
@@ -89,10 +91,16 @@ Run the built-in web server:
...
+
+You can also use [docker](https://www.docker.com/) for the installation of your project.
```
dunglas: Same here. And I would write "to install your project".

dunglas:
I left some minor comments to fix. Thanks for those improvements!

tutorial/api.md
theofidry: the `Composer` and not `composer` for the previous chapter :)

tutorial/api.md
```diff
@@ -89,10 +91,16 @@ Run the built-in web server:
...
+
+it just make a clone of the repository Api Platform and [follow documentation docker](../deployment/docker.md)
+
 When you're done with the demo app and want to create your own API:
```
theofidry: I would add another section: this part is not specific to docker but the way it is right now makes you think so which is a bit confusing

PR #93, opened 2016-08-02
Fix DataProviders links

krismas:


core/data-providers.md
```diff
@@ -64,12 +64,11 @@ services:
 
 Tagging the service with the tag `api_platform.collection_data_provider` will enable API Platform Core to automatically
 register and use this data provider. The optional attribute `priority` allows to define the order in wich are called the
-data providers. The first data provider not throwing a `ApiPlatform\Core\Exception\ResourceClassNotSupportedException` will
-be used.
+data providers. The first data provider not throwing a `ApiPlatform\Core\Exception\ResourceClassNotSupportedException` will be used.
```
teohhanhui: Please revert this. The soft line length limit is at 120 characters, it'd seem.

krismas:
I will do that... but in the documentation many lines exceeds this limit ! What are the rules?

teohhanhui:
> but in the documentation many lines exceeds this limit ! What are the rules?

@api-platform/core-team?

theofidry:
If would say 120 by default but I don't really think it's worth to enforce strict rules.

dunglas:
Agree with @theofidry. For now I use the following rule: insert a LF char instead of the next blank character after the 120th column.

dunglas:
Thanks @krismas

PR #96, opened 2016-08-03
Fix nelmio conf & chapters' linking

kbsali:
* The doc said that Nelmio is enabled by default; it is not so! (not sure whether fixing the doc is the correct solution, or if we should update the *core* default conf instead?)
* The TOC was wrong and the linking between the different chapters too...


dunglas:
The doc hasn't been updated since the integration of Swagger UI. NelmioApiDoc isn't enabled by default anymore.

dunglas:
Looks good to me. @api-platform/core-team: anyone can double check?

soyuka:
Why are they some <br>'s?

teohhanhui:
Please rebase.

kbsali:
@soyuka : some "footers" had <br>, other not, i just made this consitent...

core/angularjs-integration.md
```diff
@@ -95,5 +95,5 @@ nga.configure(admin)
 
 You can look at what we have done as another exemple [api-platform/admin](https://github.com/api-platform/admin).
 
-Previous chapter: [Performances](performances.md)
-Next chapter: [Schema Generator: Introduction](angularjs-integration.md)
+Previous chapter: [FOSUserBundle Integration](fosuser-bundle.md)<br>
```
theofidry: `<br />`

core/nelmio-api-doc.md
```diff
@@ -13,6 +13,10 @@ If you use the standalone API Platform Core bundle, copy the following configura
 ```yaml
 # app/config/config.yml
 
+api_platform:
+    ...
```
theofidry: `#...

theofidry:
I'm not sure if the `<br />` are still useful, they where here to force the line return between the links for the next and previous chapter. Maybe they are not necessary anymore with the website?

kbsali:
is it a question? :) I don't mind removing them, but it's either all with, or all without! ;)

kbsali:
@theofidry done removing the `<br>` (and indenting issues)

dunglas:
Thanks @kbsali

PR #108, opened 2016-09-14
Add steps to install NelmioApiDocBundle

rudidev08:
Following steps 1 and 2 from https://github.com/nelmio/NelmioApiDocBundle/blob/master/Resources/doc/index.rst is needed to get NelmioApiDocBundle installed if user started with core API. However, step 3 doesn't seem to be required (routing), and step 4 is done by API Platform docs. Could you please confirm if routing step should be skipped, or should dev still do it?

### DIFF
--- a/core/nelmio-api-doc.md
+++ b/core/nelmio-api-doc.md
@@ -5,7 +5,7 @@
 [NelmioApiDocBundle](https://github.com/nelmio/NelmioApiDocBundle) (since version 2.9) has built-in support for API Platform.
 Installing it will give you access to a human-readable documentation and a nice sandbox. It is an alternative to the builtin API Platform documentation.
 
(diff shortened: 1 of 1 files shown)


teohhanhui:
The docs is outdated on this. Say for example line 5 is inaccurate.

teohhanhui:
> However, step 3 doesn't seem to be required (routing), and step 4 is done by API Platform docs. Could you please confirm if routing step should be skipped, or should dev still do it?

You need to add a route for it. What you're seeing at `/doc` by default is the Swagger UI.

dunglas:
Thanks @rudidev08

teohhanhui:
@dunglas Please see my comments above. This should not have been merged because it's misleading... :)

PR #110, opened 2016-09-23
Fix typo in the name of a service

meyerbaptiste:


dunglas:
Thanks!

teohhanhui:
This should be reverted.

Please see https://github.com/api-platform/core/blob/5c706e24a54c3879677aa4d357ce4fa346983855/tests/Fixtures/app/config/config.yml#L40 and https://github.com/api-platform/core/blob/5c706e24a54c3879677aa4d357ce4fa346983855/tests/Fixtures/app/config/config.yml#L81-L82 for the right way to do it.

PR #111, opened 2016-09-23
Remove "Registering a Custom Format in the Negotiation Library" section

meyerbaptiste:
Since Negotiation 2.0 this section is useless and doesn't work anymore.


dunglas:
Thanks Baptiste!

PR #109, opened 2016-09-19
New getting started

dunglas:
WIP

PR #114, opened 2016-09-27
YAML needs to be api_resource.yml not resource.yml

sschueller:
Also added some more example paramters. I am not sure if the xml also should be api_resource.xml?


teohhanhui:
This is already being taken care of in #109 :smile:

PR #115, opened 2016-09-30
Update doc custom filter

Eraac:


dunglas:
Thank you @Eraac! Good catch.

PR #116, opened 2016-09-30
standardize all code samples

kbsali:
very small / detail PR to standardize the code samples throughout the doc.

### DIFF
--- a/core/data-providers.md
+++ b/core/data-providers.md
@@ -27,6 +27,8 @@ The `getCollection` method must return an `array`, a `Traversable` or a [`ApiPla
 If no data is available, you should return an empty array.
 
 ```php
+<?php
+
 // src/AppBundle/DataProvider/BlogPostCollectionDataProvider.php
 
 namespace AppBundle\DataProvider;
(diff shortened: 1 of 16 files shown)


core/data-providers.md
```diff
@@ -27,6 +27,8 @@ The `getCollection` method must return an `array`, a `Traversable` or a [`ApiPla
 If no data is available, you should return an empty array.
 
 ```php
+<?php
```
dunglas: We don't do that in the Symfony doc but maybe it's clearer for newcomers. WDYT @api-platform/core-team

kbsali: either way, but it should be the same everywhere (`<?php` was there in some examples, others not...).

Should i remove them all?

soyuka: I think it's easier to copy/paste so doesn't seam like a bad idea to me

dunglas: I agree, keep them.

kbsali:
up!

theofidry:
Thanks @kbsali

PR #117, opened 2016-09-30
Fix doc for name conversion

meyerbaptiste:
As mentioned in https://github.com/api-platform/docs/pull/110#issuecomment-249548709 by @teohhanhui, my previous fix was not the good way.
It's now fixed.


dunglas:
Thanks @meyerbaptiste

teohhanhui:
I'm sorry, but the Symfony Best Practices says the user services should start with `app.`

Simperfit:
It says it should be short as possible and unique :p

teohhanhui:
So what if there's an `ApiBundle` that uses the `api.` prefix? :stuck_out_tongue:

Simperfit:
This happens 💣

dunglas:
@teohhanhui IMO it doesn't hurt because it's just docs :p. But I agree to change it to `app.` anyway to make it consistent with the Symfony doc.

PR #124, opened 2016-10-17
Update getting-started.md

clementtalleu:
The previously link was a 404


dunglas:
Thanks @clementtalleu

Issue #112, opened 2016-09-25
Dynamically added groups for (de)normalization


benjaminrau:
Hi there.

This is just a question, not a bug report.

Can you tell me where in the following example the admin_input group is added only for Books, or isnt this the case?

For my understanding with the code from this example the group admin_input is added on each ApiResource when conditions match.

https://github.com/dunglas/doc/blob/new_getting_started/core/serialization-groups-and-relations.md#changing-the-serialization-context-dynamically

Thanks for your help.

Cheers Ben

dunglas:
Indeed, in this example, the `admin_input` group is always added in the context (for any resource class).

Maybe can it be nice to add another exemple where the new group is added only for a given entity using a `is_a()` call in a guard clause.

@benjaminrau do you think you can update the doc to cover this use case?

benjaminrau:
About my proposed change on docs: 
In my project the subject was sometimes a proxy object. So i wasnt able to do ```$subject instanceof Book``` but rather ```ClassUtils::getRealClass(get_class($subject)) === Book:class```

I dont know if that is common case, if so, i should add this to docs.

dunglas:
Yes it's a common pitfall. However even with proxies, `instanceof` should work.

benjaminrau:
Closing this because the question was answered by dunglas and ive added it to docs.

PR #126, opened 2016-10-18
[DOC] Add serialization group only when object is a book

benjaminrau:
Extend the example in docs for custom serialization groups to add group only when given subject is a book. Related to #112


dunglas:
Thanks @benjaminrau

core/serialization-groups-and-relations.md
```diff
@@ -316,6 +316,7 @@ namespace AppBundle\Serializer;
 use ApiPlatform\Core\Serializer\SerializerContextBuilderInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
+use AppBundle\Entity\Book;
```
teohhanhui: `use` statements should be sorted.

PR #128, opened 2016-10-19
Fix navigation for API Platform Core library

toofff:


dunglas:
Thanks!

teohhanhui:
Is there a way for us to stop using relative links? IIRC Symfony Docs does not use relative links.

dunglas:
If we don't use relative links, the doc will not be readable on GitHub (the Symfony docs doesn't display on GitHub anyway because of the RST format).

It's important because unless Symfony, I want to publish only the last version of the doc online (less confusing and better for SEO) and redirect to GitHub for older versions.

PR #129, opened 2016-10-20
Fix pagination again

toofff:
- [x]  Fix paginate between the chapters `Distribution` and `Core library`
- [x]  Add line break between previous link and next link


dunglas:
Thanks @toofff

PR #131, opened 2016-10-21
Change link restangular for part ngAdmin

toofff:


dunglas:
@toofff after double checking, it's not ok. The current link should be ok (it points to the "Restangular" section).

toofff:
OK, I close this PR ;)

PR #135, opened 2016-11-01
fix broken link to download

ip512:
the link to https://api-platform.com/download is not working, I'm not sure if this endpoint should be fixed or if the link needs to be updated


dunglas:
Thanks!

PR #136, opened 2016-11-05
Add --rm parameter to docker-compose run command examples

mamiefurax:
Each time you run a docker-compose run command it will create a new container and not reuse an older one. So when the command you've run is finished the container stop but remain in memory. In order to avoid creating deads containers each time you run a command it's better to add the --rm parameter.


teohhanhui:
I don't understand why the documentation would recommend doing `docker-compose run` as a standard part of the installation anyway. It should not be necessary. If the docs is written in such a way, there's something wrong...

dunglas:
Thank you @mamiefurax

PR #138, opened 2016-11-08
feat: remove composer install

Simperfit:
Removing composer install commands

### DIFF
--- a/deployment/docker.md
+++ b/deployment/docker.md
@@ -6,9 +6,8 @@ stack is shipped with the API Platform distribution.
 
 To install it, run the following commands (Docker must be installed on your system):
 
-    docker-compose up # Download, build and run Docker images
-    docker-compose run --rm web composer install -o -n # Install Composer dependencies
-    docker-compose run --rm web bin/console doctrine:schema:create # Create the MySQL schema
+    docker-compose up -d # Download, build and run Docker images
+    docker-compose exec web bin/console doctrine:schema:create # Create the MySQL schema
 
 Your project will be accessible at the URL `http://127.0.0.1`.
 
--- a/deployment/docker.md
+++ b/deployment/docker.md
@@ -6,8 +6,8 @@ stack is shipped with the API Platform distribution.
 
 To install it, run the following commands (Docker must be installed on your system):
 
-    docker-compose up -d # Download, build and run Docker images
-    docker-compose exec web bin/console doctrine:schema:create # Create the MySQL schema
+    $ docker-compose up -d # Download, build and run Docker images
+    $ docker-compose exec web bin/console doctrine:schema:create # Create the MySQL schema
 
 Your project will be accessible at the URL `http://127.0.0.1`.
 
--- a/distribution/index.md
+++ b/distribution/index.md
@@ -52,7 +52,7 @@ Apache and MySQL using [Docker Compose](https://docs.docker.com/compose/):
 The first time you start the containers, Docker downloads and builds images for you. It will take some time, but don't worry,
 this is done only once. Starting servers will then be lightning fast.
 
-In order to see container's logs you will have to do 
+In order to see container's logs you will have to do:
 
     $ docker-compose logs -f # follow the logs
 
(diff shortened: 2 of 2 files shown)


deployment/docker.md
```diff
@@ -7,7 +7,6 @@ stack is shipped with the API Platform distribution.
 To install it, run the following commands (Docker must be installed on your system):
 
     docker-compose up # Download, build and run Docker images
```
teohhanhui: `docker-compose up -d`

Run in the background (detached mode).

(This needs to be changed in other places as well...)

deployment/docker.md
```diff
@@ -7,7 +7,6 @@ stack is shipped with the API Platform distribution.
...
-    docker-compose run --rm web composer install -o -n # Install Composer dependencies
     docker-compose run --rm web bin/console doctrine:schema:create # Create the MySQL schema
```
teohhanhui: `docker-compose exec web bin/console doctrine:schema:create`

Execute the command in the running container, instead of starting a new one and removing it right after.

(This needs to be changed in other places as well...)

distribution/index.md
```diff
@@ -60,20 +60,15 @@ with its awesome [Symfony](https://confluence.jetbrains.com/display/PhpStorm/Get
...
-given command in the container. You may want [to create an alias](http://www.linfo.org/alias.html) to easily run commands
-inside the container. Here, we are installing libraries required by the project using the `composer` tool included in the
-API Platform image.
-
 The API Platform Standard Edition comes with a dummy entity for test purpose: `src/AppBundle/Entity/Foo.php`. We will remove
 it later, but for now, create the related database table:
 
     docker-compose run --rm web bin/console doctrine:schema:create
 
+The `web` container is where your project stands. Prefixing a command by `docker-compose run --rm web` allows to execute the
```
teohhanhui: Incorrect. `docker-compose exec ...` is what you want.

deployment/docker.md
Simperfit: I don't think this should be by default

deployment/docker.md
Simperfit: agreed

deployment/docker.md
Simperfit: I've added an explication about using detached mode && logs.

teohhanhui: If you don't use detached mode, you cannot proceed with the next step of this guide.

(Well, of course you could still open a new terminal, but that's probably not what's expected.)

distribution/index.md
```diff
@@ -52,6 +52,14 @@ Apache and MySQL using [Docker Compose](https://docs.docker.com/compose/):
 The first time you start the containers, Docker downloads and builds images for you. It will take some time, but don't worry,
 this is done only once. Starting servers will then be lightning fast.
 
+If you don't want to see the output directly with the running containers you should do
+
+    $ docker-compose up -d # detached mode
+
+In order to see container's logs you will have to do 
+
+    $ docker-compose logs
```
teohhanhui: Probably `docker-compose logs -f` to **f**ollow.

deployment/docker.md
Simperfit: Right ! in order to let people follow it may be better to use detached mode

distribution/index.md
```diff
@@ -60,19 +64,14 @@ with its awesome [Symfony](https://confluence.jetbrains.com/display/PhpStorm/Get
...
-The `web` container is where your project stands. Prefixing a command by `docker-compose run --rm web` allows to execute the
-given command in the container. You may want [to create an alias](http://www.linfo.org/alias.html) to easily run commands
-inside the container. Here, we are installing libraries required by the project using the `composer` tool included in the
-API Platform image.
-
 The API Platform Standard Edition comes with a dummy entity for test purpose: `src/AppBundle/Entity/Foo.php`. We will remove
 it later, but for now, create the related database table:
 
-    docker-compose run --rm web bin/console doctrine:schema:create
+    docker-compose exec web bin/console doctrine:schema:create
```
dunglas: Please add a `$` like in other commands.

distribution/index.md
```diff
@@ -47,11 +47,15 @@ Then, if you do not already have Docker on your computer, [it's the right time t
 Open a terminal, and navigate to the directory containing your project skeleton. Then, run the following command to start
 Apache and MySQL using [Docker Compose](https://docs.docker.com/compose/):
 
-    $ docker-compose up
+    $ docker-compose up -d # Running in detached mode
 
 The first time you start the containers, Docker downloads and builds images for you. It will take some time, but don't worry,
 this is done only once. Starting servers will then be lightning fast.
 
+In order to see container's logs you will have to do
```
dunglas: Missing colon  (without space before the colon)

dunglas:
Thanks @Simperfit

PR #139, opened 2016-11-08
feat: update events docs

Simperfit:


core/events.md
```diff
@@ -12,20 +12,35 @@ To do so, API Platform Core leverages [events triggered by the Symfony HTTP Kern
 You can also hook your own code to those events. They are very handy and powerful extension points available at all points
 of the request lifecycle.
 
+API Platform comes with a (class)[https://github.com/api-platform/core/blob/master/src/EventListener/EventPriorities.php] `ApiPlatform\Core\EventListener\EventPriorities` that declare the events priorities by using constant
```
dunglas: The [`ApiPlatform\Core\EventListener\EventPriorities`](https://github.com/api-platform/core/blob/master/src/EventListener/EventPriorities.php) class comes with a convenient set of class's constants corresponding to commonly used priorities:

core/events.md
```diff
@@ -12,20 +12,35 @@ To do so, API Platform Core leverages [events triggered by the Symfony HTTP Kern
...
+`PRE_DESERIALIZE`  | `kernel.request` | 3        |
+`POST_DESERIALIZE` | `kernel.request` | 1        |
+`PRE_VALIDATE`     | `kernel.view`    | 65       |
+`POST_VALIDATE`    | `kernel.view`    | 63       |
+`PRE_WRITE`        | `kernel.view`    | 33       |
+`POST_WRITE`       | `kernel.view`    | 31       |
+`PRE_RESPOND`      | `kernel.view`    | 9        |
+`POST_RESPOND`     | `kernel.view`    | 7        |
+
 Built-in event listeners are:
```
dunglas: I would move the listener example just after this table and the following table at the end of this file.

Btw, the example must be updated to use the `POST_WRITE` constant.

dunglas:
Thanks @Simperfit

PR #147, opened 2016-11-17
Add documentation for eager loading

meyerbaptiste:


core/performance.md
```diff
@@ -236,6 +238,166 @@ services:
...
+
+    # ...
+```
+
+Be careful when you exceed your limit, it's often the result of circular references. The use of [serializer groups](serialization-groups-and-relations.md)
+can be a good solution to fix the issue.
+
+##### Force eager
+
+As mentioned above, by default we force eager loading for all relations. This behaviour can be modified in the
```
Simperfit: `This behaviour can be modified with the configuration in order to apply it only on join relations having the ...`

core/performance.md
```diff
@@ -236,6 +238,166 @@ services:
...
+    eager_loading:
+        force_eager: false
+
+    # ...
+```
+
+##### Override at resource and operation level
+
+When eager loading is enabled, whatever the status of the `force_eager` parameter, you can easily override it directly
+from the configuration of each resource. For now, you can do this at the resource level, at the operations level, or both:
```
Simperfit: You say that there will be more way to do this, but you are not saying what will be next. IMO you don't need to say `For now`.

core/performance.md
Simperfit: IMO this can still be better

core/performance.md
```diff
@@ -236,6 +238,166 @@ services:
             - {name: api_platform.doctrine.orm.query_extension.item}
 ```
 
+#### Eager loading
+
+By default Doctrine comes with [lazy loading](http://doctrine-orm.readthedocs.io/en/latest/reference/working-with-objects.html#by-lazy-loading).
+Usually a killer time-saving feature and also a performance killer with large applications.
```
dunglas: I agree with this statement.

soyuka: haha I do agree too, this feature should give more warning than it actually does

soyuka:
We may want to remove my part on performances as it's not right anymore because there are no partial select anymore https://github.com/meyerbaptiste/doc/blob/3216e636ec21917790456d9a2554edb96bed865c/core/performance.md#unserialized-properties-hydratation.

meyerbaptiste:
Done @soyuka.

Simperfit:
Thanks @meyerbaptiste

PR #149, opened 2016-11-18
Fix operations configuration

meyerbaptiste:


dunglas:
Thanks @meyerbaptiste

PR #150, opened 2016-11-18
Fix syntax and index

meyerbaptiste:


dunglas:
Thanks @meyerbaptiste

PR #153, opened 2016-11-22
Fix path for the configuration of resources files

meyerbaptiste:
According to https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php#L291, path was wrong, my bad.


Simperfit:
Thanks @meyerbaptiste

PR #160, opened 2016-12-07
Add new loader_paths option into configuration

digilist:
See api-platform/core#869

### DIFF
--- a/core/configuration.md
+++ b/core/configuration.md
@@ -108,6 +108,12 @@ api_platform:
             mime_types: ['application/ld+json']
             
         # ...
+    
+    # Custom paths to look for api resource configurations
+    loader_paths:
+        annotation: []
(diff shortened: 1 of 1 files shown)


digilist:
Sorry, I did something stupid and deleted my fork too early. It looks like I cannot recover it. Are you able to modify this PR? For example with the following path:

"%kernel.root_dir%/../src/api/user_resources.yml".

Otherwise, I think that I need to submit a new PR.

Simperfit:
@digilist you cou please submit a new PR, i can't edit the file ?

dunglas:
I propose to merge this PR as is and to open a new one with the change suggested by @sroze. We cannot restore the fork either.

digilist:
I opened a new PR.

PR #166, opened 2016-12-27
Update index.md

renan-taranto:
The isbn provided as an example of an invalid number was valid actually.


dunglas:
Thanks @renan-taranto

PR #167, opened 2016-12-27
Fixes Gherkin books.feature file

renan-taranto:
Related to https://github.com/api-platform/api-platform/issues/197

### DIFF
--- a/distribution/testing.md
+++ b/distribution/testing.md
@@ -18,7 +18,6 @@ the API without having to write a single line of PHP.
 
 ```gherkin
 # features/books.feature
-
 Feature: Manage books and their reviews
   In order to manage books and their reviews
   As a client software developer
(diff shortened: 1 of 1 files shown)


dunglas:
First of all, thanks for your fixes! It's really appreciated.

Escaping slashes with backslashes is not mandatory. I would prefer to remove this to make the code more readable. Can you change that please @renan-taranto?

renan-taranto:
@dunglas I forked the Api Platform core, and changed it to stop escaping slashes (https://github.com/renan-taranto/core/tree/patch-1).
The problem is that I'm totally noob on contributing to projects like this. Some guidance on how to properly run it locally so I could run the tests would be much appreciated.

dunglas:
@renan-taranto I'm not sure to understand, the documentation is just a bunch of plain old Markdown files in a separated repo (this one). You can edit them locally using your preferred text editor or directly on GitHub by editing your branch.

But don't worry, I've added a commit directly in your branch and everything is good now. Thanks for your contribution!

renan-taranto:
@dunglas I tought you asked me to change the API Platform core to stop escaping slashes when encoding to json. Alright then.

PR #168, opened 2017-01-03
Add warning for JMS Serializer

tcitworld:
https://github.com/api-platform/core/issues/190

### DIFF
--- a/core/configuration.md
+++ b/core/configuration.md
@@ -110,6 +110,14 @@ api_platform:
         # ...
 ```
 
+## JMS Serializer
+Disable the short service "serializer" alias name as it conflicts with API Platform.
+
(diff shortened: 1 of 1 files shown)


dunglas:
Hi,

First of all, thanks for your contribution @tcitworld. However, this already mentioned in the documentation: https://api-platform.com/docs/troubleshooting#using-api-platform-and-jms-serializer-in-the-same-project

Do you think it should be made more visible?

tcitworld:
At least I didn't see it, so I guess so. :)
Maybe the troubleshooting item in the table of contents should be put somewhere else.

dunglas:
Closing for now as this is already documented.

PR #169, opened 2017-01-03
symfony event was not good for PRE_RESPOND and POST_RESPOND…

fernandodbc:
Just a little typo fix in the event doc


teohhanhui:
No, the documentation is correct. Please check the code:

https://github.com/api-platform/core/blob/3fe7ceb0a725f6fada9fb5c398249f800cebb5fd/src/Bridge/Symfony/Bundle/Resources/config/api.xml#L129-L131

PR #171, opened 2017-01-18
Add an example of dynamic validation groups

### LINKED ISSUE #152: Dynamic validation groups
Can you please provide examples for this scenario, because documentation https://api-platform.com/docs/core/validation#using-validation-groups is not so obvious.

@dunglas @kbsali

dunglas:
Closes #152

### DIFF
--- a/core/validation.md
+++ b/core/validation.md
@@ -78,8 +78,6 @@ class Book
      */
     public static function validationGroups(self $book)
     {
-        // 
-        
         return ['a'];
     }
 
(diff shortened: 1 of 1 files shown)


core/validation.md
```diff
@@ -54,6 +54,47 @@ If you need to dynamically determine which validation groups to use for an entit
...
+class Book
+{
+    /**
+     * Return dynamic validation groups.
+     *
+     * @param self $book Contains the instance of Book to validate.
+     *
+     * @return string[]
+     */
+    public static function validationGroups(self $book)
```
teohhanhui: Ideally this should belong in another class, no?

dunglas: It depends of the use case, if it's only some lines of code I think it's ok to put it in the entity.
But the goal here is just to give a short example, introducing a new class is overkill for a doc.

aledeg:
Silly question.
Is it possible to use a service instead of a class? If so, how can I do it?

teohhanhui:
We can make it work with a service with some extra glue (in the Symfony Bridge).

dunglas:
Yes but it's currently not supported.

PR #172, opened 2017-01-18
Enhance docs for users of Windows

dunglas:
Sorry I pushed directly to the main repository without doing a PR in 2c0e644224c66428184436031d559a40e97c1f37.

This PR complete the previous commit.

### DIFF
--- a/troubleshooting.md
+++ b/troubleshooting.md
@@ -6,7 +6,7 @@ This is a list of common pitfalls on using API Platform, and how to avoid them.
 
 ### With Docker Toolbox on Windows
 
-If you get errors like the following when running `docker-compos up` on Windows:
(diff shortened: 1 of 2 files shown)


troubleshooting.md
```diff
@@ -6,7 +6,7 @@ This is a list of common pitfalls on using API Platform, and how to avoid them.
 
 ### With Docker Toolbox on Windows
```
teohhanhui: We should add a note to warn that Docker Toolbox is not the recommended method (and mention Docker for Windows).

dunglas: The problem is: Docker for Windows only works for pro editions. And Docker itself already warn about that.

PR #173, opened 2017-01-25
Add Exception configuration documentation

GregoireHebert:


core/exceptions.md
```diff
@@ -0,0 +1,94 @@
+# Exceptions
```
dunglas: I would rename this entry "Error handling" (`errors.md`).

core/exceptions.md
```diff
@@ -0,0 +1,94 @@
...
+
+```YAML
+# app/config/config.yml
+
+ api_platform:
+
+    # ...
+
+    exception_to_status:
+        # By default there are two exceptions defined that you can customize.
```
dunglas:
```yaml
# app/config/config.yml

services:
# Map exceptions to HTTP status codes using the `exception_to_status` configuration key 
exception_to_status:
    # The 2 following exceptions are handled by default
    Symfony\Component\Serializer\Exception\ExceptionInterface: 400                     # Use a raw status code (recommended)
    ApiPlatform\Core\Exception\InvalidArgumentException: 'HTTP_BAD_REQUEST' # Or with a constant of `Symfony\Component\HttpFoundation\Response`

    AppBundle\Exception\MyException: 401 # Custom exceptions can easily be handled
```

core/exceptions.md
```diff
@@ -0,0 +1,94 @@
...
+        Foo\Bar\BazException: 401
+
+        # ...
+```
+
+## Custom Exception
+
+As in any php application, your exceptions have to extends the \Exception class or any of it's children.
+
+```PHP
```
dunglas: php

core/exceptions.md
```diff
@@ -0,0 +1,94 @@
...
+<?php
+
```
dunglas: // src/AppBundle/Exception/ProductNotFoundException.php

core/exceptions.md
```diff
@@ -0,0 +1,94 @@
...
+namespace AppBundle\Exception;
+
+/**
+ * Class ProductNotFoundException
```
dunglas: I would remove this docblock.

core/exceptions.md
```diff
@@ -0,0 +1,94 @@
...
+ */
+class ProductNotFoundException extends \Exception
```
dunglas: `final`

core/exceptions.md
```diff
@@ -0,0 +1,94 @@
...
+{
+
+    public function __construct($code = 0) {
```
dunglas: I would remove this constructor.

core/exceptions.md
```diff
@@ -0,0 +1,94 @@
...
+        parent::__construct("Product Not found", $code);
+    }
+}
+```
+
+```PHP
```
dunglas: php

core/exceptions.md
```diff
@@ -0,0 +1,94 @@
...
+<?php
+
```
dunglas: // src/AppBundle/Manager/CartManager.php

core/exceptions.md
```diff
@@ -0,0 +1,94 @@
...
+namespace AppBundle\Manager;
+
+use AppBundle\Entity\Product;
+use AppBundle\Exception\ProductNotFoundException;
+
+/**
+ * Class CartManager
```
dunglas: I would remove this constructor.

core/exceptions.md
```diff
@@ -0,0 +1,94 @@
...
+ */
+class CartManager
```
dunglas: `final`

core/exceptions.md
```diff
@@ -0,0 +1,94 @@
...
+{
+    const NOTFOUND_DOESNOTEXISTS = 51;
```
dunglas: I would strip the `NOTFOUND` prefix and a visibility to all those constants.

core/exceptions.md
```diff
@@ -0,0 +1,94 @@
...
+    const NOTFOUND_OUTOFSTOCK = 53;
+
+    /**
+     * @param integer $id
+     *
+     * @throws ProductNotFoundException
+     *
+     * @return Product $product
+     */
+    public function addProductToCart($id)
```
dunglas: `public function addProductToCart(int $id)`

core/exceptions.md
```diff
@@ -0,0 +1,94 @@
...
+ * Class CartManager
+ */
+class CartManager
+{
+    const NOTFOUND_DOESNOTEXISTS = 51;
+    const NOTFOUND_DEACTIVATED = 52;
+    const NOTFOUND_OUTOFSTOCK = 53;
+
+    /**
+     * @param integer $id
```
dunglas: int

core/exceptions.md
```diff
@@ -0,0 +1,94 @@
...
+        $product = /*...*/;
+
+        if (!$product->getVirtualStock()) {
+            // Using internal codes for a better understanding of what's going on
+            throw new ProductNotFoundException(self::NOTFOUND_OUTOFSTOCK);
+        }
+
+        /*...*/
+
+        return $cart;
```
dunglas: I would return nothing (CQRS like)

core/exceptions.md
```diff
@@ -0,0 +1,94 @@
...
+    public function addProductToCart($id)
+    {
+        $product = /*...*/;
+
+        if (!$product->getVirtualStock()) {
+            // Using internal codes for a better understanding of what's going on
+            throw new ProductNotFoundException(self::NOTFOUND_OUTOFSTOCK);
+        }
+
+        /*...*/
```
dunglas: Can be removed

dunglas:
I would merge both sections (the last section should be an example illustrating the config). I propose to start with the tiny exception example (it can simplified IMO, and it should be more integrated with API Platform, for instance used as an event listener).

dunglas:
By the way it would be nice to show that API Platform automatically handle validation errors.

core/exceptions.md
```diff
@@ -0,0 +1,94 @@
...
+
+        return $cart;
+    }
+}
+```
+
+It doesn't have to be an HttpException, any type of Exception can be thrown. The best part is that API Platform already takes care of how the error is handled and returned. If you configured your API to respond in a JsonLD format, your error will be returned in JsonLD format as well.
```
dunglas:
```
The exception doesn't have to be a Symfony's `HttpException`. Any type of `Exception` can be thrown. The best part is that API Platform already takes care of how the error is handled and returned. For instance the API is configured to respond in JSON-LD, the error will be returned in this format as well.
```

GregoireHebert:
I'll do the example for validation errors and transform the error as an event listener tomorrow.

GregoireHebert:
Ok I made some changes, but I not really convinced that what you had in mind.
Let me know :)

PR #175, opened 2017-02-06
Add nulls comparison to OrderFilter

teohhanhui:
Documents api-platform/core#838

### DIFF
--- a/core/filters.md
+++ b/core/filters.md
@@ -321,9 +321,35 @@ will not be applied unless you configure a default order direction to use:
 services:
     offer.order_filter:
         parent:    'api_platform.doctrine.orm.order_filter'
-        arguments: [ { id: 'ASC', name: 'DESC' } ]
+        arguments: [ { id: { default_direction: 'ASC' }, name: { default_direction: 'DESC' } } ]
         tags:      [ { name: 'api_platform.filter', id: 'offer.order' } ]
 ```
+
+### Comparing with Null Values
+
+When the property used for ordering can contain `null` values, you may want to specify how `null` values are treated in
+the comparison:
+
+Description                          | Strategy to set
+-------------------------------------|---------------------------------------------------------------------------------------------
+Use the default behavior of the DBMS | `null`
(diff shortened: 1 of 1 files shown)


core/filters.md
```diff
@@ -321,10 +321,36 @@ will not be applied unless you configure a default order direction to use:
 services:
     offer.order_filter:
         parent:    'api_platform.doctrine.orm.order_filter'
-        arguments: [ { id: 'ASC', name: 'DESC' } ]
+        arguments: [ { id: { default_direction: 'ASC' }, name: { default_direction: 'DESC' } } ]
```
dunglas: Does the short syntax still works? It looks very verbose here.

core/filters.md
```diff
@@ -321,10 +321,36 @@ will not be applied unless you configure a default order direction to use:
...
         tags:      [ { name: 'api_platform.filter', id: 'offer.order' } ]
 ```
 
+### Comparing with Null Values
```
dunglas: `null` Values?

core/filters.md
teohhanhui: The shorthand syntax is retained for BC, but I'd like to see it deprecated. What do you think?

teohhanhui: I think we can use `~` for one of the two properties here. It's good to show the usual / more common case where you don't specify a default direction.

core/filters.md
teohhanhui: I think we should keep that out of the headings and put it in plain English.

core/filters.md
dunglas: I prefer to keep the short syntax as is because it's the most common use case and it must be as less verbose as possible.

teohhanhui:
I'd argue that the most common use case is actually `~` (`null`)... I think
it's much preferable to be explicit, instead of having a default direction
which is not visible.

Sometimes I feel like calling it a misfeature. :laugh:

core/filters.md
teohhanhui: I do not agree that it's "the most common use case".

dunglas:
Thanks @teohhanhui

PR #176, opened 2017-02-27
fix: Some typo on the doc

Gregcop1:


Simperfit:
Thanks @Gregcop1

teohhanhui:
I'm curious why we don't stick to https://schema.org/Book for this? Is there a good reason?

dunglas:
It was to make it clear that it's not related at all with Schema.org.

dunglas:
@Simperfit this patch should have been merged in 2.0

PR #177, opened 2017-02-27
fix: Update the service's name in the doc

Gregcop1:
Dunno if the doc is outdated or if it's an mistake in the docker-compose configuration's file


teohhanhui:
Thanks for spotting that, @Gregcop1!

PR #179, opened 2017-02-28
fixed item provider example

iCodr8:


dunglas:
Merged in 2.0 with 6162e97d31842ad3cdffc1f9f9579d3efc315697.

Thanks @iCodr8

PR #184, opened 2017-03-15
Delete extra `}` in hydra context

jakesyl:
This makes the example syntactically correct.


soyuka:
Good catch, thanks @jakesyl

PR #187, opened 2017-03-29
fix minor visual bug on docs

Tim59:
@soyuka , I discover a bug when i read the docs, on the filters parts


soyuka:
Nice catch, thanks @Tim59 !

PR #189, opened 2017-03-31
Fix trough through

soyuka:


dunglas:
Can this be applied to v2?

PR #193, opened 2017-04-12
Fix list indentation in TOC (index.md)

teohhanhui:
> List items may consist of multiple paragraphs. Each subsequent paragraph in a list item must be indented by either 4 spaces or one tab

http://daringfireball.net/projects/markdown/syntax#list

Also see https://github.com/waylan/Python-Markdown/issues/3#issuecomment-63399983


teohhanhui:
Not an RFC, but oh well... Motivation: I was trying to view the docs on GitHub but noticed that it's one long, flat list. :sweat_smile:

I also sneaked in another neat trick: Indenting the body of the list item. It looks super neat but renders the same. Especially useful when the body spans multiple paragraphs.

dunglas:
Good to know! Thanks.

PR #195, opened 2017-04-24
Add doc for 'api_resources_directory' configuration value.

omissis:
Add documentation for the configuration value that adds the possibility to customize the name of the directory where api resources can be stored.


dunglas:
Thanks @omissis

PR #203, opened 2017-04-28
correct a few typos

jon1328:
Correcting a few typos in the documentation. I understand those corrections should be branched off from 2.0.


teohhanhui:
Thanks @jon1328!

PR #208, opened 2017-05-15
[FIX] Correct doc about docker-compose commands

anaelChardan:
Hi !

I've find an issue about docker-compose commands to execute in your documentation, the web container does not exist as it is name php :).

Regards.


dunglas:
Merged in 0b2f6c2618759a90f67ab3582c4396e57cab6f22. Thanks @anaelChardan

anaelChardan:
Welcome, thanks for all :)

PR #212, opened 2017-05-23
Change confusing reference to removed service definition. 

sh41:
Either the service definition should have stayed in the document or the line saying "First register the following service:" should have been removed. I went with the former. Was removed in c500601deb59085928b1b289fab40b4ce221596b.


sh41:
Now see from https://github.com/api-platform/api-platform/issues/84 that the service definition should have been dropped. Have removed and changed the confusing reference to the removed service definition.

soyuka:
Thnaks @sh41

PR #221, opened 2017-06-09
Fix JWT URL

Sybio:
Wrong path to jwt.io website on page https://api-platform.com/docs/core/jwt.

Fixing https://api-platform.com/docs/core/jwt.io to https://jwt.io/


Simperfit:
Thanks @Sybio !

PR #225, opened 2017-06-18
Adjust when the properties is null

jairoFg12:


dunglas:
Can you tell a bit more about this change?

core/filters.md
```diff
@@ -464,13 +464,15 @@ final class RegexpFilter extends AbstractFilter
...
     {
         $description = [];
-        foreach ($this->properties as $property => $strategy) {
-            $description['regexp_'.$property] = [
-                'property' => $property,
-                'type' => 'string',
-                'required' => false,
-                'swagger' => ['description' => 'Filter using a regex. This will appear in the Swagger documentation!'],
-            ];
+        if (!empty($this->properties)) {
```
dunglas: It would be better to use a guard clause:

```php
public function getDescription(string $resourceClass): array
{
    if (!$this->properties) {
        return []:
    }

   $description = [];
   // ...
```

dunglas:
Thanks @jairoFg12

PR #228, opened 2017-06-21
Example Doctrine filter usage, flagging the issue on the priority for…

djoos:
…the filter to work as expected

More information on Doctrine filters can be found at http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/filters.html
Using a great tutorial on implementing a "UserAware" Doctrine filter in Symfony (http://blog.michaelperrin.fr/2014/12/05/doctrine-filters/), this information not only allows an API Platform dev to implement a Doctrine filter, but also sail around the issue where having a too low priority makes the Pagination go off, as discussed in https://github.com/api-platform/core/issues/1185.

Please let me know if any changes are needed to get this merged. Thanks in advance!

Kind regards,
David

### DIFF
--- a/core/filters.md
+++ b/core/filters.md
@@ -725,7 +725,7 @@ final class UserFilterConfigurator
     public function onKernelRequest(): void
     {
         if (!$user = $this->getUser()) {
-            return;
+            throw new \RuntimeException('There is no authenticated user.');
         }
 
         $filter = $this->em->getFilters()->enable('user_filter');
--- a/core/filters.md
+++ b/core/filters.md
@@ -597,7 +597,7 @@ final class UserAware
 }
 ```
 
-Then, Let's mark the `Order` entity as a "user aware" entity.
+Then, let's mark the `Order` entity as a "user aware" entity.
 
 ```php
 <?php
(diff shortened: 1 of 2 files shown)


core/filters.md
```diff
@@ -522,6 +522,231 @@ class Offer
...
+    public function __construct(ObjectManager $em, TokenStorageInterface $tokenStorage, Reader $reader)
+    {
+        $this->em = $em;
+        $this->tokenStorage = $tokenStorage;
+        $this->reader = $reader;
+    }
+
+    public function onKernelRequest(): void
+    {
+        if (!$user = $this->getUser()) {
```
dunglas: This one looks tricky. If a user is anonymous, shouldn't an exception be thrown instead? It looks like the current implementation can cause security concerns with some firewall setups.

dunglas:
Thank you very much for this doc entry @djoos. It's really appreciated. I made some cosmetic / CS  changes in new commits.
What do you think of my remaining comment?

core/filters.md
djoos: Ok, I see your point - I actually kept the logic here mainly to filtering while security would be handled on the firewall level. It's up to the implementation to decide what needs to happen exactly when dealing with an anonymous user; skipping the filter is indeed most likely not desirable...

Perhaps filtering on id "-1" is an option, instead of returning? Feels quite like a hack though.

What do you think?

dunglas: In my projects I add something like `WHERE 1=2` but it's definitely a trick. Throwing an exception telling that the security must be enabled looks good enough for a doc entry to me.

djoos: Yeah, I agree!

dunglas:
Thanks @djoos

PR #235, opened 2017-08-18
Fix api-platform/core #1246: configure default order on ApiResource annotation

vincentchalamon:


core/default-order.md
```diff
@@ -0,0 +1,37 @@
+# Override default order
```
dunglas: What do you think about naming the file `order.md` or `ordering.md` and merging everything related to ordering (the global config option, the filter and this config option) in this file?

It will be easier to understand for a new comer IMP.

Btw the title must be in Title Case (`Override Default Order`)

core/default-order.md
```diff
@@ -0,0 +1,37 @@
...
+
+API Platform Core provides an easy way to override default order in your collection.
```
dunglas: `of the collections`?

dunglas: I think we can get rid of `Core` too.

core/default-order.md
```diff
@@ -0,0 +1,37 @@
...
+
+By default, it will order by resource identifier(s) using ASC direction. If you want to customize this order, you must add an `order` attribute on your ApiResource annotation:
```
dunglas: > By default, it will order by resource identifier(s) using ASC direction. 

IIRC, it's not true. By default don't change the order of the underlying DBMS, and it is not defined nor consistent: https://dba.stackexchange.com/a/6053

`@ApiResource` annotation. Maybe it is worth mentioning that it can also be configured using XML and YAML.

core/default-order.md
```diff
@@ -0,0 +1,37 @@
...
+
+// src/AppBundle/Entity/Book.php
+
+namespace AppBundle\Entity;
+
+use ApiPlatform\Core\Annotation\ApiProperty;
+use ApiPlatform\Core\Annotation\ApiResource;
+
+/**
+ * @ApiResource(attributes={"order"={"foo", "ASC"}})
```
dunglas: See https://github.com/api-platform/core/pull/1321/files#r134408183

core/default-order.md
```diff
@@ -0,0 +1,37 @@
...
+ */
+class Book
+{
+    // ...
+
+    /**
+     * ...
+     * @ApiProperty()
```
dunglas: This annotation is useless, you can remove it.

core/default-order.md
vincentchalamon: I didn't explain this default order in the order filter cause it's a different feature. I better think we should mention it in the order filter and in the global config option, but keep it separated.

core/default-order.md
vincentchalamon: > `of the collections`

Why plural?

vincentchalamon: I took `Core` keyword from many `.md` files in this project. If we get rid of it, please open a new PR to remove it everywhere.

core/default-order.md
vincentchalamon: The `Bridge/Doctrine/Orm/Extension/OrderExtension` default order collection by identifier(s): https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Extension/OrderExtension.php#L56

core/default-order.md
vincentchalamon: Cf. https://github.com/api-platform/core/pull/1321#discussion_r134409607

core/default-order.md
dunglas: IMO having a specific doc for all ordering features would be better in term of understanding for a newcomer, it was my point.

core/default-order.md
dunglas: Ok it's a bit more complicated. It's order by default only if you use the Symfony bridge (https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php#L92), if you use the class directly, `$this->order` is `null` and the collection isn't ordered: https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/Orm/Extension/OrderExtension.php#L54

But ok to keep it as is.

core/default-order.md
vincentchalamon: @dunglas Could it be interesting to configure a default order by an association, like `author.username`?

core/default-order.md
```diff
@@ -0,0 +1,88 @@
...
+    public $foo;
+
+    /**
+     * ...
+     */
+    public $bar;
+}
+```
+
+It's also possible to configure the default filter on an association property:
```
meyerbaptiste: Can you remove this part? So we can merge your PR. We will add it when the https://github.com/api-platform/core/pull/1324 will be merged.

core/default-order.md
```diff
@@ -0,0 +1,88 @@
+# Override Default Order
+
+API Platform Core provides an easy way to override default order in your collection.
+
+By default, it will order by resource identifier(s) using ASC direction. If you want to customize this order, you must add an `order` attribute on your ApiResource annotation:
```
teohhanhui: By default, items in the collection are ordered in ascending (ASC) order by their resource identifier(s).

core/default-order.md
```diff
@@ -0,0 +1,88 @@
+# Override Default Order
+
+API Platform Core provides an easy way to override default order in your collection.
```
teohhanhui: ... override the default order of items in your collection.

meyerbaptiste:
Thanks @vincentchalamon, don't forget to add the removed paragraph when api-platform/core#1324 will be merged:
```markdown
It's also possible to configure the default filter on an association property:

<?php

// src/AppBundle/Entity/Book.php

namespace AppBundle\Entity;

/**
 * @ApiResource(attributes={"order"={"author.username"}})
 */
class Book
{
    // ...

    /**
     * @var User
     */
    public $author;
}
```

PR #239, opened 2017-08-23
Update the exception to status config

meyerbaptiste:
See https://github.com/api-platform/core/pull/1251


dunglas:
Thanks @meyerbaptiste!

PR #243, opened 2017-08-30
Mention whitelist on  serializer filters

antograssiot:
Document api-platform/core#1333 and api-platform/core#1339

### DIFF
--- a/core/filters.md
+++ b/core/filters.md
@@ -489,6 +489,18 @@ services:
         tags:      [ { name: 'api_platform.filter', id: 'book.group' } ]
 ```
 
+To avoid uncontrolled data exposure, you can also specify a whitelist of serialization groups as a third argument:
+
+```yaml
+# app/config/services.yml
+
+services:
+    book.group_filter:
+        parent:    'api_platform.serializer.group_filter'
+        arguments: [ 'groups', false, ['allowed_group', 'safe_group'] ]
+        tags:      [ { name: 'api_platform.filter', id: 'book.group' } ]
+```
+
 ### Property filter
 
 The property filter add the possibility to filter serialization properties.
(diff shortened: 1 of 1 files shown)


core/filters.md
```diff
@@ -489,6 +489,19 @@ services:
         tags:      [ { name: 'api_platform.filter', id: 'book.group' } ]
 ```
 
+To avoid uncontrolled data exposure, you can also specify a whitelist of serialization groups as a third argument
+`true`:
```
Simperfit: why is true here ?

core/filters.md
```diff
@@ -557,6 +570,19 @@ services:
         tags:      [ { name: 'api_platform.filter', id: 'book.property' } ]
 ```
 
+To avoid uncontrolled data exposure, you can also specify a whitelist of properties as a third argument
+`true`:
```
Simperfit: same here?

core/filters.md
```diff
@@ -489,6 +489,19 @@ services:
...
+To avoid uncontrolled data exposure, you can also specify a whitelist of serialization groups as a third argument
+`true`:
+
+```yaml
+# app/config/services.yml
+
+services:
+    book.group_filter:
+        parent:    'api_platform.serializer.group_filter'
+        arguments: [ 'groups', false, ['allowed_group', 'safe_group'] ]
```
meyerbaptiste: You have to add the default value (`null`) in the first service definition above (L442). Same for the other filter.

antograssiot: why that ? null is the default...

core/filters.md
antograssiot: oupsie

antograssiot:
Should I change something else here ?

PR #244, opened 2017-08-31
Fix api-platform/core issue #1326 (DateFilter before/after filters on before or equals/after or equals)

blesenechal:
Doc PR for issue api-platform/core#1326 and PR api-platform/core#1343


Simperfit:
Thanks @blesenechal

PR #247, opened 2017-09-07
Update troubleshooting.md

mysiar:
it is a little bit misleading, JMS serializer on its own doesn't replace serializer service, bundle does


dunglas:
Thanks!

antograssiot:
This paragraph isn't relevant since JMS serializer bundle v2
https://github.com/schmittjoh/JMSSerializerBundle/blob/master/UPGRADING.md

PR #259, opened 2017-09-11
Add use of service in dynamic validation group

pierre-H:
Add doc for https://github.com/api-platform/core/pull/1198

### DIFF
--- a/core/validation.md
+++ b/core/validation.md
@@ -95,29 +95,33 @@ class Book
 }
 ```
 
-You can also use a custom service :
+Alternatively, you can use a service to retrieve the groups to use:
+
 ```php
 <?php
 
-// src/AppBundle/Service/ValidatorGroupsGenerator.php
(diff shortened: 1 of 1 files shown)


dunglas:
I made some slight changes. Can you double check?

core/validation.md
```diff
@@ -95,6 +95,70 @@ class Book
...
+
+```php
+<?php
+
+// src/AppBundle/Entity/Book.php
+
+namespace AppBundle\Entity;
+
+use ApiPlatform\Core\Annotation\ApiResource;
+use AppBundle\Validator\AdminGroupsGeneratorGroupsGenerator;
```
soyuka: Hmm is it not `+use AppBundle\Validator\AdminGroupsGenerator`?

Simperfit: You are right.

meyerbaptiste:
Thanks @pierre-H!

PR #262, opened 2017-09-11
Update all service configs according to changes in 2.1

dunglas:


meyerbaptiste:
Some changes are missing in:
* core/operation-path-naming.md
* core/extensions.md

dunglas:
@meyerbaptiste thanks, fixed

PR #272, opened 2017-09-19
Fix role in code example

seza443:
Doc says

> only admins are able to create books

But in the code, the role is `ROLE_USER` for the `POST` operation. It should be `ROLE_ADMIN`


dunglas:
Good catch! Can you rebase against the `2.1` branch please?

seza443:
Done 😄

soyuka:
Thanks @seza443

PR #284, opened 2017-09-22
Fixed heroku tutorial for api platform 2.1+

mickaelandrieu:
Hello,

well... this article is really outdated and I've spent so MUCH time figure out how to make this deployement works as expected.

Luckely, I'll be the last one \o/

@dunglas I'll removed the installation of your package: once it will be tagged as stable, we may update again this article 👍

### DIFF
--- a/deployment/heroku.md
+++ b/deployment/heroku.md
@@ -85,7 +85,10 @@ using the external parameters feature of the Symfony container:
     {
       "url": "https://github.com/heroku/heroku-buildpack-php"
     }
-  ]
+  ],
+  "scripts": {
+    "postdeploy": "php bin/console doctrine:schema:create"
+  }
 }
 ```
 
@@ -101,7 +104,7 @@ directories must be private.
 Create a new file named `Procfile` at the root of the application directory with the following content:
 
 ```yaml
-web: bin/heroku-php-apache2 web/
+web: vendor/bin/heroku-php-apache2 web/
 ```
 
 Our application is ready to be deployed, but Heroku dynos are not persistent and file stored directly on the filesystem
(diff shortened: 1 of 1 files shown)


deployment/heroku.md
```diff
@@ -85,22 +47,65 @@ using the external parameters feature of the Symfony container:
...
+
+```php
+<?php
+
+namespace AppBundle\Heroku;
+
+/**
+ * Setup database configuration from environment variables (Heroku deployment)
+ */
+class Database
```
dunglas: Can't we use directly the URL (now that it is possible) instead of introducing this class?
Just like here: https://github.com/dunglas/api-platform/blob/flex-js/api/config/packages/doctrine.yaml#L3

mickaelandrieu: Hello,  thanks for the review.

I have 2 questions:

* Is this behavior (already) available in 2.1?
* Is dbal automatically "fill" all the database parameters if you use ``DATABASE_URL`` environment variable?

dunglas: It’s a feature of Doctrine bundle and it’s available for a while. I don’t think DoctrineBundle populates other parameters IMO it deals directly with the URL.

mickaelandrieu: Ok, I remove this part. For the rest of the article, is is good? :)

dunglas: Yes perfect! It will be way easier with the new distribution!

dunglas:
Can you rebase this PR @mickaelandrieu?

mickaelandrieu:
@dunglas should be good now

dunglas:
Thanks @mickaelandrieu

PR #290, opened 2017-09-27
Update getting-started.md

davidpodhola:
Following https://github.com/api-platform/admin#usage

Sorry, having just browser now it is easier this way. Thanks for understanding.


dunglas:
Thank you very much!

PR #295, opened 2017-09-28
Update admin authentification

pierre-H:
Add https://github.com/api-platform/admin/pull/51 changes and change the request for compatibility with LexikJWTAuthenticationBundle (tested and works)

### DIFF
--- a/admin/authentication-support.md
+++ b/admin/authentication-support.md
@@ -15,11 +15,15 @@ export default (type, params) => {
   switch (type) {
     case AUTH_LOGIN:
       const { username, password } = params;
-      const request = new Request(`${entrypoint}/login_check`, {
-        method: 'POST',
-        // the next two lines are compatible with LexikJWTAuthenticationBundle. Change this if you use something else ...
-        body: `_username=${username}&_password=${password}`,
-        headers: new Headers({ 'Content-Type': 'application/x-www-form-urlencoded' }),
+      const request = new Request(`${API_ENTRYPOINT}/login`, {
+          body: JSON.stringify({
+              username: params.username,
+              password: params.password,
+          }),
+          headers: new Headers({
+              'Content-Type': 'application/json',
+          }),
+          method: 'POST',
       });
 
       return fetch(request)
--- a/admin/authentication-support.md
+++ b/admin/authentication-support.md
@@ -30,6 +30,7 @@ export default (type, params) => {
         })
         .then(({ token }) => {
           localStorage.setItem('token', token); // The JWT token is stored in the browser's local storage
+          window.location.replace('/');
         });
 
     case AUTH_LOGOUT:
--- a/admin/authentication-support.md
+++ b/admin/authentication-support.md
@@ -17,8 +17,9 @@ export default (type, params) => {
       const { username, password } = params;
       const request = new Request(`${entrypoint}/login_check`, {
         method: 'POST',
-        body: JSON.stringify({ email: username, password }),
-        headers: new Headers({ 'Content-Type': 'application/json' }),
+        // the next two lines are compatible with LexikJWTAuthenticationBundle. Change this if you use something else ...
+        body: `_username=${username}&_password=${password}`,
+        headers: new Headers({ 'Content-Type': 'application/x-www-form-urlencoded' }),
       });
 
       return fetch(request)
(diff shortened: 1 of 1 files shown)


admin/authentication-support.md
```diff
@@ -17,8 +17,9 @@ export default (type, params) => {
       const { username, password } = params;
       const request = new Request(`${entrypoint}/login_check`, {
         method: 'POST',
-        body: JSON.stringify({ email: username, password }),
-        headers: new Headers({ 'Content-Type': 'application/json' }),
+        // the next two lines are compatible with LexikJWTAuthenticationBundle. Change this if you use something else ...
+        body: `_username=${username}&_password=${password}`,
+        headers: new Headers({ 'Content-Type': 'application/x-www-form-urlencoded' }),
```
soyuka: Why can't we keep json login? It has nothing to do with `LexikJWTAuthentication` [see here](https://symfony.com/doc/current/security/json_login_setup.html)

admin/authentication-support.md
```diff
@@ -52,24 +53,60 @@ export default (type, params) => {
...
-  }
-}
+const fetchHydra = (url, options = {}) => baseFetchHydra(url, {
+    ...options,
+    headers: new Headers(fetchHeaders),
+});
+
+const restClient = api => hydraClient(api, fetchHydra);
+
+const apiDocumentationParser = entrypoint => parseHydraDocumentation(entrypoint, { headers: new Headers(fetchHeaders) })
```
soyuka: We really should not need this :p. maybe we can path headers to the react component?

admin/authentication-support.md
pierre-H: I just tried in json and lexik return 401 bad credentials :
```shell
curl -X POST \
  http://localhost/api/web/login_check \
  -H 'content-type: application/json' \
  -d '{"_username": "admin", "_password": "admin"} '
```
Only with this configuration I can login. Maybe it's my config but I only have this :
```yaml
lexik_jwt_authentication:
    private_key_path: '%jwt_private_key_path%'
    public_key_path:  '%jwt_public_key_path%'
    pass_phrase:      '%jwt_key_pass_phrase%'
    token_ttl:        '%jwt_token_ttl%'
```

admin/authentication-support.md
mysiar: I would rather keep it simplest so anybody can implement their own login and authentication with api. I use token passed in header to backend

admin/authentication-support.md
pierre-H: So maybe you're right. I thought that because the beginning of this page says that we assume that the api is secure with jwt (with a link to the doc which explains how to use LexikJWTAuthentication) so I thought that this page has to explain how to use admin with LexikJWTAuthentication but if you think that it's should be general ...

dunglas: I agree with @pierre-H. We should update the doc of the Lexik bundle to explain how to allow JSON login or keep with form-data. (it would be great if we can use JSON everywhere).

dunglas:
@mauchede would you mind reviewing this PR?

mauchede:
@pierre-H: Can you add `window.location.replace('/');` in the `AUTH_LOGIN` action? Without this refresh, the API documentation will be empty.

For example:

```js
return fetch(request)
	.then(response => {
	    if (response.status < 200 || response.status >= 300) {
		throw new Error(response.statusText);
	    }

	    return response.json();
	})
	.then(({ token }) => {
	    window.localStorage.setItem(LOCAL_STORAGE_KEY_TOKEN, token);
	    window.location.replace('/');
	});
```

admin/authentication-support.md
pierre-H: Does someone have an idea how to use Lexik bundle with json ?

pierre-H:
@mauchede done

admin/authentication-support.md
mauchede: @pierre-H : Lexik provides some handlers. You just have to use these handlers with the new [JSON authenticator](http://symfony.com/blog/new-in-symfony-3-3-json-authentication).

__Note__: The JSON authenticator is available since Symfony `3.3`.

In `app/config/security.yml`, you will write something like this:

```yml
security:
# ...
    firewalls:
        login:
            anonymous: true
            json_login:
                check_path: /login
                failure_handler: lexik_jwt_authentication.handler.authentication_failure
                success_handler: lexik_jwt_authentication.handler.authentication_success
# ...
```

In `admin`, you can now directly send JSON to the API:

```js
            const request = new Request(`${API_ENTRYPOINT}/login`, {
                body: JSON.stringify({
                    username: params.username,
                    password: params.password,
                }),
                headers: new Headers({
                    'Content-Type': 'application/json',
                }),
                method: 'POST',
            });

            return fetch(request).then( /* ... */);
```

In my repositories, you can find an example [secured API](https://github.com/mauchede/example-api-platform-api-platform/tree/example/security) and an example of [admin](https://github.com/mauchede/example-api-platform-admin/tree/example/security).

pierre-H:
Thank you @mauchede  for your explanation. I change the doc so we use json instead of form data

Simperfit:
@mauchede is it good for you ?

Simperfit:
Thanks @pierre-H

PR #296, opened 2017-09-29
Authentication Documentation improvements

sh41:
Have corrected a few typos, updated examples to match current function signatures and clarified the source of the path used to login.


dunglas:
Great! Can you just rebase against 2.1 please?

sh41:
I'm not sure I've rebased this correctly as I've managed to pick up a commit from a previous PR, that was already merged into master. Is this in a suitable state to merge or is there more I need to do?

PR #311, opened 2017-10-09
Update react.md

mysiar:
Bootstrap 3 & 4


dunglas:
I would only document Bootstrap 4. WDYT?

mysiar:
so maybe we need another document file like history or changelog to keep track this things ?
with the new release Bootstrap 3 will be gone unless somebody explicitly runs 0.1.15 or earlier

dunglas:
Thanks @mysiar

PR #316, opened 2017-10-15
Fixed ApiNormalizer config, was giving me an error with default install

Bwen:
While following this page I was getting the following error with default APIP setup:

> Symfony\Component\Debug\Exception\FatalThrowableError: Call to a member function normalize() on null in /srv/api-platform/vendor/api-platform/core/src/Serializer/AbstractItemNormalizer.php:423 Stack trace: 
> #0 /srv/api-platform/vendor/symfony/symfony/src/Symfony/Component/Serializer/Normalizer/AbstractObjectNormalizer.php(78)

### DIFF
--- a/core/serialization.md
+++ b/core/serialization.md
@@ -42,7 +42,7 @@ services:
   # ...
 
     'AppBundle\Serializer\ApiNormalizer':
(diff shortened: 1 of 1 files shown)


Bwen:
Never mind, now its just not serializing. Sorry.

PR #319, opened 2017-10-16
Update handling-relations-to-collections.md

davidpodhola:
Changes necessary to make it work on the current version also see https://github.com/api-platform/admin/issues/62#issuecomment-336856775

Probably more to come as the one-to-may is still not working, but these changes are needed for sure now.


davidpodhola:
So at the end I think the change is enough see https://github.com/api-platform/admin/issues/62#issuecomment-336872779

dunglas:
Thanks @davidpodhola. I made some minor CS changes.

PR #320, opened 2017-10-16
Update BookSpecial class

Saphyel:
You can define the methods inside of the Route annotation


dunglas:
It was done on purpose to make it explicit that FrameworkExtraBundle can be used with API Platform. But I also prefer to not use it when possible 👍 to merge this.

dunglas:
Thanks @Saphyel

PR #327, opened 2017-10-31
Admin On REST

mysiar:


dunglas:
I think it deserves an in depth explanation of how it is better to use `admin` and how it is better to use the generator. And how to switch for the one to the other. Or it will be very confusing for the user to have 2 tools doing the same thing.

mysiar:
OK. I will try to do it more explicit

mysiar:
@dunglas have a look now

mauchede:
As `hydraClient` is needed by `admin-on-rest`, do you think we have to recommend `@api-platform/admin` as dependency?

mysiar:
@mauchede this is actually very good point :)
I will look into it and update docs

mysiar:
The only thing that I do not see in hydraClient is possibility to add auth token

mauchede:
`hydraClient` does not manage authentication token, but `fetchHydra` does. If you wish to define a token, you have to initialize `fetchHydra`:

```js
import { fetchHydra as baseFetchHydra, hydraClient as baseHydraClient } from '@api-platform/admin';

export const fetchHeaders = {
    'Authorization': `Bearer ${window.localStorage.getItem('token')}`,
};

export const fetchHydra = (url, options = {}) => baseFetchHydra(url, {
    ...options,
    headers: new Headers(fetchHeaders),
});

export default api => baseHydraClient(api, fetchHydra);
```

mysiar:
I'm not sure if hydraClient can be used, there is no need to parse anything, resources are generated

mysiar:
Another solution would to take the client out of doucmentation and add it to the templates as util
WDYT?

dunglas:
It needs a rebase.

mysiar:
@dunglas rebase done

Simperfit:
it needs a rebase again

mysiar:
looks like this branch is already to old, should I close it and move doc files to 2.3 instead and create new PR ?

mysiar:
new PR to 2.3 branch

PR #333, opened 2017-11-06
fix code filed

cr3a7ure:
code field*

### DIFF
--- a/core/default-order.md
+++ b/core/default-order.md
@@ -61,7 +61,7 @@ class Book
 
 It's also possible to configure the default filter on an association property:
 
-<?php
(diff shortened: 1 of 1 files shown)


core/default-order.md
```diff
@@ -61,7 +61,7 @@ class Book
 
 It's also possible to configure the default filter on an association property:
 
-<?php
+```<?php
```
dunglas: It must like the following to be correctly highlighted. I also think that it should target the 2.1 branch. 

    ```php
    <?php

dunglas:
Fixed in #337. Thanks.

PR #343, opened 2017-11-16
Update "Using Custom Components" section

acidline:
Exemple is adapted from new implementation by @mauchede ["Customize CRUD Component"](https://github.com/api-platform/admin/pull/63)

### DIFF
--- a/admin/getting-started.md
+++ b/admin/getting-started.md
@@ -56,27 +56,31 @@ import parseHydraDocumentation from 'api-doc-parser/lib/hydra/parseHydraDocument
 
 const entrypoint = 'https://demo.api-platform.com';
 
-const apiDocumentationParser = entrypoint => parseHydraDocumentation(entrypoint)
-  .then(api => {
-    api.resources.map(resource => {
-      const books = api.resources.find(r => 'books' === r.name);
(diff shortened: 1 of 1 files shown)


admin/getting-started.md
```diff
@@ -56,27 +56,31 @@ import parseHydraDocumentation from 'api-doc-parser/lib/hydra/parseHydraDocument
...
+
+    const books = api.resources.find(({ name }) => 'books' === name);
+    const description = books.fields.find(f => 'description' === f.name);
+    
+    description.field = props => (
+        <RichTextField {...props} source="description" />
+    );
+    
+    description.input = props => (
+        <RichTextInput {...props} source="description" addLabel={true} />
```
mauchede: When I tried it did not work because `addLabel` and `addField` have to be declared in the `input` context (and not in the rendered component context). The following code will work:

```js
description.input = props => (
    <RichTextInput {...props} source="description" />
);

description.input.defaultProps = {
        addField: true,
        addLabel: true,
};
```

dunglas:
It needs a rebase :)

dunglas:
Thanks @acidline!

PR #344, opened 2017-11-16
Fix the ApiNormalizer decorator

meyerbaptiste:


core/serialization.md
```diff
@@ -417,19 +417,23 @@ services:
...
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
-final class ApiNormalizer extends AbstractItemNormalizer
+final class ApiNormalizer implements NormalizerInterface, DenormalizerInterface
 {
     private $decorated;
 
     public function __construct(NormalizerInterface $decorated)
     {
+        if (!$decorated instanceof DenormalizerInterface) {
```
acidline: Is not "NormalizerInterface" instead of "DenormalizerInterface" ?

dunglas: I think he wants to be sure that the current class implement both interfaces. We're sure it's a `NormalizerInterface` instance because of the typehint.

dunglas:
Thanks @meyerbaptiste

core/serialization.md
acidline: That's right. Thank's for the explication.

PR #346, opened 2017-11-17
Removed non existing configuration key

Nek-:
I don't know where comes this key from. But it looks like this is not valid with api-platform...


meyerbaptiste:
Thx @Nek-

PR #348, opened 2017-11-22
Swagger ApiProperty variable should not be quoted

viviengaetan:
When copying/pasting the annotation from the current doc we got ```[Syntax Error] Expected Doctrine\Common\Annotations\DocLexer::T_IDENTIFIER, got 'attributes' at position 48 in property AppBundle\Entity\UserPortfolioAudit::$pfType.``` because attributes should not be quoted.

### DIFF
--- a/core/swagger.md
+++ b/core/swagger.md
@@ -53,10 +53,10 @@ final class SwaggerDecorator implements NormalizerInterface
 
 
 	// e.g. add a custom parameter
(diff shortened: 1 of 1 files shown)


dunglas:
Good catch, thanks @viviengaetan

viviengaetan:
No problem - anyway we are glad to see a new documentation, one of our developer was grumpy when reading the old one ;)

PR #350, opened 2017-11-24
Added documentation for group contexts on item base

Toflar:
I had to introduce permissions on item base so I thought I'd document this for others having the same issue :-) Maybe there's even a better way but that's how I did it and got it to work.

### DIFF
--- a/core/serialization.md
+++ b/core/serialization.md
@@ -383,6 +383,73 @@ final class BookContextBuilder implements SerializerContextBuilderInterface
 If the user has `ROLE_ADMIN` permission and the subject is an instance of Book, `admin_input` group will be dynamically added to the denormalization context.
 The variable `$normalization` lets you check whether the context is for normalization (if true) or denormalization.
 
+## Changing the Serialization Context on a per item base
+
+The example above shows how you can modify the normalization/denormalization context based on the current user permissions for all the books that are being normalized/denormalized. Sometimes, however, the permissions vary depending on what book is being processed. Think of ACL's: User A may retrieve Book A but not Book B. In that case, we have to leverage the power of the Symfony Serializer and register our own normalizer that adds the group on every single item (priority `64` is just an example, make sure your normalizer gets loaded first):
+
+```yaml
+# app/config/services.yml
+services:
+    'AppBundle\Serializer\BookAttributeNormalizer':
+        arguments: [ '@security.token_storage' ]
+        tags:
+            - { name: 'serializer.normalizer', priority: 64 }
+```
+
+The Normalizer class is a bit harder to understand because it has to be aware of the Serializer instance itself to pass on the object to normalize to the other normalizer once it added the groups and at the same time it has to make sure, there's no recursion. Here's how this could look like:
+
+```php
+<?php
+// src/AppBundle/Serializer/BookAttributeNormalizer.php
+
+class BookAttributeNormalizer implements NormalizerInterface, SerializerAwareInterface
+{
+    use SerializerAwareTrait;
+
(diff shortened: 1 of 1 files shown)


core/serialization.md
```diff
@@ -383,6 +383,73 @@ final class BookContextBuilder implements SerializerContextBuilderInterface
...
+    public function __construct(TokenStorageInterface $tokenStorage)
+    {
+        $this->tokenStorage = $tokenStorage;
+    }
+
+    public function normalize($object, $format = null, array $context = [])
+    {
+        // Get permissions from user in $this->tokenStorage
+        // for the current $object (book)
+        if ($this->userHasPermissionsForBook($object)) {
```
vincentchalamon: Method `userHasPermissionsForBook` is missing

core/serialization.md
```diff
@@ -383,6 +383,73 @@ final class BookContextBuilder implements SerializerContextBuilderInterface
...
+
+```php
+<?php
+// src/AppBundle/Serializer/BookAttributeNormalizer.php
+
+class BookAttributeNormalizer implements NormalizerInterface, SerializerAwareInterface
+{
+    use SerializerAwareTrait;
+
+    const BOOK_ATTRIBUTE_NORMALIZER_ALREADY_CALLED = 'BOOK_ATTRIBUTE_NORMALIZER_ALREADY_CALLED';
```
vincentchalamon: Please specify the publicity of this constant: `private` or `public`

core/serialization.md
```diff
@@ -383,6 +383,73 @@ final class BookContextBuilder implements SerializerContextBuilderInterface
...
+    private $tokenStorage;
+
+    public function __construct(TokenStorageInterface $tokenStorage)
+    {
+        $this->tokenStorage = $tokenStorage;
```
vincentchalamon: If you wanna check user permissions, it's safer to use the `AuthorizationChecker`

core/serialization.md
```diff
@@ -383,6 +383,73 @@ final class BookContextBuilder implements SerializerContextBuilderInterface
 If the user has `ROLE_ADMIN` permission and the subject is an instance of Book, `admin_input` group will be dynamically added to the denormalization context.
 The variable `$normalization` lets you check whether the context is for normalization (if true) or denormalization.
 
+## Changing the Serialization Context on a per item base
```
alanpoulain: basis?

core/serialization.md
```diff
@@ -383,6 +383,73 @@ final class BookContextBuilder implements SerializerContextBuilderInterface
...
+```yaml
+# app/config/services.yml
+services:
+    'AppBundle\Serializer\BookAttributeNormalizer':
+        arguments: [ '@security.token_storage' ]
+        tags:
+            - { name: 'serializer.normalizer', priority: 64 }
+```
+
+The Normalizer class is a bit harder to understand because it has to be aware of the Serializer instance itself to pass on the object to normalize to the other normalizer once it added the groups and at the same time it has to make sure, there's no recursion. Here's how this could look like:
```
alanpoulain: it has to make sure there's no recursion

core/serialization.md
Toflar: In my case it's just checking whether there's a relation in the DB or not ;-) But I think this is something that's documented on the Symfony side or how would you add that to the docs?

vincentchalamon: It's documented in Symfony docs, but you could add the link of it to help people :-)

Toflar: Done in https://github.com/api-platform/docs/pull/350/commits/bcc315428c3b453294a2ca47a54f0ea32a890187 :)

dunglas:
Thanks @Toflar!

PR #352, opened 2017-11-30
Fix default Kernel location

pyrech:


sroze:
Isn't it `src/Kernel.php` in Symfony4?

meyerbaptiste:
Right, but 2.1 doesn't use SF4. We have to change this during the merge in `master`.

soyuka:
Thanks @pyrech

PR #353, opened 2017-12-02
Fix missing serializer in decorated normalizer

beleneglorion:
When decorated the serializer is not set in the .inner service  anymore ?
Adding the SerializerAwareInterface on the decorator and passing serializer to the decorated one solve the problem.


dunglas:
Good catch, thanks for the fix!

PR #355, opened 2017-12-04
Fix autowiring deprecation notice

norkunas:


dunglas:
Good catch, thanks @norkunas

PR #359, opened 2017-12-07
Feat missing api receive

### LINKED ISSUE #357: Question: Why is `_api_receive` not documented
Sorry I have a lot of questions these days, but I am really looking forward to upgrade a big project from api-platform v1 to v2 and I have a lot of custom logic.

To migrate easily, I need a way to acces the original data fetched from the persistence layer  in my controller. 

First I created a EventSubscriber cloning the `$request->attributes->get('data')`. It works fine for "simple" cases, like controller access before updating entity (like "you can only change ownership of an item if you own the item"), but sometimes I need more control between the entity and the request content.

I found the `_api_receive` argument to be a very good candidate in blocking the DeserializeListener to do it's job, but this feature is undocumented. (and it unfortunatly also skip the validation process).

Is it a "private" / internal feature or is just missing from the documentation ? (if so, I can make a PR of course)

jdeniau:
Fixes #357

### DIFF
--- a/core/operations.md
+++ b/core/operations.md
@@ -558,7 +558,7 @@ API Platform automatically retrieves the appropriate PHP entity then deserialize
 updates the entity with data provided by the user.
 
 If you want to bypass the automatic retrieval of the entity, you can set the parameter `_api_receive` to `false`. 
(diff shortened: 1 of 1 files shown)


jdeniau:
Done @dunglas

dunglas:
Thanks @jdeniau

PR #361, opened 2017-12-09
Fixed function name to match what was intended by the example.

BallisticPain:


dunglas:
Good catch, thanks for proofreading!

PR #369, opened 2017-12-26
Added basic doc for dealing with DTO

lyrixx:
Fixed https://github.com/api-platform/core/issues/1590

### DIFF
--- a/core/dto.md
+++ b/core/dto.md
@@ -1,0 +1,114 @@
+# Handling Data Transfer Objects (DTOs)
+
+## How to use a DTO for Writing
+
+Sometimes it's easier to use a DTO than an Entity when performing simple
+operation. For example, the application should be able to send an email when
+someone has lost its password.
+
+So let's create a basic DTO for this request:
+
+```php
+namespace App\Api\Dto;
+
+use ApiPlatform\Core\Annotation\ApiResource;
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * @ApiResource(
+ *      collectionOperations={
+ *          "post"={
+ *              "method"="POST",
+ *              "path"="/users/forgot-password-request",
+ *          },
+ *      },
+ *      itemOperations={},
+ * )
+ */
+final class ForgotPasswordRequest
+{
+    /**
+     * @Assert\NotBlank()
+     * @Assert\Email()
+     */
+    public $email;
+}
+```
+
+In this case, we disable all operations except `POST`.
+
+Then, thanks to [the event system](events.md), it's possible to intercept the
+`POST` request and to handle it.
+
+First, an event subscriber is needed:
+
+```php
+<?php
+
+namespace App\Api\EventSubscriber;
+
+use ApiPlatform\Core\EventListener\EventPriorities;
+use App\Entity\User;
+use App\Manager\UserManager;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpKernel\Event\GetResponseForControllerResultEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+final class UserSubscriber implements EventSubscriberInterface
+{
+    private $userManager;
+
+    public function __construct(UserManager $userManager)
+    {
+        $this->userManager = $userManager;
+    }
+
+    public static function getSubscribedEvents()
+    {
+        return [
+            KernelEvents::VIEW => ['sendPasswordReset', EventPriorities::POST_VALIDATE],
+        ];
+    }
+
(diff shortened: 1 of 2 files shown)


core/dto.md
```diff
@@ -0,0 +1,116 @@
...
+First, an event subscriber is needed:
+
+```php
+<?php
+
+namespace App\Api\EventSubscriber;
+
+use ApiPlatform\Core\EventListener\EventPriorities;
+use App\Manager\UserManager;
+use Joli\Entity\User;
```
dunglas: `App\Entity\User`?

core/dto.md
```diff
@@ -0,0 +1,116 @@
+# DTO - Data Transfer Object
+
+## How to use a DTO for writing
+
+Sometimes it's easier to use a DTO than an Entity when performing simple that.
```
dunglas: Extra "that" to remove?

core/dto.md
```diff
@@ -0,0 +1,116 @@
+# DTO - Data Transfer Object
```
dunglas: May I suggest: `Handling Data Transfer Objects (DTOs)`

core/dto.md
```diff
@@ -0,0 +1,116 @@
...
+
+## How to use a DTO for writing
```
dunglas: We use title case for titles: `How to use a DTO for Writing`

core/dto.md
```diff
@@ -0,0 +1,116 @@
...
+
+Sometimes it's easier to use a DTO than an Entity when performing simple that.
+For exemple, if the application should be able to send an email when someone has
```
dunglas: example

core/dto.md
```diff
@@ -0,0 +1,116 @@
...
+{
+    /**
+     * @Assert\NotBlank()
+     * @Assert\Email()
+     */
+    public $email;
+}
+```
+
+In this case, we disable all operation exept `POST`.
```
dunglas: operation**s** ex**c**ept

core/dto.md
```diff
@@ -0,0 +1,116 @@
...
+
+Then, thanks to the [the event system](events.md), it's possible to intercept
```
dunglas: double `the`

core/dto.md
```diff
@@ -0,0 +1,116 @@
...
+
+    public function __construct(UserManager $userManager)
+    {
+        $this->userManager = $userManager;
+    }
+
+    public static function getSubscribedEvents()
+    {
+        return [
+            KernelEvents::VIEW => [
```
dunglas: IIRC, you can remove the outer array.

core/dto.md
alanpoulain: except

core/dto.md
alanpoulain: example

core/dto.md
alanpoulain: that? :p

core/dto.md
```diff
@@ -0,0 +1,116 @@
...
+        if ($user) {
+            $this->userManager->requestPasswordReset($user);
+        }
+
+        $event->setResponse(new JsonResponse(null, 204));
+    }
+}
+```
+
+Then this class should be registred as a service, then tagged.
```
alanpoulain: registered

core/dto.md
lyrixx: yes it's possible. But usually the listener can be re-used to handle many operation.
And it could avoid https://github.com/api-platform/core/issues/1590#issuecomment-353392828

Anyway, I removed the outer array

lyrixx:
@alanpoulain & @dunglas Thanks for the review. I have addressed your comments.

dunglas:
Thanks @lyrixx! Great piece of docs.

aitboudad:
added one missing step #371

core/dto.md
```diff
@@ -0,0 +1,116 @@
...
+        return [
+            KernelEvents::VIEW => ['sendPasswordReset', EventPriorities::POST_VALIDATE],
+        ];
+    }
+
+    public function sendPasswordReset(GetResponseForControllerResultEvent $event)
+    {
+        $request = $event->getRequest();
+
+        if ('api_forgot_password_requests_post_collection' !== $request->attributes->get('_route')) {
```
aitboudad: wouldn't better to check the model instance instead ?
```php
if (
    !Request::METHOD_POST === $request->getMethod()
    || !$event->getControllerResult() instanceof ForgotPasswordRequest
) {
    return;
}
```

lyrixx: Actually, Now I prefer to match the route name. This reduce side effect.

But anyway, It's not really perfect IMHO

aitboudad: ahh ok, IMO there is no perfect way ;)

PR #370, opened 2017-12-27
Doc. how to change the path of a subresource

lyrixx:
Fixes https://github.com/api-platform/core/issues/1596


dunglas:
Thanks @lyrixx

PR #372, opened 2017-12-28
Run Proselint in the CI and fix some errors

dunglas:
The documentation is now automatically checked with [Proselint](https://github.com/amperser/proselint).

* `conduct.md` file is ignored because it contains [Contributor Covenant](https://www.contributor-covenant.org/): we don't want to change this text
* `heroku.md` is ignored because of a false positive: I've found no easy way to ignore false positives using proselint

### DIFF
--- a/.proselintrc
+++ b/.proselintrc
@@ -1,0 +1,7 @@
+{
+  "checks": {
+    "typography.symbols": false,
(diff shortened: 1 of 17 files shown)


alanpoulain:
Proselint really doesn't like superlatives!

PR #374, opened 2018-01-03
Grammar fix

bdunogier:
Not much to add :)


sroze:
Good catch, thanks @bdunogier

dunglas:
Thanks @bdunogier

PR #377, opened 2018-01-23
Documenting how to have swagger on custom uri

nWidart:
This adds some documentation on how to keep the api on one endpoint, and move swagger ui to a different one.

### DIFF
--- a/core/swagger.md
+++ b/core/swagger.md
@@ -179,3 +179,27 @@ Will produce the following Swagger documentation:
   }
 }
 ```
+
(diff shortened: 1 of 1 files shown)


nWidart:
Thank you for the review. Sorry for the inconsistencies. 👍

dunglas:
Thanks @nWidart, I changed the file name according to API Platform 2.1 (we'll take care of updating the paths to match the Flex structure for all the docs when we'll release 2.2).

PR #388, opened 2018-02-08
Fixed Authentification to Authentication

nepda:
https://english.stackexchange.com/questions/7844/is-authentification-a-real-word


dunglas:
Good catch ("authentification" is a french, and as you can see some docs authors are French 😄).

PR #400, opened 2018-02-20
Update LexikJWTAuthenticationBundle configuration

antograssiot:
Using this configuration is not the recommended way anymore. It triggers 4 deprecation including `"lexik_jwt_authentication.security.authentication.entry_point.main" service is deprecated`


dunglas:
Thanks @antograssiot

PR #404, opened 2018-02-23
Extending Cache-Tags docs for varnish

iCodr8:
See: https://github.com/api-platform/api-platform/issues/580


Simperfit:
Thanks @iCodr8.

PR #405, opened 2018-02-23
Document the release policy

dunglas:
The important part:

* **stable** (currently the **2.2** branch): regular bug fixes are integrated in this version
* **old-stable** (currently **2.1** branch): security fixes are integrated in this version, regular bug fixes are **not** backported in it
* **development** (**master** branch): new features target this branch

 Older versions (1.x, 2.0...) **are not maintained**. If you still use them, you must upgrade as soon as possible.

ping @api-platform/core-team

### DIFF
--- a/extra/releases.md
+++ b/extra/releases.md
@@ -1,0 +1,16 @@
+# The Release Process
+
+API Platform follows the [Semantic Versioning](https://semver.org) strategy.
+
+Only 3 versions are maintained at the same time:
+
+* **stable** (currently the **2.2** branch): regular bug fixes are integrated in this version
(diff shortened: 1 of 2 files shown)


joelwurtz:
What about the timing of releases ? Can we expect consistent time between each releases or will this mainly depend on contribution activity ?

dunglas:
@joelwurtz I've written this in the docs: `New versions of API Platform are released **when they are ready**, on the behalf of the API Platform Core Team.`

So it will mainly depend on contribution activity for now. We may change this later. My current idea is to release a new version minor version of API Platform ~1 month after the release of every minor Symfony version (so every 6 months). But I don't want to enforce this hardly for now, we may release versions more frequently if the activity level is high.

PR #407, opened 2018-02-24
Describe the fact that first GET collectionOperation is used for resourceClass IRI

benjaminrau:
Users sometimes create custom collection operation and add it as first item of collectionOperation but dont do that with the intention to overwrite the default resource Class IRI. If the route path contains custom route params this results in an error though and debugging it is quite hard (need do make your way down into IriConverter to discover the custom route is used for router).


benjaminrau:
There might be different reasons for the issue users face - i have doubts my docs PR is sufficient. At least it doesnt match this issue: https://github.com/api-platform/core/issues/830

dunglas:
Thanks @benjaminrau. I just remove the "for performance" part because it's not really the reason, it's just that this behavior sounds "logical" and intented to me.

PR #413, opened 2018-02-28
Minor changes to the Swagger documentation page

metaturso:
> The following configuration will provide additional context to your Swagger definitions
- In api-platform ^2.1 this is not the case. Defining a custom `swagger_context` has the effect of wiping out any default documentation. Therefore this wasn't being treated as "additional context" and should not be referred as such.

> docker-compose exe php
- Fixed the wrong docker compose command "exe" instead of "exec".


dunglas:
Good catch, thanks @metaturso

PR #423, opened 2018-03-07
Fix RestrictedDataProviderInterface code examples

teohhanhui:
Method signature was changed before release?


dunglas:
Indeed is has changed before release. Good catch!

PR #427, opened 2018-03-12
Fix reference to wrong Doctrine namespace.

BonnieDoug:
Originally references the RangeFilter when it should be the OrderFilter.


dunglas:
Good catch, thanks @BonnieDoug

PR #441, opened 2018-03-20
Filters - consistency on usage example + date example

BatsaxIV:
Hi,

This is a proposal to add an example on the date filter and change some usages example, for consistency, with the (already used) syntax : `<choice1|choice2|choice3>`


dunglas:
Thanks!

Issue #442, opened 2018-03-20
Schema Generator > Getting Started


alexislefebvre:
I followed the documentation at https://api-platform.com/docs/schema-generator/getting-started#model-scaffolding

I see a warning when launching the command:

```bash
$ vendor/bin/schema generate-types api/src/ api/config/schema.yaml
[warning] The property "address" (type "Person") has several types. Using the first one ("PostalAddress") or possible options("PostalAddress", "Text").
```

And a `Thing.php` file is created even if it's not mentioned in the documentation.

```bash
$ tree api/                                                                                                                                                                     0 < 15:50:20
api/
├── config
│   └── schema.yaml
└── src
    └── AppBundle
        └── Entity
            ├── Person.php
            ├── PostalAddress.php
            └── Thing.php

4 directories, 4 files
```

--- 

So I think that there are several problems on the page:

1. The command generates a warning
2. `Thing` is shown in the first YAML file only but are not shown in the second YAML file (_The following classes will be generated:_) and after content of `Person.php` and `PostalAddress.php` files

alexislefebvre:
Now I understand that `Thing` come from the Schema, I'll open a PR with a suggestion to avoid possible confusion.

PR #443, opened 2018-03-20
“Schema generator“ > improve “Getting started”

alexislefebvre:
See #442


schema-generator/getting-started.md
```diff
@@ -30,13 +35,7 @@ types:
             familyName: ~
             givenName: ~
             additionalName: ~
-            gender: ~
             address: ~
-            birthDate: ~
-            telephone: ~
-            email: ~
-            url: ~
-            jobTitle: ~
```
meyerbaptiste: Maybe I missed something but why do you remove these properties?

alexislefebvre: I think that we don't need 10 properties in order to understand this code sample. I make the generated code longer and it adds no value.

dunglas: Agreed.

meyerbaptiste: Makes sense.

PR #445, opened 2018-03-22
Remove obsolete informations

gorghoa:
admin-on-rest has moved react and react-dom to its **dev** dependencies since https://github.com/marmelab/admin-on-rest/commit/9836464ca275b1fbf84d932e27b4d7f3fb35d374#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L66

Following this step from the getting-started guide leads to compilation error (module missing from node_modules)

### DIFF
--- a/admin/getting-started.md
+++ b/admin/getting-started.md
@@ -15,14 +15,6 @@ Now, go to the newly created `my-admin` directory:
 
     $ cd my-admin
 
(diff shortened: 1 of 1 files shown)


gorghoa:
thanks for your review dear fellows :)

dunglas:
Thanks @gorghoa

PR #446, opened 2018-03-22
[data-providers] - document the SubresourceDataProviderInterface usage

gorghoa:
Document the SubresourceDataProviderInterface in custom context.

### DIFF
--- a/core/data-providers.md
+++ b/core/data-providers.md
@@ -16,8 +16,10 @@ For a given resource, you can implement two kind of interfaces:
   is used when fetching a collection.
 * the [`ItemDataProviderInterface`](https://github.com/api-platform/core/blob/master/src/DataProvider/ItemDataProviderInterface.php)
   is used when fetching items.
+* the [`SubresourceDataProviderInterface`](https://github.com/api-platform/core/blob/master/src/DataProvider/SubresourceDataProviderInterface.php)
+  is used when fetching subresource items.
 
-Both implementations can also implement a third, optional interface called
+All three implementations can also implement a fourth, optional interface called
 ['RestrictedDataProviderInterface'](https://github.com/api-platform/core/blob/master/src/DataProvider/RestrictedDataProviderInterface.php)
 if you want to limit their effects to a single resource or operation.
 
@@ -26,7 +28,7 @@ Note, that if your entity is not Doctrine-related, you need to flag the identifi
 
 ## Custom Collection Data Provider
 
-First, your `BlogPostCollectionDataProvider` has to implement the [`CollectionDataProviderInterface`](https://github.com/api-platform/core/blob/master/src/DataProvider/CollectionDataProviderInterface.php):
+First, your `BlogPostCollectionDataProvider` has to implement the [`CollectionDataProviderInterface`](https://github.com/api-platform/core/blob/master/src/DataProvider/CollectionDataProviderInterface.php).
 
 The `getCollection` method must return an `array`, a `Traversable` or a [`ApiPlatform\Core\DataProvider\PaginatorInterface`](https://github.com/api-platform/core/blob/master/src/DataProvider/PaginatorInterface.php) instance.
 If no data is available, you should return an empty array.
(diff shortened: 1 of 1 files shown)


gorghoa:
I think this is a @soyuka based feature, if you could review me, I’ll promise next time you’ll come to Amiens, weather will be sunny :sun_with_face:

core/data-providers.md
```diff
@@ -161,3 +163,60 @@ final class BlogPostItemDataProvider implements ItemDataProviderInterface, Seria
...
+Assuming our `BlogPost` entity has a property `comments` holding a one to many relation to a `Comment` entity, configured as a [subresource](./operations/#subresources).
+
+First, your `BlogPostCommentsSubresourceDataProvider` has to implement the [`SubresourceDataProvider`](https://github.com/api-platform/core/blob/master/src/DataProvider/SubresourceDataProviderInterface.php):
+
+The `getSubresource` method may return virtually anything, but some behaviors are worth noting:
+
+  * If an `array`, a `Traversable` or a [`ApiPlatform\Core\DataProvider\PaginatorInterface`](https://github.com/api-platform/core/blob/master/src/DataProvider/PaginatorInterface.php) is returned, then it will be serialized as a collection like any `CollectionDataProvider`.
+  * If a managed entity is returned, then it will be serialized as an item like any `ItemDataProvider`.
+  * Scalar will be serialized without any further transformation.
+  * Objects will be considered as a collection (don’t rely on this behavior use an `array`, a `Traversable` or a [`ApiPlatform\Core\DataProvider\PaginatorInterface`](https://github.com/api-platform/core/blob/master/src/DataProvider/PaginatorInterface.php))
```
soyuka: Mhh, it's true but it's not really how we should say that. In fact the entity is serialized:
- as a collection if the relation you have the subresource on is a `toMany` 
- as an item if the relation you have the subresource on is a `toOne`

I'm not sure it makes sense.

core/data-providers.md
```diff
@@ -161,3 +163,60 @@ final class BlogPostItemDataProvider implements ItemDataProviderInterface, Seria
...
+    # ...
+    'App\DataProvider\BlogPostCommentsSubresourceDataProvider':
+        tags: [ { name: 'api_platform.subresource_data_provider', priority: 2 } ]
+        # Autoconfiguration must be disabled to set a custom priority
+        autoconfigure: false
+```
+
+Tagging the service with the tag `api_platform.subresource_data_provider` will enable API Platform Core to automatically
+register and use this data provider. The optional attribute `priority` allows to define the order in which the
+data providers are called. The first data provider not throwing a `ApiPlatform\Core\Exception\ResourceClassNotSupportedException`
```
soyuka: not supported (see RestrictedDataProvider interface, throwing should be deprecated),

gorghoa: To be clear, I should remove this part?

>  The optional attribute `priority` allows to define the order in which the
+data providers are called. The first data provider not throwing a `ApiPlatform\Core\Exception\ResourceClassNotSupportedException`

core/data-providers.md
gorghoa: Based on the `collection` key of the subresource config?

core/data-providers.md
soyuka: Whoops I'm missing a part of the sentence lol.

Instead of:

> The first data provider not throwing a `ApiPlatform\Core\Exception\ResourceClassNotSupportedException` will be used.

Say:

> The first supported data provider according to the priority will be used. 

Indeed if we use the `RestrictedDataProviderInterface` we should not need to throw.

core/data-providers.md
```diff
@@ -16,8 +16,10 @@ For a given resource, you can implement two kind of interfaces:
   is used when fetching a collection.
 * the [`ItemDataProviderInterface`](https://github.com/api-platform/core/blob/master/src/DataProvider/ItemDataProviderInterface.php)
   is used when fetching items.
+* the [`SubresourceDataProviderInterface`](https://github.com/api-platform/core/blob/master/src/DataProvider/SubresourceDataProviderInterface.php)
+  is used when fetching subresource items.
```
dunglas: Can be on the same line

dunglas:
@gorghoa great improvement, do you think you can finish this one?

floxblah:
Hello i found about this SubresourceDataProviderInterface randomly and this kind of make me succeed to implement a custom path to get all the subressource of an entity [(see my post on stack overflow for more details)](https://stackoverflow.com/questions/68830342/custom-path-to-get-all-the-subressource-of-an-entity)

Is this interface "usable" ? is it or will it be deprecated in future release of api platform ?

soyuka:
> Is this interface "usable" and supported by api platform ? is it or will it be deprecated in a future release?

It'll be deprecated in 2.7 and removed in 3.0.

PR #447, opened 2018-03-22
Enhance yaml-xml serialization groups configuration doc

GregoireHebert:
fix api-platform/api-platform#184

### DIFF
--- a/core/getting-started.md
+++ b/core/getting-started.md
@@ -196,6 +196,26 @@ resources:
             pagination_items_per_page: 25    # optional
 ```
 
+And tell your configuration how to load it:
+
+```yaml
+# api/config/packages/api_platform.yaml
+api_platform:
+    mapping:
+        paths: ['%kernel.project_dir%/config/api_platform']
+```
+
+Note that the resources configuration only support the normalization / denormalization context definition in `yaml` and `xml`. To configure the groups of serialization for each attribute, you need to define your groups like you would usually on a standard Symfony project. See the [Symfony doc](https://symfony.com/doc/current/components/serializer.html#attributes-groups).
+
+You might also need to update your framework configuration with this:
+
+``` yaml
+# api/config/packages/framework.yaml
+framework:
+    serializer:
+        enabled: true
+```
+
 **You're done!**
 
 You now have a fully featured API exposing your entities.


GuigZ-:
Thanks for your help today @GregoireHebert !

core/getting-started.md
```diff
@@ -196,6 +196,26 @@ resources:
             pagination_items_per_page: 25    # optional
 ```
 
+And tell your configuration how to load it:
+
+```yaml
+# api/config/packages/api_platform.yaml
+api_platform:
+    mapping:
+        paths: ['%kernel.project_dir%/config/api_platform']
```
meyerbaptiste: I would also like to keep the default value:
```yaml
paths:
    - '%kernel.project_dir%/src/Entity'
    - '%kernel.project_dir%/config/api_platform'
```

GregoireHebert: Agreed, Just added it :)

core/getting-started.md
```diff
@@ -196,6 +196,28 @@ resources:
...
+        paths: 
+            - '%kernel.project_dir%/src/Entity'
+            - '%kernel.project_dir%/config/api_platform'
+```
+
+Note that the resources configuration only support the normalization / denormalization context definition in `yaml` and `xml`. To configure the groups of serialization for each attribute, you need to define your groups like you would usually on a standard Symfony project. See the [Symfony doc](https://symfony.com/doc/current/components/serializer.html#attributes-groups).
+
+You might also need to update your framework configuration with this:
+
+``` yaml
```
dunglas: This is done automatically when using Flex. I'll drop this part.

core/getting-started.md
```diff
@@ -196,6 +196,28 @@ resources:
...
 
+And tell your configuration how to load it:
+
+```yaml
+# api/config/packages/api_platform.yaml
+api_platform:
+    mapping:
+        paths: 
+            - '%kernel.project_dir%/src/Entity'
+            - '%kernel.project_dir%/config/api_platform'
```
dunglas: Can you add a comment explaining that this is only useful if you use XML or YAML?

GregoireHebert:
I rephrased the explanation above and add a some comments on the configuration example.

GregoireHebert:
ping @dunglas ;)

dunglas:
Thanks @GregoireHebert

PR #448, opened 2018-03-23
Fix documentation for custom operation configuration

jonag:
While trying the recommended custom operation configuration from the documentation, I found out that the attribute `method` seems to be mandatory.


dunglas:
Thanks @jonag.

We could default to `GET` if `controller` is set, WDYT @meyerbaptiste?

meyerbaptiste:
Yes, I thought the same thing!

PR #449, opened 2018-03-23
The IRI for a resource is generated from an itemOperation

jonag:
From my experience, I always need to define a GET operation for the item to generate the IRI, not a collection operation. [This line of code](https://github.com/api-platform/core/blob/35edfb6ae999cf77e04caf81d9c33cbda9aa3890/src/Bridge/Symfony/Routing/IriConverter.php#L89) seems to confirm that.

### DIFF
--- a/core/operations.md
+++ b/core/operations.md
@@ -486,7 +486,7 @@ automatically instantiated and injected, without having to declare it explicitly
 In the following examples, the built-in `GET` operation is registered as well as a custom operation called `special`.
 The `special` operation reference the Symfony route named `book_special`.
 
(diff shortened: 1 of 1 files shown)


dunglas:
I think this change isn't accurate:

* the first item operation is used to generate the URL of an item
* the first collection operated is used to generate the URL of a collection

Maybe can you update your PR to rewrite it like that?

jonag:
Indeed, I missed the part where the collectionOperation is used. I have updated my PR with your suggestion.
Is the new phrasing OK or do you prefer something like that

> Note: By default API Platform uses the first `GET` operations defined in `itemOperations` and `collectionOperations` to generate respectively the IRI of an item and of a collection for a resource class.

to avoid the repetition?

PR #452, opened 2018-03-28
Add an “About API versioning” cookbook page

### LINKED ISSUE #451: Add a cookbook on api versionning
Regarding the popularity of questions about api versioning (https://github.com/api-platform/core/issues/972, https://github.com/api-platform/core/issues/45,  https://github.com/api-platform/api-platform/issues/290, https://symfony-devs.slack.com/archives/C39FKU9AL/p1522244783000485)

And @dunglas’ comment on https://github.com/api-platform/core/issues/45#issuecomment-99607627

> A possible solution is to register a new Resource for the new version referencing the same entity class but with other serialization an validation groups then and play with the OperationFactory to have route similar to v1 (currently the name of a resource must be unique).

> Not sure if this is a good idea to introduce a feature like versioning in the bundle. It will very difficult to have a generic enough solution. Maybe a cookbook entry with some insights on how to handle such cases will be better.

I think the subject of api versioning should be present in the doc, even if this is just a “not implemented” page.

Whether the cookbook should give some hint of how api versioning could be achieved with some hacks is debatable. IMO, the doc should provide a fully viable solution or just tell “not implemented, the reccommandation is to create a new project”.

best regards,
Your faithful

gorghoa:
Add a very basic page in core conveying this message:

> API versioning is not managed: recommended way is to create a new projet.

I’m a little concerned that this page is quite blunt, I would rather soften the message somehow.

Maybe something like “API versioning is not managed **for now**, the **current** recommended way is to create a new project”. But then I  would be concerned that it could be interpreted as some broken promise :woman_shrugging: 

closes #451

### DIFF
--- a/core/api-versioning.md
+++ b/core/api-versioning.md
@@ -1,0 +1,7 @@
+# About API versioning
+
+API Platform does not provide any specific tooling for API versioning.
+
+The recommended way to do API versioning is to create a separate project.
+
+The subject has notably been discussed [here](https://github.com/api-platform/core/issues/45) and [here](https://github.com/api-platform/core/issues/972).
--- a/index.md
+++ b/index.md
@@ -123,6 +123,7 @@
     2. [Available Serializers](core/serialization.md#available-serializers)
     3. [Decorating a Serializer and Add Extra Data](core/serialization.md#decorating-a-serializer-and-add-extra-data)
 28. [Handling Data Transfer Objects (DTOs)](core/dto.md)
+29. [About API versioning](core/api-versioning.md)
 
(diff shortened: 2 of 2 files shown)


dunglas:
I’ll add something to support API evolution in 2.3 or 2.4. Can we hold this PR until it is ready?

vinceAmstoutz:

gorghoa:
:rofl:  Thank you for taking time to add a nice note when closing this PR!

See y’all one day :)

vinceAmstoutz:
> :rofl:  Thank you for taking time to add a nice note when closing this PR!
> 
> See y’all one day :)

See you soon!  😀

PR #467, opened 2018-04-11
Update index.md - remove duplicated entries

jordanaubert:
Remove duplicated entries which lead to duplicated keys for React elements.


meyerbaptiste:
Good catch, thx @jordanaubert!

PR #469, opened 2018-04-12
Update getting-started.md - update phpdoc

xElysioN:
Hi,

The phpdoc of the variable $name was [@]param instead of [@]var. (Other variables don't have it, so i decide to remove it)
http://docs.phpdoc.org/references/phpdoc/tags/param.html

And add the phpdoc to removeOffer.

### DIFF
--- a/core/getting-started.md
+++ b/core/getting-started.md
@@ -54,7 +54,7 @@ class Product // The class name will be used to name exposed resources
     public $id;
 
     /**
-     * @param string $name A name property - this description will be available in the API documentation too.
+     * @var string $name A name property - this description will be available in the API documentation too.
      *
      * @ORM\Column
      * @Assert\NotBlank


teohhanhui:
No. I think it's fine the way it is. :smile:

dunglas:
> The phpdoc of the variable $name was [@]param instead of [@]var. (Other variables don't have it, so i decide to remove it)

Good catch, but please keep it (just change `@param` in `@var` because this annotation is used to generate the public documentation.

> And add the phpdoc to removeOffer.

Please remove it (according to our CS). The information is already provided by the typehint and adds nothing.

teohhanhui:
Oops. Yeah, I overlooked the typo lol...

xElysioN:
I remove the annotation in the removeOffer function and change param by var.

meyerbaptiste:
Thanks @xElysioN!

PR #471, opened 2018-04-23
[GraphQL] Security

alanpoulain:


core/graphql.md
```diff
@@ -68,3 +72,42 @@ class Offer
...
 ```
+
+## Security (access_control)
+
+To add a security layer to your queries and mutations, follow the [security](security.md) documentation.
+
+If your security needs differ between REST and GraphQL, add the particular parts in the `graphql` key.
+
+In the example below, we want the same security rules as in REST, but we also want to allow an admin to delete a book in GraphQL only.
+Please note it's not possible to update a book in GraphQL because the operation is not defined.
```
soyuka: Please note that

core/graphql.md
```diff
@@ -59,7 +60,10 @@ use ApiPlatform\Core\Annotation\ApiResource;
  *     graphql={
  *         "query"={
  *              "filters"={"offer.date_filter"}
- *          }
+ *          },
+ *          "delete"={},
```
dunglas: The syntax `"delete", "update", "create"` should work IIRC.

core/graphql.md
```diff
@@ -68,3 +72,42 @@ class Offer
     // ...
 }
 ```
+
+## Security (access_control)
```
dunglas: \`access_control\`

alanpoulain:
Fixed.

soyuka:
thanks @alanpoulain

PR #475, opened 2018-04-25
add default documentation formats

jdeniau:
See https://github.com/api-platform/core/pull/1891

### DIFF
--- a/core/configuration.md
+++ b/core/configuration.md
@@ -179,6 +179,11 @@ api_platform:
         jsonld:
             mime_types: ['application/ld+json']
 
+    # The list of enabled documentation formats. The first one will be the default.
+    documentation_formats:
+        jsonld:
+            mime_types: ['application/ld+json']
+
         json:
             mime_types: ['application/json']
 


core/configuration.md
```diff
@@ -179,6 +179,8 @@ api_platform:
         jsonld:
             mime_types: ['application/ld+json']
 
+    # The list of enabled documentation formats. The first one will be the default.
+    documentation_formats:
         json:
```
dunglas: jsonld should be there too right?

jdeniau: Hmmm, if you think it is useful to output jsonld, I can add it back, but as swagger understand json, I removed it

dunglas: If we disable the Hydra documentation (`jsonld` format), the admin and the PWA generator will not work anymore.

jdeniau:
OK I will add it back

Le mer. 25 avr. 2018 à 16:48, Kévin Dunglas <notifications@github.com> a
écrit :

PR #478, opened 2018-05-03
feature: add helm dependencies update

Simperfit:


teohhanhui:
https://docs.helm.sh/helm/#helm-dependency-update
> On successful update, this will generate a lock file that can be used to rebuild the requirements to an exact version.

Shouldn't we check in the lock file?

dunglas:
Thanks @Simperfit. I think we need to add the lock file, indeed.

teohhanhui:
@dunglas The `requirements.lock` file has been added in https://github.com/api-platform/api-platform/pull/666

dunglas:
Do we need to update the docs then?

teohhanhui:
It depends on whether we think it's desirable to have fixed dependencies. I'm not sure it's the right thing to do.

PR #486, opened 2018-05-28
Add expose-headers note

Steveb-p:
This PR adds a note about particular requirements that need to be met when creating Admin without Docker.

Based on discussion from an issue here: https://github.com/api-platform/admin/issues/73 and SO discussion: https://stackoverflow.com/questions/50394139/unable-to-retrieve-api-documentation-on-admin-of-api-platform.

Otherwise the issue is rather hard to pinpoint without digging into CORS and checking requests and responses. Neither HydraAdmin nor anything else complains about anything. Console is empty in this case either. Just the dreaded: “Unable to retrieve API documentation”.

### DIFF
--- a/admin/getting-started.md
+++ b/admin/getting-started.md
@@ -35,6 +35,8 @@ the admin's domain to access it.
 To do so, update the value of the `CORS_ALLOW_ORIGIN` parameter in `api/.env` (it will be set to `^https?://localhost:?[0-9]*$`
 by default).
 
+If you're not using Api Platform distribution, you will need to adjust `nelmio-cors.yaml` to include `expose_headers: ['Content-Disposition', 'Content-Length', 'Link']` on the route under which api will be served (`/api` by default).
+
 Clear the cache to apply this change:
 
     $ docker-compose exec php bin/console cache:clear --env=prod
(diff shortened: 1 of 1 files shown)


admin/getting-started.md
```diff
@@ -35,6 +35,8 @@ the admin's domain to access it.
 To do so, update the value of the `CORS_ALLOW_ORIGIN` parameter in `api/.env` (it will be set to `^https?://localhost:?[0-9]*$`
 by default).
 
+If you're not using Api Platform distribution, you will need to adjust `nelmio-cors.yaml` to include `expose_headers: ['Content-Disposition', 'Content-Length', 'Link']` on the route under which api will be served (`/api` by default).
```
dunglas: According to https://github.com/api-platform/api-platform/blob/master/api/config/packages/nelmio_cors.yaml, only `Link` is mandatory. Are you sure the other ones are required?

Steveb-p: Sorry, I've followed the suggestion on SO and one of the comments in the linked issue. Now that I'm looking at it since what you've linked is probably from the docker build, it should be the only one needed ;)

dunglas:
Thanks @Steveb-p. I'm sure it will help a lot of people.
I made small modifs.

PR #487, opened 2018-05-30
update wrong property name

antograssiot:
https://api-platform.com/docs/core/file-upload#linking-a-mediaobject-resource-to-another-resource
the property name is `image` not `cover`


dunglas:
Thanks @antograssiot!

PR #489, opened 2018-05-31
Feat: add constants for PRE and POST serialize

toofff:


soyuka:
May you target 2.2 instead? Thanks!

dunglas:
Thanks Yann!

PR #493, opened 2018-06-03
Update docmentation about Behat

tuvokki:
Since it is no longer part of the standard distribution the Behat BDD library needs to be installed before you can proceed with the documented usage.

### DIFF
--- a/distribution/testing.md
+++ b/distribution/testing.md
@@ -1,6 +1,6 @@
 # Testing and Specifying the API
 
-A set of useful tools to specify and test your API are pre-configured in the API Platform distribution:
+A set of useful tools to specify and test your API are easily installable in the API Platform distribution:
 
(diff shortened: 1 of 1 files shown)


dunglas:
Good catch, thanks!

distribution/testing.md
```diff
@@ -12,6 +12,14 @@ A set of useful tools to specify and test your API are pre-configured in the API
 Take a look at [the Symfony documentation about testing](https://symfony.com/doc/current/testing.html) to learn how to use
 PHPUnit in your API Platform project.
 
+Installing Behat is easy enough following these steps:
+
+    $ docker-compose exec php composer require --dev behat/behat
```
teohhanhui: It's unnecessary to prefix `php` to any of the commands... :smile:

PR #496, opened 2018-06-06
[GraphQL] Add documentation about serialization groups

alanpoulain:
Because it can be really useful!

### DIFF
--- a/core/graphql.md
+++ b/core/graphql.md
@@ -174,3 +174,65 @@ class Book
     // ...
 }
 ```
+
+## Serialization Groups
+
+You may want to restrict some resource's attributes to your GraphQL clients.
+
+As described in the [serialization process](serialization.md) documentation, you can use serialization groups to expose only the attributes you want in queries or in mutations.
+
+If the (de)normalization context between GraphQL and REST is different, use the `graphql` key to change it.
+
+Note that:
+
+* A **query** is only using the normalization context.
+* A **mutation** is using the denormalization context for its input and the normalization context for its output.
+
+The following example shows you what can be done:
+
+```php
+<?php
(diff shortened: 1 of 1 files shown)


core/graphql.md
```diff
@@ -174,3 +174,62 @@ class Book
     // ...
 }
 ```
+
+## Serialization Groups
+
+You may need to prevent clients using your GraphQL endpoint to access some attributes of your resources.
```
soyuka: Not sure to understand this sentence. What about:

You may want to restrict some resource's attributes to your clients with GraphQL.

core/graphql.md
```diff
@@ -174,3 +174,62 @@ class Book
...
+
+Like described in the [serialization process](serialization.md) documentation, you can use serialization groups to expose only the attributes you want in queries or in mutations.
```
soyuka: Following the previous sentence:

As described in the serialization process documentation, ...

core/graphql.md
```diff
@@ -174,3 +174,62 @@ class Book
...
+
+If the (de)normalization context between GraphQL and REST is different, use the `graphql` key to change it.
+
+A query is only using a normalization context whereas a mutation is using a denormalization context for the input and a normalization context for its output.
```
soyuka: Note that:

- A **query** is using the normalization context
- A **mutation** is using the denormalization context for it's input and the normalization context for it's output

core/graphql.md
alanpoulain: What do you mean by: "Following the previous sentence:"?

soyuka: You may need ... 

As described in the ....

Simperfit:
Thanks Alan!

PR #500, opened 2018-06-06
Remove inconsistent reference to database engine

### LINKED ISSUE #499: Reference to wrong database engine (MySQL rather than Postgres)
Just a tiny issue:

On https://api-platform.com/docs/distribution, it says

> Doctrine's annotations map these entities to tables in the MySQL database.

However, if you follow the guide, you'll find that you've installed Postgres rather than MySQL. So the docs are slightly inconsistent.

I'll suggest a change.

yhoiseth:
Fixes #499.


dunglas:
Good catch, thanks!
Indeed, MySQL has been replaced by Postgres in the last version.

PR #502, opened 2018-06-07
Fix link, make clickable

### LINKED ISSUE #501: Wrong protocol and port to development server when using advanced installation method
After instructing to use `bin/console server:run`, the documentation says:

> Note: when installing API Platform this way, the API will be exposed as the `/api/` path. You need to open `https://localhost/api/` to see the API documentation.

Visiting https://localhost/api/ produces `ERR_CONNECTION_REFUSED` on the client side.

Visiting https://localhost:8000/api/ produces `ERR_CONNECTION_REFUSED` on the client side and `Invalid request (Unsupported SSL request)` on the server side.

http://localhost:8000/api/ works.

I'll create a pull request.

yhoiseth:
Fixes #501.

### DIFF
--- a/distribution/index.md
+++ b/distribution/index.md
@@ -151,7 +151,7 @@ And start the built-in PHP server or the Symfony WebServerBundle:
 All JavaScript components are also [available as standalone libraries](https://github.com/api-platform?language=javascript)
 installable with NPM or Yarn.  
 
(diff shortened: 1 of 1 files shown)


dunglas:
Thanks, good catch! I removed the link and restored the backticks for consistency and because it's bad for SEO to create link that point to nowhere for most people.

yhoiseth:
Cool. I didn't know about the SEO issue :)

PR #503, opened 2018-06-09
Sync admin's docs with the last version (React Admin)

dunglas:


init-penguin:
It seems that you've missed `authClient={authClient}` -> `authProvider={authProvider}`

dunglas:
Good catch @init-penguin, fixed

init-penguin:
Sorry to point this again, @dunglas, but i think you forgot about the import statement :)
Now it imports the `authClient` while exporting `HydraAdmin` with `authProvider`, thus you should either change the import to `import authProvider` thing (and filename for consistency too), or change the export of App.js to `authProvider={authClient}`

dunglas:
I'm not sure we need to update the the filename. The React Admin team changed the name of the property to highlight it can be other things than an API client (I guess), but they haven't updated the name of their own clients (for instance the REST client is still called "restClient"). It's why I've not renamed the file and updated the import. But I'm not sure about this.

@fzaninotto, do you have a hint regarding the naming best practice?

fzaninotto:
We try to use "provider" everywhere to avoid confusion. If you found some mentions of `restClient`, that's a mistake on our side. I'll look into the remaining mentions of pre-2.0 terminology.

PR #505, opened 2018-06-14
Fixes #424 - Wrong Kernel Event / EventPriorities in Errors Example

### LINKED ISSUE #424: Error handling example is wrong
It looks like the [example in errors handling](https://api-platform.com/docs/core/errors/#converting-php-exceptions-to-http-errors) section is taking the wrong method argument.

```php
<?php

namespace AppBundle\EventSubscriber;

use ApiPlatform\Core\EventListener\EventPriorities;
use AppBundle\Entity\Product;
use AppBundle\Exception\ProductNotFoundException;
use Symfony\Component\EventDispatcher\EventSubscriberInterface;
use Symfony\Component\HttpFoundation\Request;
use Symfony\Component\HttpKernel\Event\GetResponseForControllerResultEvent;
use Symfony\Component\HttpKernel\KernelEvents;

final class ProductManager implements EventSubscriberInterface
{
    public static function getSubscribedEvents(): array
    {
        return [
            KernelEvents::REQUEST => ['checkProductAvailability', EventPriorities::POST_DESERIALIZE],
        ];
    }

    public function checkProductAvailability(GetResponseForControllerResultEvent $event): void
    {
        $product = $event->getControllerResult();
        if (!$product instanceof Product || !$event->getRequest()->isMethodSafe(false)) {
            return;
        }

        if (!$product->isPubliclyAvailable()) {
            // Using internal codes for a better understanding of what's going on
            throw new ProductNotFoundException(sprintf('The product "%s" does not exist.', $product->getId()));
        }
    }
}
```
In the example the function uses `GetResponseForControllerResultEvent` type hint, but this is only available during [`kernel.view`](https://symfony.com/doc/current/reference/events.html#kernel-view)

So, running the example code will throw an error
```
Type error: Argument 1 passed to ... must be an instance of Symfony\Component\HttpKernel\Event\GetResponseForControllerResultEvent, instance of Symfony\Component\HttpKernel\Event\GetResponseEvent given
```

cvergne:
Hi,

Not sure if `master` is the good branch to make the fix/PR, so don't hesitate to tell me if I should do it on 2.2.

Btw, a contributing guide for the doc would be good, at least with this kind of information.


dunglas:
Fixes should be in 2.2, but can you explain a bit more what it fixes? Thanks!

cvergne:
Hi, I closed this one and opened a new one with correct branch and explanations : #508 :)

PR #506, opened 2018-06-14
--output parameter on api:swagger:export console

er1z:
Additionally, removed the broken diff result.


dunglas:
Good catch, thanks for improving the docs @er1z

PR #509, opened 2018-06-20
Fixed the event constant table

dovab:
The event constant table listed the `PRE_SERIALIZE` and `POST_SERIALIZE` as constants for the `kernel.request` event. The correct event for those constants should be `kernel.view`.


teohhanhui:
Thanks @dovab. Nice catch! :)

dovab:
You're welcome! :)

PR #511, opened 2018-06-20
Add exemple for feature api-platform/schema-generator#132

ErikSaunier:
Add exemple for feature api-platform/schema-generator#132


dunglas:
Thanks @ErikSaunier

PR #516, opened 2018-06-27
Feat: Add validation groups on operations

toofff:


core/validation.md
```diff
@@ -167,6 +167,70 @@ Of course, you can use XML or YAML configuration format instead of annotations i
 You may also pass in a [group sequence](http://symfony.com/doc/current/validation/sequence_provider.html) in place of
 the array of group names.
 
+## Using Validation Groups on operations
+
+There are [5 types of operations](operations.md) in API Platform and you can have different validation for each.
```
alanpoulain: validations?

core/validation.md
```diff
@@ -167,6 +167,70 @@ Of course, you can use XML or YAML configuration format instead of annotations i
...
+<?php
+// api/src/Entity/Book.php
+
+use ApiPlatform\Core\Annotation\ApiResource;
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * @ApiResource(
+ *     collectionOperations={
+ *         "get"={"method"="GET"},
```
alanpoulain: Maybe you can remove the explicit method declaration to be clearer?

core/validation.md
```diff
@@ -167,6 +167,70 @@ Of course, you can use XML or YAML configuration format instead of annotations i
...
+}
+```
+
+With this configuration, there are three validation groups:
+
+`Default` contains the constraints in the current class and all referenced classes that belong to no other group.
+
+`postValidation` contains the constraints on the name and author (length from 2 to 50) fields only.
+
+`putValidation` contains the constraints on the author (length from 2 to 70) fields only.
```
alanpoulain: field

core/validation.md
```diff
@@ -167,6 +167,70 @@ Of course, you can use XML or YAML configuration format instead of annotations i
...
+     */
+    private $author;
+
+    // ...
+}
+```
+
+With this configuration, there are three validation groups:
+
+`Default` contains the constraints in the current class and all referenced classes that belong to no other group.
```
alanpoulain: I don't understand the `all referenced classes` part. You need a `Valid` assert for this kind of behavior, don't you?

core/validation.md
dunglas: I would reword `There are [5 types of operations](operations.md)`. It's not exact: there are 2 builtin collection operations, and 4 item operations (PATCH is JSONAPI specific), but API Platform supports any type and any number of operations.

core/validation.md
toofff: no, what I saw with Greg, if not stated, it will not work.

toofff: ping @dunglas

toofff: When declaring the POST method, we have to declare the GET method to access it.

That's why I wanted to put it, because we often need the GET on the collection

alanpoulain: No, I'm saying you can do like this:
```
 *     collectionOperations={
 *         "get",
 *         "post"={"validation_groups"={"Default", "postValidation"}}
 *     },
```

core/validation.md
```diff
@@ -167,6 +167,70 @@ Of course, you can use XML or YAML configuration format instead of annotations i
 You may also pass in a [group sequence](http://symfony.com/doc/current/validation/sequence_provider.html) in place of
 the array of group names.
 
+## Using Validation Groups on operations
+
+There are [2 builtin collection operations, and 4 item operations](operations.md) (PATCH is JSONAPI specific), but API Platform supports any type and any number of operations.
```
alanpoulain: I would reword it like this:
```
You can have different validation for each [operation](operations.md) related to your resource.
```
Like this you don't need to duplicate the explanation about the concept of operations in API Platform.

core/validation.md
```diff
@@ -167,6 +167,70 @@ Of course, you can use XML or YAML configuration format instead of annotations i
...
+// api/src/Entity/Book.php
+
+use ApiPlatform\Core\Annotation\ApiResource;
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * @ApiResource(
+ *     collectionOperations={
+ *         "get",
+ *         "post"={"method"="POST", "validation_groups"={"Default", "postValidation"}}
```
alanpoulain: I think it's working of you remove the `method` part but I'm not sure about this.

toofff:
@dunglas @alanpoulain Done

dunglas:
Thanks @toofff

PR #519, opened 2018-07-02
[WIP] Fix some admin code examples

### LINKED ISSUE #483: Remove “Handling Relations to Collections”?
On the page https://api-platform.com/docs/admin/handling-relations-to-collections, it is written:

> Currently, API Platform Admin doesn't handle `to-many` relations. The core library [is being patched](https://github.com/api-platform/core/pull/1189) to document relations to collections through OWL.
> 
> Meanwhile, it is possible to manually configure API Platform to handle relations to collections.

But https://github.com/api-platform/core/pull/1189 was merged on 29 Jun 2017.

So it looks like this page is documenting a configuration that is not necessary anymore? Should the page be rewritten or removed entirely?

webda2l:
Fix #483


admin/handling-relations-to-collections.md
```diff
@@ -1,174 +0,0 @@
-# Handling Relations to Collections
```
dunglas: Can you keep the old name please? Renaming will break existing links.

webda2l: Sure, done.

webda2l:
Rest some 

```
***.input.defaultProps = {
   addField: true,
   addLabel: true,
};
```

required to add maybe? I don't totally get it about this..

And the code about **Managing Files and Images** that seems the old way

dunglas:
Thanks @webda2l

PR #524, opened 2018-07-06
Update doc for https://github.com/api-platform/core/pull/2070

ArnoudThibaut:
Add come documentation for https://github.com/api-platform/core/pull/2070


core/graphql.md
```diff
@@ -135,6 +135,24 @@ Or order your results like:
   }
 }
 ```
+An other difference with the REST Api filters is that the keyword `_list` must be used instead of the traditional `[]` to filter over multiple values.
```
alanpoulain: Another
API

dunglas:
Thanks @ArnoudThibaut

PR #525, opened 2018-07-06
Fix data persister typo

kbsali:


dunglas:
Good catch, thanks @kbsali

PR #528, opened 2018-07-08
Fix routePrefix documentation

rogamoore:
It needs to be either
```
@ApiResource(attributes={"route_prefix"="/library"})
```
or 
```
@ApiResource(routePrefix="/library")
```
to work correctly.


dunglas:
Thanks @rogamoore, good catch!

PR #530, opened 2018-07-09
remove index.md as it is now generated

Mathieudewet:


teohhanhui:
It's generated from https://github.com/api-platform/website/blob/master/config/outline.yaml so we're just moving it from one place to another (kind of).

dunglas:
Not really. This file contains only the order (and it must be moved in this repository IMO).
Currently, the index contains the exact titles and corresponding anchors. It's a pain to update and maintain. 👍 to delete (or generate) it.

dunglas:
Thanks @Mathieudewet

PR #535, opened 2018-07-12
Fix Behat tutorial

dunglas:


distribution/testing.md
```diff
@@ -12,14 +12,38 @@ A set of useful tools to specify and test your API are easily installable in the
...
 
+Register the Behatch extension:
+
+```yaml
+# api/behat.yml.dist
+default:
+  suites:
+    default:
+      contexts:
+        - FeatureContext: { doctrine: "@kernel" }
```
rogamoore: This should be @doctrine I guess? :)

rogamoore:
Currently this config only works with Symfony 4.x when using 
```
"behat/mink": "dev-master#d5ee350c40baff5f331a05ebdbe1927345c9ac8b as 1.7.1"
````
in composer.json.

As there is no doctrine config for the test environment it would run against the default database and only work once.

So there should be a config like this in config/packages/test/doctrine.yaml:
```
doctrine:
    dbal:
        url: 'sqlite:///%kernel.cache_dir%/test.db'
```

However we should also clear the database before each test scenario, so the FeatureContext.php should be changed to something like this (inspired by the api-platform core tests):
```
<?php
use Behat\Behat\Context\Context;
use Doctrine\Common\Persistence\ManagerRegistry;
use Doctrine\ORM\Tools\SchemaTool;

/**
 * This context class contains the definitions of the steps used by the demo
 * feature file. Learn how to get started with Behat and BDD on Behat's website.
 *
 * @see http://behat.org/en/latest/quick_start.html
 */
class FeatureContext implements Context
{
    /**
     * @var ManagerRegistry
     */
    private $managerRegistry;

    /**
     * @param ManagerRegistry $managerRegistry
     */
    public function __construct(ManagerRegistry $managerRegistry)
    {
        $this->managerRegistry = $managerRegistry;
    }

    /**
     * @BeforeScenario @createSchema
     */
    public function createDatabase()
    {
        $manager = $this->managerRegistry->getManager();
        $metadata = $manager->getMetadataFactory()->getAllMetadata();

        $schemaTool = new SchemaTool($manager);
        $schemaTool->dropSchema($metadata);
        $manager->clear();
        $schemaTool->createSchema($metadata);
    }
}

```

PR #539, opened 2018-07-16
Clarity, grammer, and style changes to serialization docs

caseyamcl:
I made some changes to the serialization documentation for clarity, grammar, and style.  There were also a few typos that I corrected.

If you think this is a worthwhile PR, I would be happy to provide suggestions for some of the other pages in the documentation as well.

### DIFF
--- a/core/serialization.md
+++ b/core/serialization.md
@@ -8,12 +8,12 @@ The main serialization process has two stages:
 
 ![Serializer workflow](images/SerializerWorkflow.png)
 
-> As you can see in the picture above, an array is used as a man in the middle. This way, Encoders will only deal with turning specific formats into arrays and vice versa. The same way, Normalizers will deal with turning specific objects into arrays and vice versa.
+> As you can see in the picture above, an array is used as a man-in-the-middle. This way, Encoders will only deal with turning specific formats into arrays and vice versa. The same way, Normalizers will deal with turning specific objects into arrays and vice versa.
 -- [The Symfony documentation](https://symfony.com/doc/current/components/serializer.html)
 
 Unlike Symfony itself, API Platform leverages custom normalizers, its router and the [data provider](data-providers.md) system to do an advanced transformation. Metadata are added to the generated document including links, type information, pagination data or available filters.
 
-The API Platform Serializer is extendable, you can register custom normalizers and encoders to support other formats. You can also decorate existing normalizers to customize their behaviors.
+The API Platform Serializer is extendable. You can register custom normalizers and encoders in order to support other formats. You can also decorate existing normalizers to customize their behaviors.
 
 ## Available Serializers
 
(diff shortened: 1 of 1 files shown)


core/serialization.md
```diff
@@ -105,21 +106,23 @@ Alternatively, you can use the verbose syntax that is strictly equivalent:
...
-key when using [the `@MaxDepth` annotation](https://symfony.com/doc/current/components/serializer.html#handling-serialization-depth)).
+In addition to the `groups` key, you can configure any Symfony Serializer option through the `$context` parameter
+(e.g. the `enable_max_depth`key when using [the `@MaxDepth` annotation](https://symfony.com/doc/current/components/serializer.html#handling-serialization-depth)).
 
-Built-in actions and the Hydra documentation generator will leverage the specified serialization and deserialization groups
-to give access only to exposed properties and to guess if they are readable and/or writable.
+Any serialization and deserialization groups that you specify will also be leveraged by the built-in actions and the Hydra
+documentation generator.
 
-## Using Different Serialization Groups per Operation
```
dunglas: This case is intended (title case)

dunglas:
Thank you, this kind of changes improve the quality of the documentation and are much appreciated!

I reverted some changes for consistency with other documentation entries:

* we use 1 space after a full period, because of this https://lifehacker.com/no-you-still-shouldnt-put-two-spaces-after-a-period-1825662114 (and because in French we never use two spaces after a period :D)
* we use title case

It would be awesome if you can proof-read other documents!

caseyamcl:
@dunglas thanks!  I appreciate being able to help with the project, and will continue to do so as time allows.

core/serialization.md
```diff
@@ -8,12 +8,12 @@ The main serialization process has two stages:
 
 ![Serializer workflow](images/SerializerWorkflow.png)
 
-> As you can see in the picture above, an array is used as a man in the middle. This way, Encoders will only deal with turning specific formats into arrays and vice versa. The same way, Normalizers will deal with turning specific objects into arrays and vice versa.
+> As you can see in the picture above, an array is used as a man-in-the-middle. This way, Encoders will only deal with turning specific formats into arrays and vice versa. The same way, Normalizers will deal with turning specific objects into arrays and vice versa.
```
teohhanhui: MITM might be understood as an attack lol... Maybe we could say "an intermediary".

core/serialization.md
```diff
@@ -8,12 +8,12 @@ The main serialization process has two stages:
...
 -- [The Symfony documentation](https://symfony.com/doc/current/components/serializer.html)
 
 Unlike Symfony itself, API Platform leverages custom normalizers, its router and the [data provider](data-providers.md) system to do an advanced transformation. Metadata are added to the generated document including links, type information, pagination data or available filters.
 
-The API Platform Serializer is extendable, you can register custom normalizers and encoders to support other formats. You can also decorate existing normalizers to customize their behaviors.
+The API Platform Serializer is extendable. You can register custom normalizers and encoders in order to support other formats. You can also decorate existing normalizers to customize their behaviors.
```
teohhanhui: In the context of PHP, "extendable" might give the wrong idea of the class being able to be `extend`ed, so perhaps we could use the synonym "extensible".

core/serialization.md
caseyamcl: @teohhanhui I agree, although I didn't change that, because I thought it was a direct quote from the Symfony Docs.  Looking at the [docs](https://symfony.com/doc/current/components/serializer.html), now I see they have updated the language over there.

core/serialization.md
caseyamcl: @teohhanhui Agreed.

PR #541, opened 2018-07-18
Add missing Field Cardinality in documentation

lologhi:
Based on [https://github.com/api-platform/schema-generator/issues/112#issuecomment-357796763](this comment) and `CardinalitiesExtractor.php`.


dunglas:
Thanks, can you also describe them as in the previous lines?

dunglas:
I guess it will be better than not documented at all :). Thanks!

Issue #458, opened 2018-04-06
Unable to change the Swagger UI Location


comxd:
@see https://github.com/api-platform/api-platform/issues/635

About the api-platform documentation at: _Changing the Swagger UI Location_

I got this error for "/" and "/docs" routes:
Serialization for the format html is not supported.

```
#api/config/routes.yaml
swagger_ui:
    path: /docs
    controller: api_platform.swagger.action.ui
```

```
# api/config/packages/api_platform.yaml
api_platform:
    # ...
    enable_swagger_ui: false
```

It's due to the route `vendor/api-platform/core/src/Bridge/Symfony/Bundle/Resources/config/routing/docs.xml : /docs.{_format}` used for the hydra doc.

For change the Swagger UI location, we need to : 
* Override the `/docs.{_format}` route
* or Choose another location of "/docs"

dimitriacseo:
Can we have a complete configuration file ?

alanpoulain:
See https://github.com/api-platform/docs/pull/1604.

PR #547, opened 2018-07-22
Fix config for custom swagger route

rogamoore:
Fixes https://github.com/api-platform/docs/issues/458

The key needs to be `api_doc` to override the existing config for `/docs` in https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/Resources/config/routing/docs.xml.

At least to make the given example work. :)
Changing `/docs` to something else doesn't conflict with the existing config.

However, the reason why I tried to use the example was to disable the docs on `/` and only enable them on `/docs`

### DIFF
--- a/core/swagger.md
+++ b/core/swagger.md
@@ -268,7 +268,7 @@ api_platform:
 
 ```yaml
 # app/config/routes.yaml
-swagger_ui:
+api_doc:
     path: /docs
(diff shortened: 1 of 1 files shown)


alanpoulain:
Superseded by https://github.com/api-platform/docs/pull/1604.

PR #548, opened 2018-07-23
customize resource creation & edition rendering

jfthuillier:


jfthuillier:
No longer necessary, will be entirely reworked.

PR #549, opened 2018-07-24
fix images

Mathieudewet:
Related to: https://github.com/api-platform/website/pull/69 and fix https://github.com/api-platform/website/issues/50


dunglas:
It doesn’t look good to me: the images must still be displayed correctly on GitHub and when the documentation is downloaded locally.

teohhanhui:
I agree with @dunglas, there should be no changes necessary here. Instead, api-platform/website needs to transform those image paths.

teohhanhui:
Sorry, I misunderstood. I think it should work once we merge. There seems to be a GitHub bug when dealing with branch names with a slash.

Mathieudewet:
Requires rights to make a call to `git remote show git@github.com:api-platform/docs.git`.

PR #550, opened 2018-07-24
Simplify file upload handling

teohhanhui:


core/file-upload.md
```diff
@@ -74,14 +74,36 @@ class MediaObject
      * @Assert\NotNull()
      * @Vich\UploadableField(mapping="media_object", fileNameProperty="contentUrl")
      */
-    public $file;
+    protected $file;
```
dunglas: Why this change? Why try to use public properties as much as possible in docs.

teohhanhui: Why? It's a bad practice.

dunglas: It makes the examples simpler, it’s less verbose, and it’s not a bad practice in this case where it brings nothing: https://dev.to/scottshipp/avoid-getters-and-setters-whenever-possible-c8m

Also, getters and setters are also a bad practice... https://www.javaworld.com/javaworld/jw-09-2003/jw-0905-toolbox.html

The only benefit of getters/setters in PHP is type safety (until the typed properties RFC is accepted). And it’s not an issue with API Platform because we check the types during the deserialization.

core/file-upload.md
```diff
@@ -97,94 +119,62 @@ that handles the file upload.
...
-        // This will be handled by API Platform and returns a validation error.
-        throw new ValidationException($this->validator->validate($mediaObject));
-    }
-}
-```
+        $mediaObject = new MediaObject();
+        $mediaObject->setFile($uploadedFile);
 
-As you can see, the action uses a form. You will need this form to be like this:
+        $this->validate($mediaObject);
```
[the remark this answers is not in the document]
jewome62: Or integrate into validate function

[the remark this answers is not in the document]
teohhanhui: This is using the ApiPlatform `ValidatorInterface`, so the correct `ValidationException` is already being thrown.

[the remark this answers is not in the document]
teohhanhui: Though, to be honest, it's all kinds of weird that the `ValidatorInterface` which is not in a "Bridge" throws an exception which is in a "Bridge".

core/file-upload.md
```diff
@@ -49,39 +49,57 @@ use ApiPlatform\Core\Annotation\ApiResource;
...
- *     "post"={
- *         "method"="POST",
- *         "path"="/media_objects",
- *         "controller"=CreateMediaObjectAction::class,
- *         "defaults"={"_api_receive"=false},
+ * @ApiResource(
+ *     iri="http://schema.org/MediaObject",
+ *     collectionOperations={
+ *         "post"={
+ *             "method"="POST",
```
dunglas: can be removed IIRC

core/file-upload.md
```diff
@@ -92,99 +110,115 @@ that handles the file upload.
...
+use ApiPlatform\Core\Util\RequestAttributesExtractor;
+use ApiPlatform\Core\Validator\ValidatorInterface;
 use App\Entity\MediaObject;
-use App\Form\MediaObjectType;
-use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
-use Symfony\Bridge\Doctrine\RegistryInterface;
-use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
-use Symfony\Component\Form\FormFactoryInterface;
+use Doctrine\Common\Persistence\ManagerRegistry;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
```
dunglas: It's not installed by default IIRC, we should explain it to the user.

teohhanhui: Indeed, it should use API Platform's expression language instead. :)

core/file-upload.md
```diff
@@ -92,99 +110,115 @@ that handles the file upload.
...
-        }
+        return $mediaObject;
+    }
+
+    /**
+    * @throws ValidationException
+    */
+    private function validate(MediaObject $mediaObject): void
+    {
+        $attributes = RequestAttributesExtractor::extractAttributes($request);
```
borNfreee: `$request` is undefined

teohhanhui: Nice catch!

core/file-upload.md
```diff
@@ -92,99 +113,111 @@ that handles the file upload.
...
-        }
+        return $mediaObject;
+    }
 
-        // This will be handled by API Platform and returns a validation error.
-        throw new ValidationException($this->validator->validate($mediaObject));
+    /**
+    * @throws ValidationException
+    */
+    private function validate(MediaObject $mediaObject, Requst $request): void
```
Jibbarth:
```suggestion
    private function validate(MediaObject $mediaObject, Request $request): void
```

core/file-upload.md
```diff
@@ -92,99 +113,111 @@ that handles the file upload.
...
+        ];
     }
 
-    public function getBlockPrefix()
+    public function onPreSerialize(GetResponseForControllerResultEvent $event)
     {
-        return '';
+        $mediaObject = $event->getControllerResult();
+
+        if (!$mediaObject instanceof MediaObject) {
```
Jibbarth: This will work only for `get item operation`, right ? Or it'll work also on `get collection operation`. 
I thought in case of collection operation, `getControllerResult` returned an `AbstractPaginator`, but may I'm wrong.

core/file-upload.md
teohhanhui: Oops! 🙈

core/file-upload.md
teohhanhui: You're right. I'll fix this.

core/file-upload.md
```diff
@@ -49,39 +49,60 @@ use ApiPlatform\Core\Annotation\ApiResource;
...
  *     },
- * })
+ *     collectionOperations={
+ *         "post"={
+ *             "controller"=CreateMediaObjectAction::class,
+ *             "defaults"={
+ *                 "_api_receive"=false,
+ *             },
+ *             "access_control"="is_granted('ROLE_USER')",
+ *             "validation_groups"={"Default", "media_object_create"},
```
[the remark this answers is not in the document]
teohhanhui: @Jibbarth Added and fixed the double array with `consumes`. It's supposed to be an array of strings: https://github.com/api-platform/core/blob/c80d4934ead61ffd4d5e08f3d86c79fa6fb75ce0/tests/Swagger/Serializer/DocumentationNormalizerV2Test.php#L163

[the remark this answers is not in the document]
teohhanhui: https://swagger.io/docs/specification/2-0/file-upload/

teohhanhui:
@dunglas Done.

dunglas:
Thanks @teohhanhui

PR #552, opened 2018-08-01
Added yaml config to security page

mikemilano:
Here's a PR which adds a YAML config on the security page as  requested: https://github.com/api-platform/api-platform/issues/785


dunglas:
Thanks!

PR #554, opened 2018-08-03
Corrected typo in BookContextBuilder example: admin_input vs admin:input

mikemilano:
The serialization documentation section `Changing the Serialization Context Dynamically` defines a serialization group as `admin:input` in the entity exampe, however it was being added as `admin_input` in the `BookContextBuilder` example.

i.e. changed from:
```
$context['groups'][] = 'admin_input';
```

to:
```
$context['groups'][] = 'admin:input';
```


soyuka:
Can you set `admin_input` everywhere instead? Thanks!

dunglas:
Thanks @mikemilano
@soyuka using a column is smart because it’s a special PHP char so it prevents conflicts. I can’t remember who I’ve seen using this trick first,
But after that I updated the doc (almost) evrywhere to use colons when applicable.

PR #556, opened 2018-08-06
Added serialization config to complete the YAML format example.

mikemilano:
The serialization groups section mentions "You can also use the YAML configuration format", but only shows the API Platform portion of the YAML format.

For developers choosing to define everything in YAML, they also need the `/config/serialization/Book.yaml` file to be defined.

In other discussions I've read on the subject, it's mentioned that the serialization component is already  documented elsewhere. I think it's important to deliver the config on a silver platter so there is nothing left for interpretation which may take the developer down a rabbit hole.

### DIFF
--- a/core/serialization.md
+++ b/core/serialization.md
@@ -118,6 +118,16 @@ App\Entity\Book:
             groups: ['write']
 ```
 
+```yaml
+# api/config/serialization/Book.yaml
+App\Entity\Book:
+    attributes:
+        name:
+            groups: ['read', 'write']
+        author:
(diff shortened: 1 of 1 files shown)


dunglas:
Thanks! It will help a lot of newcomers.

Brewal:
I didn't manage to get the yaml configuration working by naming the directory `serialization` but by naming it `serializer` as said in the Symfony doc : 

https://symfony.com/doc/current/serializer.html#using-serialization-groups-annotations

PR #560, opened 2018-08-07
Add debugging section to the core manual

### LINKED ISSUE #551: Add a `how to create dev/debug stage using xdebug` section
see https://github.com/api-platform/api-platform/pull/782

hschaeidt:
resolves #551

### DIFF
--- a/core/debugging.md
+++ b/core/debugging.md
@@ -1,0 +1,66 @@
+# Debugging
+
+The default Docker stack is shipped without a Xdebug stage. It's easy
+though to add [Xdebug](https://xdebug.org/) to your project, for development
+purposes such as debugging tests or API requests remotely.
+
+## Add a Development Stage to the Dockerfile
+
+To avoid deploying API Platform to production with an active Xdebug extension,
+it's recommended to add a custom stage to the end of the `api/Dockerfile`.
+
+```Dockerfile
+# api/Dockerfile
+FROM api_platform_php as api_platform_php_dev
+
+ARG XDEBUG_VERSION=2.6.0
+RUN set -eux; \
+	apk add --no-cache --virtual .build-deps $PHPIZE_DEPS; \
+	pecl install xdebug-$XDEBUG_VERSION; \
+	docker-php-ext-enable xdebug; \
+	apk del .build-deps
+```
+
+## Configure Xdebug with Docker Compose Override
+
+Using an [override](https://docs.docker.com/compose/reference/overview/#specifying-multiple-compose-files)  
+ file named `docker-compose.override.yml` ensures that the production
+configuration remains untouched.
+
+As example, an override could look like this:
+
+```yml
+version: "3.4"
+
+services:
+  php:
+    build:
+      target: api_platform_php_dev
+    environment:
+      # See https://docs.docker.com/docker-for-mac/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-host
+      # See https://github.com/docker/for-linux/issues/264
+      # The `remote_host` below may optionally be replaced with `remote_connect_back`
+      XDEBUG_CONFIG: >-
+        remote_enable=1
+        remote_host=host.docker.internal
+        remote_port=9000
+        idekey=PHPSTORM
+      # This should correspond to the server declared in PHPStorm `Preferences | Languages & Frameworks | PHP | Servers`
+      # Then PHPStorm will use the corresponding path mappings
+      PHP_IDE_CONFIG: serverName=api-platform
+```
+
+## Troubleshooting
+
+Inspect the installation with the following command. The requested Xdebug
+version should be displayed in the output.
+
+```bash
+$ docker-compose exec php php --version
+
+PHP 7.2.8 (cli) (built: Jul 21 2018 08:09:37) ( NTS )
+Copyright (c) 1997-2018 The PHP Group
+Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
+    with Zend OPcache v7.2.8, Copyright (c) 1999-2018, by Zend Technologies
+    with Xdebug v2.6.0, Copyright (c) 2002-2018, by Derick Rethans
+```


core/debugging.md
```diff
@@ -0,0 +1,83 @@
...
+```diff
+diff --git a/api/Dockerfile b/api/Dockerfile
+index 745a8f9..432f0d9 100644
+--- a/api/Dockerfile
++++ b/api/Dockerfile
+@@ -1,11 +1,12 @@
+# the different stages of this Dockerfile are meant to be built into separate images
+# https://docs.docker.com/compose/compose-file/#target
+
++ARG BUILD_ENV=prod
```
teohhanhui: Not used.

core/debugging.md
```diff
@@ -0,0 +1,83 @@
...
+ARG PHP_VERSION=7.2
+ARG NGINX_VERSION=1.15
+ARG VARNISH_VERSION=6.0
+
+-FROM php:${PHP_VERSION}-fpm-alpine AS api_platform_php
++FROM php:${PHP_VERSION}-fpm-alpine AS api_platform_php_prod
```
teohhanhui: Please keep this stage untouched, but instead put the `api_platform_php_dev` stage at the very end of the `Dockerfile`.

core/debugging.md
hschaeidt: It's used in line 43, to distinguish between the `prod` and `dev` image. The alternative would be to override the `build target` for the php service.

hschaeidt: The idea was to add a stage before doing all the composer stuff and before the entrypoints are set. Otherwise it would probably also require to re-set the entrypoints as they are set in api_platform_php, for the dev stage too. And that leads to duplicate code. But I'm usually not a docker user, so please correct me if I'm wrong.

core/debugging.md
```diff
@@ -0,0 +1,83 @@
+# Debugging
+
+The default docker stack is shipped without a xdebug/dev stage. It is easy
```
alanpoulain: docker => Docker
xdebug => Xdebug.
Remove `/dev`?
It is => It's
develoment => development
api => API

core/debugging.md
```diff
@@ -0,0 +1,83 @@
...
+though to add [xdebug](https://xdebug.org/) to your project, for develoment
+purposes such as debugging tests or api requests remotely.
+
+## Add a development stage to the Dockerfile
+
+To avoid deploying api-platform to production with an active xdebug extension,
```
alanpoulain: api-platform => API Platform
xdebug => Xdebug
is is => it's

core/debugging.md
teohhanhui: > The alternative would be to override the build target for the php service.

Yes, that is what should be done.

teohhanhui: This dev stage cannot be here. It'll get in the way of other stages. :smile:

It has to be at the end, as I've mentioned in https://github.com/api-platform/docs/pull/560#discussion_r208234307

teohhanhui: You only need to re-set the `ENTRYPOINT` if you're setting `CMD`. If you don't change that, it'll simply use what's already set in the base image.

core/debugging.md
teohhanhui: Also, with this change we don't have to use diff format, but just say to append to the end of the `Dockerfile`.

core/debugging.md
hschaeidt: You are right. It makes sense said like that and keeps things simpler

core/debugging.md
```diff
@@ -0,0 +1,64 @@
+# Debugging
+
+The default Docker stack is shipped without a Xdebug stage. It's easy
+though to add [Xdebug](https://xdebug.org/) to your project, for development
+purposes such as debugging tests or api requests remotely.
```
teohhanhui: api -> API

core/debugging.md
```diff
@@ -0,0 +1,64 @@
...
+it's recommended to add a custom stage to the end of the `api/Dockerfile`.
+
+```Dockerfile
+# api/Dockerfile
+FROM api_platform_php as api_platform_php_dev
+
+ARG XDEBUG_VERSION=2.6.0
+RUN apk add --no-cache $PHPIZE_DEPS; \
+    pecl install xdebug-$XDEBUG_VERSION; \
+    docker-php-ext-enable xdebug
```
teohhanhui:
```Dockerfile
ARG XDEBUG_VERSION=2.6.0
RUN set -eux; \
	apk add --no-cache --virtual .build-deps $PHPIZE_DEPS; \
	pecl install xdebug-$XDEBUG_VERSION; \
	docker-php-ext-enable xdebug; \
	apk del .build-deps
```

core/debugging.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+## Configure Xdebug with docker-compose override
+
+Using a [override](https://docs.docker.com/compose/reference/overview/#specifying-multiple-compose-files)
```
teohhanhui: a -> an

core/debugging.md
```diff
@@ -0,0 +1,64 @@
...
+# api/Dockerfile
+FROM api_platform_php as api_platform_php_dev
+
+ARG XDEBUG_VERSION=2.6.0
+RUN apk add --no-cache $PHPIZE_DEPS; \
+    pecl install xdebug-$XDEBUG_VERSION; \
+    docker-php-ext-enable xdebug
+```
+
+## Configure Xdebug with docker-compose override
```
teohhanhui: Configure Xdebug with Docker Compose Override

(since we use title case)

core/debugging.md
```diff
@@ -0,0 +1,64 @@
+# Debugging
+
+The default Docker stack is shipped without a Xdebug stage. It's easy
+though to add [Xdebug](https://xdebug.org/) to your project, for development
+purposes such as debugging tests or api requests remotely.
+
+## Add a development stage to the Dockerfile
```
teohhanhui: Add a Development Stage to the Dockerfile

(since we use title case)

core/debugging.md
```diff
@@ -0,0 +1,64 @@
...
+  php:
+    build:
+      target: api_platform_php_dev
+    environment:
+      # See https://docs.docker.com/docker-for-mac/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-host
+      # The `remote_host` below may optionally be replaced with `remote_connect_back`
+      #
+      # Choose some other port than `9000`, as this one is already occupied by FPM
+      # Make sure to also change your listening port in the IDE to this value
+      XDEBUG_CONFIG: remote_enable=1
```
teohhanhui:
```yaml
      XDEBUG_CONFIG: >-
        remote_enable=1
        ...
```

(folded scalar, with "strip" block chomping)

core/debugging.md
```diff
@@ -0,0 +1,64 @@
...
+
+Using a [override](https://docs.docker.com/compose/reference/overview/#specifying-multiple-compose-files)  
+ file named `docker-compose.override.yml` ensures that the production
+configuration remains untouched.
+
+As example, an override could look like this:
+
+```yml
+version: '3.4'
+ services:
```
teohhanhui: wrong indentation (remove leading space)

core/debugging.md
```diff
@@ -0,0 +1,64 @@
...
+        remote_port=9999
+        idekey=PHPSTORM
+      # This should correspond to the server declared in PHPStorm `Preferences | Languages & Frameworks | PHP | Servers`
+      # Then PHPStorm will use the corresponding path mappings
+      PHP_IDE_CONFIG: serverName=api-platform
+```
+
+## Troubleshooting
+
+Inspect the installation with the following command, the requested Xdebug
```
teohhanhui: Inspect the installation with the following command. The requested Xdebug version should be displayed in the output.

hschaeidt:
Thanks for this in-depth review. I will fix the rest soon.

dunglas:
Thanks @hschaeidt. Very useful PR!

PR #573, opened 2018-08-29
Getting Started external link broken

DanielSchwiperich:
there's a whitespace causing the link for __'the Symfony Serializer's configuration files or annotations.'__ to be rendered not absolute on api-platform.com. github is more forgiving so the link works here when rendered as markdown. 

See broken link: https://api-platform.com/docs/core/getting-started


dunglas:
Thanks!

PR #576, opened 2018-09-02
Feature/document api respond parameters

Jibbarth:
Hello,

As discussed in Slack, I create this PR to document a little bit the `_api_respond` parameters. I added it in the events page, but I don't know if it's enough or should I add more info.

I also add a big part on the DTO page, to explain how use it without an `ApiResource` declared.

My initial need was to retrieve a Settings custom in my application, but I didn't think it was a good example. 
I thought also to propose the retrievement of the `kernel.environment` parameter, but I didn't find it pertinent. 
Finally, browsing the [symfony/demo](https://github.com/symfony/demo) repository, I found the Locales list, so I based my example on it. 
Let me know if the example is pertinent or not. 

Thanks all of you for your awesome works !

### DIFF
--- a/core/dto.md
+++ b/core/dto.md
@@ -9,6 +9,7 @@ someone has lost its password.
 So let's create a basic DTO for this request:
 
 ```php
+<?php
 // api/src/Api/Dto/ForgotPasswordRequest.php
 
 namespace App\Api\Dto;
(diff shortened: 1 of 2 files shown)


dunglas:
Thanks @Jibbarth. I made some slight modifications to the text. Such PRs are very appreciated!

dvc:
Typo: 
`public function __invoke(): LocalesDTO` 
should be 
`public function __invoke(): LocalesList`

Issue #581, opened 2018-09-05
Improve "Error starting userland proxy" Trobuleshooting


alexfinnarn:
https://github.com/api-platform/docs/blob/2.3/extra/troubleshooting.md#using-docker

The only thing I can think to add is information on how to check what processes are listening on which ports. 

```bash
# Find out if any service listens on port 80.
sudo lsof -n -i :80 | grep LISTEN
```

Relates to https://github.com/api-platform/api-platform/issues/821

PR #582, opened 2018-09-05
Update troubleshooting.md

alexfinnarn:
#581


alexfinnarn:
Yeah, my first commit was more of a "I've had this tab open for too many days" type of thing and I was going to go back and add the Linux one. I think `lsof` is for UNIX-based systems, based on the man page, and I added what appears to be the way to do it on Windows.

https://stackoverflow.com/questions/48198/how-can-you-find-out-which-process-is-listening-on-a-port-on-windows

PR #588, opened 2018-09-20
Custom serialization groups for custom operations

bpolaszek:


dunglas:
Awesome, thanks @bpolaszek!

PR #589, opened 2018-09-21
Custom operation: enhancing the visibility of an important remark

Deuchnord:
The information about necessity that the `__invoke()` method’s parameter must be called `$data` was easy to miss in the large amount of information. This proposal makes it more visible by putting it in its own paragraph and by writing it in bold.


dunglas:
Thanks @Deuchnord. I just added a link to the appropriate Symfony documentation entry.

PR #590, opened 2018-09-23
Update file-upload.md

sedatsevgili:
Fixes missing import of ApiProperty class in "Linking a MediaObject Resource to Another Resource" example.


dunglas:
Good catch, thanks!

PR #591, opened 2018-09-26
[schema-generator] Semi full cardinality explaination

giorgioma:


giorgioma:
Does gatsby support markdown tables?
I am not familiar with it, but I am guessing that is the error

alanpoulain:
Cardinality is not necessarily related to Doctrine, I don't think we should add this information like this.
Also the Doctrine attribute is not chosen with only the cardinality.

PR #592, opened 2018-09-26
RichTextInput is not part of react-admin.

Federkun:
RichTextInput is not part of react-admin. Import from ra-input-rich-text instead.

### DIFF
--- a/admin/getting-started.md
+++ b/admin/getting-started.md
@@ -198,7 +198,8 @@ Example to add a minLength validator on the `description` field:
 
 ```javascript
 import React, { Component } from 'react';
-import { minLength, RichTextInput } from 'react-admin';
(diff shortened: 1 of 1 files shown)


dunglas:
The changes in the file upload docs look not related. Can you remove these commits?

Federkun:
Sure thing.
I just noticed that the file upload docs related to the admin section seems outdated. Took me a while to find https://github.com/api-platform/admin/pull/92 and pull it off; I may submit another PR to provide a consisted example on what's present on the [core](https://api-platform.com/docs/core/file-upload/) section.

dunglas:
Thanks @Federkun.

>  I may submit another PR to provide a consisted example on what's already present on the core section

It would be very appreciated!

PR #596, opened 2018-10-02
fix not working code

x1wins:
1. change ReactDom to ReactDOM https://stackoverflow.com/questions/33125669/uncaught-referenceerror-reactdom-is-not-defined
2. change foo to book
3. add bash code block


dunglas:
Thanks for fixing this!

PR #598, opened 2018-10-05
[serializer] Mapping path when using yaml-xml

ahmed-hachena:
see [#184 (comment)](https://github.com/api-platform/api-platform/issues/184#issuecomment-334961642)


alanpoulain:
Thank you @ahmed-hachena, it will be useful!

Issue #602, opened 2018-10-06
[GraphQL] Remove the cache after requiring graphql-php


alanpoulain:
The route `/graphql` is not found after requiring graphql-php. The cache needs to be removed first.

PR #604, opened 2018-10-06
Add the cache clear command to the Graphql section

Pandoraaa:
Fix issue https://github.com/api-platform/docs/issues/602 
Following the instructions on the GraphQL part, the route ```/graphql``` is not found after requiring graphql-php. So I added the cache:clear command in order to fix it.


soyuka:
Thanks @Pandoraaa !

PR #606, opened 2018-10-06
Install yarn in before_install

alanpoulain:
`yarn` is not installed by Travis because there is no `yarn.lock` or `package.json`.


.travis.yml
```diff
@@ -1,4 +1,6 @@
-language: node
+language: node_js
+node_js:
+  - '6'
```
soyuka: Can we use LTS instead?

alanpoulain:
@soyuka I think it's mergeable now.

soyuka:
Thanks @alanpoulain

PR #609, opened 2018-10-08
Fix minor GraphQL and Subresource issues

alanpoulain:


core/graphql.md
```diff
@@ -12,9 +12,9 @@ Once enabled, you have nothing to do: your schema describing your API is automat
 
 ## Enabling GraphQL
 
-To enable GraphQL and GraphiQL interface in your API, simply require the graphql-php package using Composer:
+To enable GraphQL and GraphiQL interface in your API, simply require the [graphql-php](https://webonyx.github.io/graphql-php/) package using Composer and clear the cache one more time:
```
soyuka: why `one more time`?

alanpoulain: Because `composer require` is doing it once and it has to be done twice in order to work.

soyuka: mhh okay, in the sentence I guess we can still skip it and just say "clear the cache" no?

alanpoulain: But the user will wonder why, doesn't it? I thought we could make it explicit.
I could reformulate the sentence like this:
`and clear the cache (it needs to be cleared twice)`

dunglas: I don't understand why it should be cleared twice. Isn't it mandatory to use `rm -Rf var/cache/*`?

alanpoulain: No it is not. Clearing the cache twice seems to suffice. I'm not really sure why though. Maybe an issue with the autoloader.

dunglas:
Thanks @alanpoulain (feel free to merge such PRs by yourself btw)

PR #614, opened 2018-10-11
use for ex "/api_doc" instead "/doc" for path

msitruk:
In case you need to "Manually Registering the Swagger UI Controller", use another path than "/doc" because that make conflict with Symfony and produce a "Serialization for the format html is not supported." error.


comxd:
In my case I use `/docs/`, note the trailing slash.

alanpoulain:
Superseded by https://github.com/api-platform/docs/pull/1604.

PR #618, opened 2018-10-15
Added Symfony rewrite rules link

Sieberkev:
When installing API Platform via Symfony Flex and Composer on an Apache or Nginx server, the rewrite rules need to be enabled or you'll get a lot of 404 errors. I've added a line of information and the link to the official Symfony documentation with all the relevant information.


dunglas:
Thanks @Sieberkev!

PR #619, opened 2018-10-16
Added example usage of react-admin Admin component

helios-ag:
Added information about how to build Admin app on top of react-admin's Admin component,
and show example of usage custom components for create, edit, show and list actions.


dunglas:
Thank you very much @helios-ag. Nice addition!

PR #620, opened 2018-10-16
Adds an example of modifying items by listening to the POST_READ event

yaffol:
I was trying to modify an API response to add a dynamically generated image URL to any Image items returned from the API. I tried to modify the example BookSubscriber to do this, but struggled to find out how to obtain the item(s) that would be returned by the API from a read request. @TimoBakx helped me out, and this PR adds an example of how to do this.

### DIFF
--- a/core/events.md
+++ b/core/events.md
@@ -71,6 +71,73 @@ Alternatively, [the subscriber must be registered manually](http://symfony.com/d
 [Doctrine events](http://doctrine-orm.readthedocs.org/en/latest/reference/events.html#reference-events-lifecycle-events)
 are also available (if you use it) if you want to hook at the object lifecycle events.
 
+Different events require different methods for obtaining the item(s) created or retrieved by the api request. 
+In the following example, we will add a dynamically generated image URL to each image item, whether individual or part of 
+a collection, that is returned by the API:
+
+```php
+<?php
+// api/src/EventSubscriber/ImageUrlSubscriber.php
+
+namespace App\EventSubscriber;
+
+
+use ApiPlatform\Core\EventListener\EventPriorities;
+use App\Entity\ImageInterface;
+use App\Service\Images;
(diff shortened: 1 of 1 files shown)


core/events.md
```diff
@@ -71,6 +71,73 @@ Alternatively, [the subscriber must be registered manually](http://symfony.com/d
 [Doctrine events](http://doctrine-orm.readthedocs.org/en/latest/reference/events.html#reference-events-lifecycle-events)
 are also available (if you use it) if you want to hook at the object lifecycle events.
 
+Different events require different methods for obtaining the item(s) created or retrieved by the api request. 
+In the following example, we will add a dynamically generated image URL to each image item, whether individual or part of 
+a collection, that is returned by the API:
+
+```php
+<?php
+// api/src/EventSubscriber/ImageUrlSubscriber.php
```
TimoBakx: I think starting from `src/` would be more clear.

TimoBakx: Never mind, I just saw that this is the standard.

dunglas:
Hi, and thanks for contributing to improve the docs, this kind of changes are very appreciated.
As for #622, don't you think a normalizer decorator would be a better place to this? Then the code will not be coupled to the HTTP layer anymore, and will be reusable, for instance if you want to send the document to a worker using the Messenger component, or to another internal service using Kafka or a RabbitMQ bus. WDYT?

patrickvale:
Thanks for the feedback. As to #622 @TimoBakx would have to answer that.

TimoBakx:
I agree with @dunglas on this. A decorator would work better for most cases. The only case in which a `POST_READ` listener would be a better option is if you need the data during other parts of handling the request (other listeners further down the line or a controller in case of a custom action). In which case, the question is if you shouldn't put the logic in that spot in the first place.

That being said, I have yet to rewrite and test my part of it to use the decorator instead of a listener.

PR #622, opened 2018-10-17
Added an example about appending data in bulk

TimoBakx:
A while ago, I had some problems appending data pulled from an external source to a collection using a normalizer decorator. It was very slow when pulling bigger collections.
Working with a listener worked out better, as I could pull the data for the entire collection in 1 query.

After talking with @patrickvale about #620, I thought that adding an altered example of the same idea with the focus on performance would be a good idea.

### DIFF
--- a/core/performance.md
+++ b/core/performance.md
@@ -265,7 +265,7 @@ api_platform:
 
 More details are available on the [pagination documentation](pagination.md#partial-pagination).
 
-### Appending data in bulk
+### Appending Data in Bulk
 
 As stated in the [serialization documentation](serialization.md#decorating-a-serializer-and-adding-extra-data), you can use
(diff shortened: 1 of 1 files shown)


dunglas:
Hi, thanks for contributing this change.
However, I'm not sure that it's the cleanest approach. Wouldn't moving this logic from a listener to a collection normalizer decorator be more reusable?

TimoBakx:
I agree. I haven't seen an example of a collection normalizer though. Is there one in the documentation? Or do you have a code example I can use to rewrite this documentation?

A regular normalizer would iterate over each item individually and would therefore generate too much traffic between the normalizer and the source of the data (database or (in my case) an external NAV server).

Thank you in advance.

PR #624, opened 2018-10-17
Update security doc

jewome62:
Add sentence to redirect to extension to restrict a collection

### DIFF
--- a/core/security.md
+++ b/core/security.md
@@ -121,5 +121,7 @@ App\Entity\Book:
     # ...
 ```
 
-Into access control expression for collection, `object` property is not available.
-For add restriction on this, you need implement [a extension](extensions.md) to make a custom query.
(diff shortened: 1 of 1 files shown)


core/security.md
```diff
@@ -120,3 +120,6 @@ App\Entity\Book:
             access_control_message: 'Sorry, but you are not the book owner.'
     # ...
 ```
+
+Into access control expression for collection, `object` property is not available.
```
dunglas: Actually it is, but it is the collection instance.

core/security.md
```diff
@@ -120,3 +120,6 @@ App\Entity\Book:
...
+For add restriction on this, you need implement [a extension](extensions.md) to make a custom query.
```
dunglas: to add restrictions? you need to?

dunglas:
Thanks @jewome62. I made some slight modifications.

PR #627, opened 2018-10-20
:pencil2: fix (core-file-upload): remove unused imports from CreateMediaObjectAction

GregoireHebert:


dunglas:
Thanks @GregoireHebert

PR #633, opened 2018-10-26
Less complex normalizer

TimoBakx:
I was implementing this example today and I noticed that it is possible to use `NormalizerAwareInterface` instead of `SerializerAwareInterface`. By using the first interface and corresponding trait, the `instanceof NormalizerInterface` check is no longer needed.

I also shortened the constant (for readability) and removed the now quite small `passOn` method to use the code inside directly in the `normalize` method.

### DIFF
--- a/core/serialization.md
+++ b/core/serialization.md
@@ -438,7 +438,7 @@ services:
 ```
 
 The Normalizer class is a bit harder to understand, because it must ensure that it is only called once and that there is no recursion. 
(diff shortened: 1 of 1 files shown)


dunglas:
As this is a fix, can you rebase against 2.3 please @TimoBakx?

TimoBakx:
Rebased and retargeted to 2.8.

dunglas:
Thanks @TimoBakx

PR #634, opened 2018-10-28
add @ORM\Table(name="fos_user") to user entity

NMathar:
this is to not use the default protected word User for an sql table https://symfony.com/doc/master/bundles/FOSUserBundle/index.html#a-doctrine-orm-user-class


dunglas:
Good catch, thanks!

PR #638, opened 2018-11-01
Examples for custom controllers

martinharder:
Many are missing examples how to use pagination with custom controllers.
This could help.

### DIFF
--- a/core/pagination.md
+++ b/core/pagination.md
@@ -424,7 +424,7 @@ class GetBooksByFavoriteAuthorAction extends Controller
 {
     public function __invoke(Request $request, BookService $bookService): Paginator
     {
-        $page = (int)$request->query->get('page', 1);
+        $page = (int) $request->query->get('page', 1);
 
         return $bookService->getBooksByFavoriteAuthor($page);
     }
--- a/core/pagination.md
+++ b/core/pagination.md
@@ -345,3 +345,143 @@ class Book
 ```
 
 Please note that this parameter will always be forced to false when the resource have composite keys due to a [bug in doctrine](https://github.com/doctrine/doctrine2/issues/2910)
+
+## Custom Controller Action
+
+In case you're using a custom controller action make sure you return the Paginator object to get the full hydra response with hydra:view (which contains information about first, last, next and previous page). The following examples show how to handle it within a service method which is called in the controller. You will need to use the Doctrine Paginator and pass it to the ApiPlatform Paginator.
+
+First example:
+
+```php
+<?php
+
+//src/Service/Book/BookService.php
+
+namespace App\Service\Book;
+
+use App\Entity\Book;
+use App\Entity\User;
+use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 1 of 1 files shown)


core/pagination.md
```diff
@@ -345,3 +345,127 @@ class Book
 ```
 
 Please note that this parameter will always be forced to false when the resource have composite keys due to a [bug in doctrine](https://github.com/doctrine/doctrine2/issues/2910)
+
+## Custom Controller Action
+
+In case you're using a custom controller action make sure you return the Paginator object to get the full hydra response with hydra:view (which contains information about first, last, next and previous page). The following examples show how to handle it within a service method which is called in the controller. You will need to use the Doctrine Paginator and pass it to the ApiPlatform Paginator.
```
dunglas:
```suggestion
In case you're using a custom controller action make sure you return the `Paginator` object to get the full hydra response with `hydra:view` (which contains information about first, last, next and previous page). The following examples show how to handle it within a service method which is called in the controller. You will need to use the Doctrine Paginator and pass it to the API Platform Paginator.
```

core/pagination.md
```diff
@@ -345,3 +345,127 @@ class Book
...
+
+use App\Entity\Book;
+use App\Entity\User;
+use Doctrine\ORM\EntityManagerInterface;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+use Doctrine\ORM\Tools\Pagination\Paginator as DoctrinePaginator;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Paginator;
+use Doctrine\Common\Collections\Criteria;
+
+class BookService
```
dunglas: Could be the `BookRepository` directly right?

core/pagination.md
```diff
@@ -345,3 +345,127 @@ class Book
...
+// src/Controller/Book/GetBooksByFavoriteAuthorAction.php
+
+namespace App\Controller\Book;
+
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Paginator;
+use App\Service\Book\BookService;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\Request;
+
+class GetBooksByFavoriteAuthorAction extends Controller
```
dunglas: Should be `AbstractController` IIRC

core/pagination.md
```diff
@@ -345,3 +345,127 @@ class Book
...
+{
+    public function __invoke(Request $request, BookService $bookService): Paginator
+    {
+        $page = (int)$request->query->get('page', 1);
```
dunglas:
```suggestion
        $page = (int) $request->query->get('page', 1);
```

core/pagination.md
martinharder: Sure.

dunglas:
Thanks @martinharder!

martinharder:
You're welcome!

PR #640, opened 2018-11-02
Add a yaml example for specifying an identifier.

### LINKED ISSUE #639: How do you specific an identifier in yaml
The documentation gives an example for defining an identifier with annotations only - https://api-platform.com/docs/core/serialization/#entity-identifier-case

How do you specify an identifier with yaml? I'm happy to update the docs if someone can tell me how to specify it.

trsteel88:
Closes #639

### DIFF
--- a/core/serialization.md
+++ b/core/serialization.md
@@ -642,6 +642,15 @@ class Book
 }
 ```
 
+You can also use the YAML configuration format:
+
+```yaml
+# api/config/api_platform/resources.yaml
+App\Entity\Book:
+    properties:
+        identifier: [id]
+```
+
 In some cases, you will want to set the identifier of a resource from the client (e.g. a client-side generated UUID, or a slug).
 In such cases, you must make the identifier property a writable class property.  Specifically, to use client-generated IDs, you 
 must do the following:
--- a/core/serialization.md
+++ b/core/serialization.md
@@ -648,7 +648,8 @@ You can also use the YAML configuration format:
 # api/config/api_platform/resources.yaml
 App\Entity\Book:
     properties:
-        identifier: [id]
+        id:
+            identifier: true
 ```
 
 In some cases, you will want to set the identifier of a resource from the client (e.g. a client-side generated UUID, or a slug).


trsteel88:
Actually, I don't think this is correct. I still get the warning ```Unable to generate an IRI for the item of type \"App\\Api\\Model\\MyNonDoctrineModel\"```

antograssiot:
Do you have a getter on your property and enable the get on item operation ?

trsteel88:
Yes.

```php
final class MyNonDoctrineModel
{
    /**
     * @var string
     */
    private $id;

    /**
     * @var Organisation
     */
    private $organisation;

    /**
     * @var Template
     */
    private $template;

    /**
     * @var float
     */
    private $price;

    public function __construct(Organisation $organisation, Template $template)
    {
        $this->organisation = $organisation;
        $this->template = $template;
        $this->id = $this->organisation->getId().'-'.$this->template->getId();
    }

    public function getId(): string
    {
        return $this->id;
    }
```

trsteel88:
My resource looks like this

```yaml
App\Api\Model\ MyNonDoctrineModel:
    properties:
        identifier: [id]
    attributes:
        access_control: "is_granted('ROLE_USER')"
    itemOperations:
        get:
            method: GET
            path: /organisations-templates/{id}
            defaults:
                _api_receive: false
            controller: App\Api\Action\OrganisationTemplatePrice\ViewAction
    collectionOperations:
        get:
            method: GET
            path: /organisations-templates
            controller: App\Api\Action\OrganisationTemplatePrice\ListAction
```

It's definitely a configuration issue with defining the identifiers because what I have configured are not coming through when I do a dump within IdentifiersExtractor.php

If I override the array of identifiers within that code it works.

antograssiot:
isn't that 
```
App\Api\Model\MyNonDoctrineModel:
    properties:
        id:
            identifier: true
...
```

instead ?

trsteel88:
I'm away from my computer at the moment. But I am pretty sure I tried that
and it didn't work either.

trsteel88:
You were right @antograssiot. Thanks!

dunglas:
Thanks a lot @trsteel88 and @antograssiot, this kind of enhancements in the docs are very helpful!

trsteel88:
You're welcome!

PR #641, opened 2018-11-04
Fixes api-platform/api-platform#884

antograssiot:
It seems that showcasing annotation is not always enough, users sometime struggle with YAML or XML.
This might help...


dunglas:
Thank you very much!

PR #646, opened 2018-11-07
Document input/output resource attribute

soyuka:
https://github.com/api-platform/core/pull/2235 ping @silverbackdan

### DIFF
--- a/core/dto.md
+++ b/core/dto.md
@@ -2,6 +2,8 @@
 
 ## How to Use a DTO for Writing
 
+!> The following isn't recommended anymore, please use [Input/Output](specify-an-input-and-or-output-class) instead
+
 Sometimes it's easier to use a DTO than an Entity when performing simple
 operation. For example, the application should be able to send an email when
 someone has lost its password.
@@ -125,6 +127,8 @@ services:
 ```
 
 ## How to Use a DTO for Reading
+
+!> The following isn't recommended anymore, please use [Input/Output](specify-an-input-and-or-output-class) instead
 
 Sometimes, you need to retrieve data not related to an entity.
 For example, the application can send the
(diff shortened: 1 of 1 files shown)


core/dto.md
```diff
@@ -350,3 +350,73 @@ use Swagger\Annotations as SWG;
      */
     public function __invoke()
 ```
+
+### Specify an input and/or output class
```
dunglas:
```suggestion
### Specify an Input and/or Output Class
```

core/dto.md
```diff
@@ -350,3 +350,73 @@ use Swagger\Annotations as SWG;
...
+// api/src/Entity/Data.php
+
+namespace App\Entity;
+
+use ApiPlatform\Core\Annotation\ApiResource;
+use App\Dto\Input;
+use App\Dto\Output;
+
+/**
+ * @ApiResource(attributes={
```
dunglas: You could use the shortcuts `inputClass` and `outputClass` directly

core/dto.md
```diff
@@ -350,3 +350,73 @@ use Swagger\Annotations as SWG;
...
+class InputDataPersister implements DataPersisterInterface
+{
+    public function supports($data): bool
+    {
+        return $data instanceof Input;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
```
dunglas: Could be remove in the example

core/dto.md
```diff
@@ -350,3 +350,73 @@ use Swagger\Annotations as SWG;
...
+```php
+<?php
+
+namespace App\DataPersister;
+
+use ApiPlatform\Core\DataPersister\DataPersisterInterface;
+use App\Dto\Input;
+use App\Dto\Output;
+
+class InputDataPersister implements DataPersisterInterface
```
dunglas:
```suggestion
final class InputDataPersister implements DataPersisterInterface
```

core/dto.md
```diff
@@ -350,3 +350,73 @@ use Swagger\Annotations as SWG;
...
+
+        return $output;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function remove($data)
+    {
+        return null;
```
dunglas:
```suggestion
        // TODO: implement removal
        return null;
```

dunglas:
Shouldn't we remove the previous sections in the same file (https://api-platform.com/docs/core/dto/#how-to-use-a-dto-for-writing).
Using your solution is much better and cleaner, so we should recommend always using it.

soyuka:
Yeah we should def remove the previous sections, but we need swagger support first :|. 
I'll add a deprecation sentence.

PR #650, opened 2018-11-09
Config file path correction

silverbackdan:
The framework form configuration should be in the `api/config/packages/framework.yaml` file, not `api/config/packages/api_platform.yaml`


silverbackdan:
Sure, can do. Is it worth keeping the framework config for SF3 users and adjusting the text accordingly? I'm happy to delete if not, thought it may be worth asking.

PR #653, opened 2018-11-13
Remove section on troubleshooting JMS Serializer

teohhanhui:
As it is no longer relevant with the latest stable version of JMSSerializerBundle

### DIFF
--- a/extra/troubleshooting.md
+++ b/extra/troubleshooting.md
@@ -23,15 +23,3 @@ netstat -a -b
 ```
 
 You can change the port to be used in the `docker-compose.yml` file (default is port 80).
-
-## Using API Platform and JMS Serializer in the same project
-
-By default, [JMS Serializer Bundle](http://jmsyst.com/bundles/JMSSerializerBundle) replaces the `serializer` service by its own. However, API Platform requires the Symfony serializer (and not the JMS one) to work properly.
-Fortunately, this behavior can be deactivated using the following configuration:
-
-```yaml
-# api/config/packages/api_platform.yaml
-
-jms_serializer:    
-    enable_short_alias: false
-```


dunglas:
Maybe can we just indicate for which version it is relevant instead on dropping the section?

teohhanhui:
But it was for v1, which is very old by now. [2.0.0](https://github.com/schmittjoh/JMSSerializerBundle/releases/tag/2.0.0) was released on May 17, 2017. [3.0.0](https://github.com/schmittjoh/JMSSerializerBundle/releases/tag/3.0.0) was released 4 days ago.

teohhanhui:
v1 of JMSSerializerBundle is not compatible with Symfony 4 anyway, so I think we can safely drop this section.

teohhanhui:
When you wrote the section almost 2 years ago (https://github.com/api-platform/docs/pull/161), JMSSerializerBundle was still on v1.

dunglas:
We still support Symfony 3.4. Suggesting to upgrade JMSSerializer is ok, but I would like to keep the solution for v1, it could be useful when you are in the process of upgrading an old app.

teohhanhui:
Updated.

extra/troubleshooting.md
```diff
@@ -26,12 +26,15 @@ You can change the port to be used in the `docker-compose.yml` file (default is
...
 
-By default, [JMS Serializer Bundle](http://jmsyst.com/bundles/JMSSerializerBundle) replaces the `serializer` service by its own. However, API Platform requires the Symfony serializer (and not the JMS one) to work properly.
-Fortunately, this behavior can be deactivated using the following configuration:
+For the latest versions of [JMSSerializerBundle](http://jmsyst.com/bundles/JMSSerializerBundle), there is no conflict so everything should work out of the box.
+
+In an older version (v1) of JMSSerializerBundle, the `serializer` alias is registered for the JMS Serializer service by default. However, API Platform requires the Symfony Serializer (and not the JMS one) to work properly. Fortunately, this behavior can be deactivated using the following configuration:
 
 ```yaml
-# api/config/packages/api_platform.yaml
+# app/config/config.yml
```
teohhanhui: Because it's JMSSerializerBundle we're talking about (very old and incompatible with Symfony 4), I don't think it makes sense to refer to the Flex directory structure here.

teohhanhui:
Rebased and it's green now. :green_apple: :green_heart:

PR #656, opened 2018-11-14
Document deprecations

dunglas:
Adapted from https://dunglas.fr/2018/07/api-platform-2-3-major-perf-improvement-api-evolution-deprecation-better-dev-tools-and-much-more/


dunglas:
It's automatic now. It should just be added in https://github.com/api-platform/website/blob/master/config/outline.yaml after the merge (it could be nice to move this file in this repo btw).

PR #659, opened 2018-11-15
Fix XML Mapping for normalization_context groups

dpfaffenbauer:
`<group>` is invalid, it throws an XML Validation exception

```
ERROR 1871] Element '{https://api-platform.com/schema/metadata}group': This element is not expected.
```


dunglas:
Good catch, thanks @dpfaffenbauer

PR #665, opened 2018-11-17
use correct format for di const values

bendavies:


antograssiot:
Seems a bit app specific maybe.
If accepeted, `UniqueConstraintViolationException` would also be a good condidate I think.

antograssiot:
Totaly. Sorry shouldn't be doing that before coffee. :rofl: 
 Sorry

dunglas:
Thanks @bendavies!

PR #668, opened 2018-11-18
Update operations.md

matias-pizarro:
Replace the base Controller class with AbstractController (Controller class is deprecated), remove an unused use statement and delete the 'Action' suffix in the method names.


dunglas:
Thanks for keeping the docs up to date, it's very appreciated!

PR #676, opened 2018-11-30
Troubleshooting error 502 "upstream sent too big header while reading response header from upstream"

sushicodeur:


extra/troubleshooting.md
```diff
@@ -35,3 +35,58 @@ Fortunately, this behavior can be deactivated using the following configuration:
...
+    location ~ ^/index\.php(/|$) {
+        # Comment the next line and uncomment the next to enable dynamic resolution (incompatible with Kubernetes)
+        fastcgi_pass php:9000;
+        #resolver 127.0.0.11;
+        #set $upstream_host php;
+        #fastcgi_pass $upstream_host:9000;
+        
+        # Bigger buffer size to handle cache invalidation headers expansion
+        fastcgi_buffer_size          128k;
+        fastcgi_buffers              4 256k;
```
teohhanhui: The value used here for `fastcgi_buffers` seems very atypical, and is probably a complete overkill for most cases.

dunglas: Do you suggest to use a lower value by default? The size of header can grow very fastly when the invalidation mechanism is enabled...

Simperfit: I think 32k is enough in most of case, 256k seems very very high.

teohhanhui: Something like:

```
fastcgi_buffer_size 32k;
fastcgi_buffers 8 16k;
```

@dunglas:
> The size of header can grow very fastly when the invalidation mechanism is enabled...

We often forget how big a kilobyte actually is when it comes to text. :smile:

teohhanhui: If the response header grows past 32 KB because of the cache tags, I think it's a **feature** for it to not work, because you have a very serious problem. :laughing:

sushicodeur: Sorry for this strange buffer values, it doesn't mean a lot to me, so I though the bigger the better :)

Should this troubleshooting put an accent on the `shortName` annotation to have shorter urls as a way to reduce the size of sent headers ?

dunglas:
Merging as is, lets tweak the value in a next PR.

dunglas:
Thanks @sushicodeur!

PR #677, opened 2018-12-07
Missing configuration to connect to xdebug from VSCode

AlanCrevon:
See https://xdebug.org/docs/all_settings#remote_connect_back for more information.


dunglas:
Thanks @AlanCrevon

PR #681, opened 2018-12-15
Replace @type tag by @var tag in the phpdoc

aaa2000:
With the @type tag, there is the error 

```
[Semantical Error] The annotation "@type" in property App\Entity\Person::$code was never imported. Did you maybe forget to add a "use" statement for this annotation?
```


dunglas:
Thanks!

PR #683, opened 2018-12-16
host & schemes are not options

ghanbari:
host & schemes are not options, they are direct route attributes.


dunglas:
Thanks @ghanbari

PR #684, opened 2018-12-17
name field is not defined in offer

ghanbari:
in previous example, offer dont has a name field, but have description.
description's (or name) filter type is partial then it can not get multi value and filter with them.


dunglas:
Thanks @ghanbari

PR #686, opened 2018-12-18
Update heroku.md

younesdiouri:
Without composer symfony/apache-pack, you cannot access any route of your app in production.


dunglas:
Thanks @younesdiouri

PR #687, opened 2018-12-21
Removal of deprecated extractProperties method in docs

ThomasBerends:
Based on a conversation in Slack I found out that extractProperties is not supported anymore. Using the code in the documentation will do nothing.


dunglas:
Thanks @ThomasBerends. And indeed we need to document this @soyuka! Would you mind doing the change or opening an issue?

Issue #690, opened 2018-12-23
Calling isMethodSafe() throws an error in the Form Data doc


BPagoaga:
From https://github.com/api-platform/docs/blob/2.3/core/form-data.md, the line:

`if ($request->isMethodSafe() || $request->isMethod(Request::METHOD_DELETE)) {`

throws an error since symfony 4.1: 

`Checking only for cacheable HTTP methods with Symfony\Component\HttpFoundation\Request::isMethodSafe() is not supported.`

using `isMethodSafe(false)` solves the problem.

PR #691, opened 2018-12-23
Call Request::isMethodSafe(false) instead of Request::isMethodSafe()

BPagoaga:
Solving #690 

According to https://github.com/symfony/http-foundation/blob/master/Request.php#L1429, `isMethodSafe` should be passed `false` as an argument.


soyuka:
Thanks @BPagoaga

PR #695, opened 2019-01-02
[ADMIN] Add missing code in example

mnlkrs:
Must've been forgotten:
- `import { RichTextField } from 'react-admin';` is imported but never actually used
- "The `field` property of the `Field` class allows to set the component used to render a property in list and show screens." - however, the field property is never set

Code was validated to work.


alanpoulain:
Thank you @mnlkrs

PR #700, opened 2019-01-07
Update vuejs.md

Fabious:
wrong folder name, as seen here : https://github.com/api-platform/client-generator/tree/master/templates/vue


dunglas:
Thanks @Fabious

PR #701, opened 2019-01-07
Simplify swagger api_keys config example

teohhanhui:


core/jwt.md
```diff
@@ -72,9 +72,8 @@ Want to test the routes of your JWT-authentication-protected API?
 # api/config/packages/api_platform.yaml
 api_platform:
     swagger:
-         api_keys:
-             apiKey:
```
teohhanhui: The key is thrown away in any case... :see_no_evil:

teohhanhui: Hmm... No, I was wrong.

PR #703, opened 2019-01-08
Improve the docs for DTOs

dunglas:
Promote the new features introduced in 2.4

### DIFF
--- a/core/dto.md
+++ b/core/dto.md
@@ -13,7 +13,7 @@ namespace App\Entity;
 
 use ApiPlatform\Core\Annotation\ApiResource;
 use App\Dto\BookInput;
-use App\Dto\Output;
+use App\Dto\BookOutput;
 
 /**
  * @ApiResource(
--- a/core/dto.md
+++ b/core/dto.md
@@ -38,14 +38,14 @@ The `input_class` and `output_class` attributes are taken into account by all th
 
 ## Disabling the Input or the Output
 
-Both the `input_class` and the `output_class` attributes can be the to `false`.
-If `input_class` is `false`, the deserialization process will be skipped, and the no data persisters will be called.
+Both the `input_class` and the `output_class` attributes can be set to `false`.
+If `input_class` is `false`, the deserialization process will be skipped, and no data persisters will be called.
 If `output_class` is `false`, the serialization process will be skipped, and no data providers will be called. 
 
 ## Creating a Service-Oriented endpoint
 
 Sometimes it's convenient to create [RPC](https://en.wikipedia.org/wiki/Remote_procedure_call)-like endpoints.
-For example, the application should be able to send an email when someone has lost its password.
+For example, the application should be able to send an email when someone has lost his password.
 
 So let's create a basic DTO for this request:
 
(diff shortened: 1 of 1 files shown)


core/dto.md
```diff
@@ -1,421 +1,130 @@
...
+The `input_class` and `output_class` attributes are taken into account by all the documentation generators (GraphQL and OpenAPI, Hydra).
 
-            $methods = $route->getMethods();
-            $uri = $route->getPath();
+## Disabling the Input or the Output
 
-            foreach ($methods as $method) {
-                // Add available mimesTypes
-                $swaggerContext['produces'] ?? $swaggerContext['produces'] = $mimeTypes;
+Both the `input_class` and the `output_class` attributes can be the to `false`.
```
borNfreee: can be SET to `false`

core/dto.md
```diff
@@ -1,421 +1,130 @@
...
+If `input_class` is `false`, the deserialization process will be skipped, and the no data persisters will be called.
+If `output_class` is `false`, the serialization process will be skipped, and no data providers will be called.
```
borNfreee: here `and no data ...`
on previous line `and the no data...`

I think `the` is not needed on line 42

core/dto.md
```diff
@@ -1,421 +1,130 @@
...
-     *                  }
-     *              }
-     *          }
-     *     }
-     * )
-     */
-    public function __invoke()
-```
+Sometimes it's convenient to create [RPC](https://en.wikipedia.org/wiki/Remote_procedure_call)-like endpoints.
+For example, the application should be able to send an email when someone has lost its password.
```
borNfreee: IMO, `someone has lost its` -> `someone has lots their`

core/dto.md
```diff
@@ -1,421 +1,130 @@
...
+     * @Assert\NotBlank
+     * @Assert\Email
+     */
+    public $email;
 }
 ```
 
-The `input_class` attribute will be used in the denormalization phase, when transforming the user data to a resource instance.
-The `output_class` attribute will be used in the normalization phase, this class represents how the `Data` resource will be given in the `Response`.
+In this case, we disable all operations except `POST`. We also set the `output_class` attribute to `false` to hint
```
borNfreee: is the line break required after `hint` word?

core/dto.md
```diff
@@ -1,421 +1,130 @@
...
+```yaml
+# api/config/services.yaml
+services:
+    # ...
+    'App\DataPersister\ForgotPasswordRequestDataPersister': ~
+        # Uncomment only if autoconfiguration is disabled
+        #tags: [ 'api_platform.data_persister' ]
+```
+
+Instead of a custom data provider, you'll probably want to leverage the Symfony Messenger Component integration.
```
borNfreee: I suppose data persister, not data provider as previously we told about `ForgotPasswordRequestDataPersister`.

Also, could you please add a link for `Symfony Messenger Component integration`. As a user, I would like to click to it in order to compare approaches.

core/dto.md
```diff
@@ -1,421 +1,130 @@
...
-// api/src/Api/Dto/ForgotPasswordRequest.php
+// api/src/Entity/Book.php
 
-namespace App\Api\Dto;
+namespace App\Entity;
 
 use ApiPlatform\Core\Annotation\ApiResource;
-use Symfony\Component\Validator\Constraints as Assert;
+use App\Dto\BookInput;
+use App\Dto\Output;
```
Jibbarth:
```suggestion
use App\Dto\BookOutput;
```

core/dto.md
dunglas: Good catch, I'll add a link... when I'll have written the docs for the messenger integration :D Probably tomorrow

dunglas:
Thanks for the review @borNfreee, changes done!

PR #706, opened 2019-01-08
Document OpenAPI v3 support

dunglas:
https://github.com/api-platform/core/pull/2302

### DIFF
--- a/core/swagger.md
+++ b/core/swagger.md
@@ -1,12 +1,38 @@
-# Swagger / Open API Support
+# OpenAPI Specification Support (formerly Swagger)
 
-API Platform natively support the [Open API](https://www.openapis.org/) (formerly Swagger) API specification format.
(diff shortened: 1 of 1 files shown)


mnapoli:
Hi! Just saw that in the documentation but I cannot get it to work. Is this in the latest stable release?

I found https://github.com/api-platform/core/pull/2302 which added support for OpenAPI v3, it was merged, but I don't know if it is released yet. Since it's documented I'm guessing it is?

dunglas:
It's in the last beta

dunglas:
I chosen to publish on the website the docs for releases **and** pre-releases (beta), but not for `master`, because pre-releases are shown first on the GitHub download page (the recommended way for to download the distribution).

A selector on the website would be great, but would require some work on https://github.com/api-platform/website

PR #709, opened 2019-01-10
fixed broken links

disembarkedone:
fixed links for operations, pagination, filters, content-negotiation

### DIFF
--- a/distribution/index.md
+++ b/distribution/index.md
@@ -440,12 +440,12 @@ You just saved a new book resource through the bookshop API! API Platform automa
 an instance of the corresponding PHP entity class and uses Doctrine ORM to persist it in the database.
 
 By default, the API supports `GET` (retrieve, on collections and items), `POST` (create), `PUT` (update) and `DELETE` (self-explaining)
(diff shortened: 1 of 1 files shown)


dunglas:
The change doesn't look good: it breaks the link when you try: https://github.com/api-platform/docs/blob/78c10bc4daf9faf28872122e429e0502250f5ff1/distribution/index.md

Why do you think it should be changed? It's broken on the website?

disembarkedone:
I was afraid it would break elsewhere.

The links I was hoping to fix do not work here: https://api-platform.com/docs/distribution/

dunglas:
Ok, AFAIU it has been introduced by https://github.com/api-platform/website/pull/91. I see no other fixes than reverting this PR and **not** using slashes for such pages. Otherwise the docs will be broken on Github.

ping @raoulclais @Fabious

dunglas:
An alternative fix would be to append `.html` to URLs corresponding to transformed `.md` pages.

dunglas:
Fixed by https://github.com/api-platform/website/pull/93. Thanks for the report and for helping us to debug @disembarkedone!

PR #711, opened 2019-01-14
Fix Swagger decorator service

oschwartz:


dunglas:
Thanks @oschwartz!

borNfreee:
Isn't it a BC break?

dunglas:
It's not at the code level, but indeed it could wdyt @vincentchalamon?

dunglas:
Related to https://github.com/api-platform/api-platform/issues/938#issuecomment-454001663

antograssiot:
isn't that related to Symfony 4.2.2 also and the changes that create issue with data providers etc ? I tried it from the doc recently and I'm pretty sure it was still working.
Decorating the other normlizer is still valid anyway

dunglas:
Good catch, it's probably related. But it means that we should set the priority of our decorator to `-1` or a similar negative value then.

dunglas:
We really need to change the priorities of all our normalizers...

guilliamxavier:
Xref #722

PR #715, opened 2019-01-17
Wrong exemple file name

jdeniau:
The name of the class is `CartManager` so I renamed `ProductManager.php` to `CartManager.php`


alanpoulain:
Thanks @jdeniau

PR #717, opened 2019-01-17
Elasticsearch documentation

meyerbaptiste:


core/data-providers.md
```diff
@@ -1,12 +1,12 @@
 # Data Providers
 
 To retrieve data exposed by the API, API Platform uses classes called **data providers**. A data provider using [Doctrine
-ORM](http://www.doctrine-project.org/projects/orm.html) to retrieve data from a database is included with the library and
-is enabled by default. This data provider natively supports paged collections and filters. It can be used as is and fits
-perfectly with common usages.
+ORM](http://www.doctrine-project.org/projects/orm.html) to retrieve data from a database and a data provider using
```
dunglas: could we mention Mongo too?

core/filters.md
```diff
@@ -563,6 +563,141 @@ It means that the filter will be **silently** ignored if the property:
...
+ */
+class User
+{
+    // ...
+}
+```
+
+Given that the collection endpoint is `/users`, you can filter users by gender and age.
+
+`/users?gender=female` will return all users whose gender is `female`.
```
dunglas: It could be nice to add an example of how to do a full text search.

dunglas: Adding an example of how to use boosts (if supported) would be nice too

meyerbaptiste: Match filter added. Boosts are not yet supported.

dunglas:
Thanks @meyerbaptiste

PR #718, opened 2019-01-17
Added chapter on how to set up Behat when working outside the Official Distribution

manuelkiessling:


dunglas:
We switched from Behat to PHPUnit. I'll close this one, but feel free to reopen to add a doc explaining how to integrated Behat! Thanks for helping improving the docs.

PR #719, opened 2019-01-19
[Swagger] Don't disable for changing location

Jibbarth:
In a fresh install of ApiPlatform 2.4, I tried to change SwaggerUI location to /swagger

Following the docs give me this error : 

```
Controller "api_platform.swagger.action.ui" does neither exist as service nor as class
```

It seems that there is no need to disable the UI in configuration.

This is my configuration files and all work fine.
```yaml
# config/packages/api_platform.yaml
api_platform:
    mapping:
        paths: ['%kernel.project_dir%/src/Entity']

    enable_swagger_ui: true
    enable_re_doc: false

# config/routes/api_platform.yaml
api_platform:
    resource: .
    type: api_platform
    prefix: /

api_doc:
    path: /swagger
    controller: api_platform.swagger.action.ui
```

### DIFF
--- a/core/swagger.md
+++ b/core/swagger.md
@@ -348,17 +348,7 @@ Again, you can use the `openapi_context` key instead of the `swagger_context` on
 
 ## Changing the Location of Swagger UI
 
-Sometimes you may want to have the API at one location, and the Swagger UI at a different location. This can be done by disabling the Swagger UI from the API Platform configuration file and manually adding the Swagger UI controller.
-
-### Disabling Swagger UI or of ReDoc
-
-```yaml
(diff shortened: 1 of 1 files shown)


Jibbarth:
:thinking: After verification, Swagger UI is available in `/swagger` and `/`, that is not the desired comportment.

I think a bug has been introduced when we move the `api_platform.swagger.action.ui` from `src/Bridge/Symfony/Bundle/Resources/config/swagger.xml` to `src/Bridge/Symfony/Bundle/Resources/config/swagger-ui.xml` in https://github.com/api-platform/core/pull/2384

PR #721, opened 2019-01-21
Add a tip to use MakerBundle to create resources

Deuchnord:
When API Platform is installed, an argument for Symfony MakerBundle's `make:entity` command is added to add some default configurations for API resources.
It looks like this tip is not documented, so this PR adds this information to the _Getting Started_ page.

### DIFF
--- a/core/getting-started.md
+++ b/core/getting-started.md
@@ -143,6 +143,12 @@ class Offer
 It is the minimal configuration required to expose `Product` and `Offer` entities as JSON-LD documents through an hypermedia
 web API.
 
+**Tip**: if you create your entities with Symfony's [MakerBundle](https://symfony.com/doc/current/bundles/SymfonyMakerBundle/index.html), you can also create your API resource entities with the `--api-resource` argument:
+
+```sh
+php bin/console make:entity --api-resource
+```
+
 If you are familiar with the Symfony ecosystem, you noticed that entity classes are also mapped with Doctrine ORM annotations
 and validation constraints from [the Symfony Validator Component](http://symfony.com/doc/current/book/validation.html).
 This isn't mandatory. You can use [your preferred persistence](data-providers.md) and [validation](validation.md) systems.
--- a/core/getting-started.md
+++ b/core/getting-started.md
@@ -145,7 +145,7 @@ web API.
 
 **Tip**: if you create your entities with Symfony's [MakerBundle](https://symfony.com/doc/current/bundles/SymfonyMakerBundle/index.html), you can also create your API resource entities with the `--api-resource` argument:
 
-```sh
+```bash
 php bin/console make:entity --api-resource
 ```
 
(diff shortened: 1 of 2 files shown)


core/getting-started.md
```diff
@@ -143,6 +143,12 @@ class Offer
 It is the minimal configuration required to expose `Product` and `Offer` entities as JSON-LD documents through an hypermedia
 web API.
 
+**Tip**: if you create your entities with Symfony's [MakerBundle](https://symfony.com/doc/current/bundles/SymfonyMakerBundle/index.html), you can also create your API resource entities with the `--api-resource` argument:
+
+```sh
```
alanpoulain:
```suggestion
```bash
```

core/getting-started.md
```diff
@@ -143,6 +143,12 @@ class Offer
 It is the minimal configuration required to expose `Product` and `Offer` entities as JSON-LD documents through an hypermedia
 web API.
 
+**Tip**: if you create your entities with Symfony's [MakerBundle](https://symfony.com/doc/current/bundles/SymfonyMakerBundle/index.html), you can also create your API resource entities with the `--api-resource` argument:
```
dunglas: Tip: you can use Symfony MakerBundle to create you entity thanks to the —api-resource option:

core/getting-started.md
```diff
@@ -143,6 +143,12 @@ class Offer
...
+
+```bash
+php bin/console make:entity --api-resource
```
dunglas: docker-compose exec ...

core/getting-started.md
```diff
@@ -143,6 +143,12 @@ class Offer
 It is the minimal configuration required to expose `Product` and `Offer` entities as JSON-LD documents through an hypermedia
 web API.
 
+**Tip**: you can also use Symfony [MakerBundle](https://symfony.com/doc/current/bundles/SymfonyMakerBundle/index.html) thanks to the `--api-resource` option:
+
+```bash
+docker-compose exec php bin/console make:entity --api-resource
```
antograssiot:
```suggestion
$ docker-compose exec php bin/console make:entity --api-resource
```
maybe ?

alanpoulain: Personally, I don't really like it: it is easier to copy / paste without it.

Deuchnord: I agree, but since the dollar sign is present [here](https://api-platform.com/docs/schema-generator/getting-started/#model-scaffolding), shouldn't we put it here too, for the sake of the coherence?

alanpoulain: Some part of the doc doesn't have it.

antograssiot: forget about it then, that was a question, it's already a nice improvement

dunglas:
Can you move this note in the general getting started guide instead?

dunglas:
Thanks @Deuchnord

PR #722, opened 2019-01-23
Update swagger.md - Problem with decoration

BatsaxIV:
Hi guys,

Following the documentation, my decoration of the `api_platform.swagger.normalizer.documentation` service was not working and I figure out that there was another service already decorating this normalizer with a higher priority: `api_platform.swagger.normalizer.api_gateway` (see https://github.com/api-platform/core/blob/5c8820d2ccb1ffb584a668469918477dda6c2d90/src/Bridge/Symfony/Bundle/Resources/config/swagger.xml#L37).

Because decorating `api_platform.swagger.normalizer.api_gateway` does not _sound_ logical to modify the documentation, another solution would be to keep decorating `api_platform.swagger.normalizer.documentation` and add a tag with a higher priority (but loosing the api_gateway feature).

So maybe, there is a much important work on the core regarding this point, I let you the judge of that 👨‍⚖️ 

Jérémy.

### DIFF
--- a/core/swagger.md
+++ b/core/swagger.md
@@ -35,7 +35,7 @@ $ docker-compose exec php bin/console api:openapi:export --output=swagger_docs.j
 ## Overriding the OpenAPI Specification
 
 Symfony allows to [decorate services](https://symfony.com/doc/current/service_container/service_decoration.html), here we
-need to decorate `api_platform.swagger.normalizer.documentation`.
+need to decorate `api_platform.swagger.normalizer.api_gateway` (which is already decorating the initial service `api_platform.swagger.normalizer.documentation`).
 
 In the following example, we will see how to override the title of the Swagger documentation and add a custom filter for
 the `GET` operation of `/foos` path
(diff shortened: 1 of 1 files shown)


dunglas:
Adding a tag would be better!

BatsaxIV:
And adding information on the dedicated paragraph for Amazon API ? (https://api-platform.com/docs/core/swagger/#compatibilily-layer-with-amazon-api-gateway)

Something like:

> if you have decorating the `api_platform.swagger.normalizer.documentation`and want to use amazon api, please change...

dunglas:
Actually, we should fix the priority of the API Gateway normalizer in core directly. It should be -1 or something like that (ping @vincentchalamon)

BatsaxIV:
Seems a good idea :)

vincentchalamon:
@dunglas this would require some changes in the ApiGatewayNormalizer (specially about the `supportsNormalization` method), but :+1:. I'll try to handle it this week

dunglas:
Closing as fixed in code.

PR #723, opened 2019-01-24
Add missing allow_plain_identifiers option to conf reference

Steveb-p:
Adds `allow_plain_identifiers` option to configuration reference.


dunglas:
Thanks @Steveb-p

PR #724, opened 2019-01-25
Clean serialization incomplete informations from getting started section

jdeniau:
The serialization part of the "getting started" section was partial and unuseful. I think it has been moved some time ago but every time I need to find the (de)normalization context, I'm confused by the getting started page !

I replaced it by a link to the serialization page moved the useful links on argument passed to `serialize` / `deserialize` function in the serialization page.

Thanks


dunglas:
Good one, thanks!

PR #727, opened 2019-02-07
Update elasticsearch.md

ismail1432:
Add services file location


dunglas:
Why? It looks like a regression to me.

ismail1432:
Should be closed I make a mistake :+1:

PR #728, opened 2019-02-08
Invalid tag: normalizer should be denormalizer

smoelker:
The document says to tag the service with  `api_platform.identifier.normalizer` but this should be `api_platform.identifier.denormalizer` in order to work correctly.


dunglas:
Thanks!

PR #738, opened 2019-02-20
Doc cursor based pagination

soyuka:


core/pagination.md
```diff
@@ -320,6 +320,38 @@ class Book
 }
 ```
 
+## Cursor based pagination
+
+To configure your resource to use the cursor-based pagination, select your unique sorted field as well as the direction you’ll like the pagination to go via filters and enable the `pagination_via_cursor` option.
```
antograssiot: Not sure how but I would insist a bit on the fact the order and range filters declaration are mandatory for this to work properly (at least for now)

soyuka: good remark fixed thanks

PR #741, opened 2019-02-20
Add example for the Custom ElasticSearch Filter

ismail1432:
I don't know if this example is very good but it works like a charm in my case and I use it. 
Maybe @meyerbaptiste can have a look :+1:


dunglas:
Thanks @ismail1432

PR #743, opened 2019-02-21
$data variable renamed to $object

victorKochkarev:
The update is made in order to keep documentation more consistent with DataTransformerInterface interface.

### DIFF
--- a/core/dto.md
+++ b/core/dto.md
@@ -66,19 +66,19 @@ final class BookInputDataTransformer implements DataTransformerInterface
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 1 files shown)


dunglas:
I would prefer to change it the other way around (calling the parameter `$data` is mandatory in controllers, because it's the name of the Request attribute, so let's use `$data` everywhere).

teohhanhui:
I think assigning it to a properly named variable would help readability tremendously.

victorKochkarev:
Thank you teohhanhui, I will follow this advice.

PR #747, opened 2019-02-28
Fix blackfire installation documentation

gries:
The way the installtion is setup permissions for */tmp* would be changed.
Making a subdirectory and performing installation there is also the recommended way on the official page: 
https://blackfire.io/docs/integrations/docker

Fun Fact: Installing blackfire the old way breaks some api-platform installations as big requests can't be buffered in /tmp anymore.


dunglas:
Thanks @gries

PR #749, opened 2019-03-02
Fix links

milosa:
Fixed link to GraphQL mutations documentation and added https to the query documentation link.


dunglas:
Thanks!

PR #754, opened 2019-03-07
Custom filter on specific properties using ApiFilter annotation

vvatelot:
When using `ApiFilter` annotation, it seems that you have to set the concerned properties in the annotation itself.

### DIFF
--- a/core/filters.md
+++ b/core/filters.md
@@ -950,7 +950,7 @@ class Offer
     // ...
 }
 ```
-When using `ApiFilter` annotation, the declared properties in the `services.yaml` will not be taken into account. You have to use the `ApiFilter` way (see the [documentation](#apifilter-annotation))
+When using `ApiFilter` annotation, the declared properties in the `services.yaml` will not be taken into account. You have to use the `ApiFilter` way (see the [documentation](#apifilter-annotation)).
 
 Finally you can use this filter in the URL like `http://example.com/offers?regexp_email=^[FOO]`. This new filter will also
 appear in Swagger and Hydra documentations.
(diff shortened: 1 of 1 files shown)


soyuka:
Isn't this explained in the first section of https://api-platform.com/docs/core/filters/#apifilter-annotation ?

vvatelot:
@soyuka yes you're right!
My point is, when reading the documentation (to create a custom filter), I follow the steps:

1. Creating the filter class
2. Declare the filter in the `services.yaml`
3. Chose between `ApiResource` or `ApiFilter` annotation

In my case, I prefer to use `ApiFilter`, but in this case the properties declared in the arguments of `services.yaml` are not taken into account. So yes, after taking some time to find it out it is now clear that when using `ApiFilter` you have to refer to the https://api-platform.com/docs/core/filters/#apifilter-annotation section.

So, my proposal is maybe not accurate, and it should just mention what I mention here with a referal to the https://api-platform.com/docs/core/filters/#apifilter-annotation section

What do you think?

soyuka:
Yeah totally we could add a reference in the custom filters doc

core/filters.md
```diff
@@ -950,8 +950,9 @@ class Offer
     // ...
 }
 ```
+When using `ApiFilter` annotation, the declared properties in the `services.yaml` will not be taken into account. You have to use the `ApiFilter` way (see the [documentation](#apifilter-annotation))
```
alanpoulain:
```suggestion
When using `ApiFilter` annotation, the declared properties in the `services.yaml` will not be taken into account. You have to use the `ApiFilter` way (see the [documentation](#apifilter-annotation)).
```

soyuka:
Thanks @vvatelot @alanpoulain !

PR #755, opened 2019-03-10
Correct the project directory from api to bookshop-api the example project 

ssmusoke:
This is to make it consistent with the project name created with the first command

### DIFF
--- a/deployment/heroku.md
+++ b/deployment/heroku.md
@@ -15,9 +15,7 @@ If you don't already have one, [create an account on Heroku](https://signup.hero
 toolbelt](https://devcenter.heroku.com/articles/getting-started-with-php#set-up). We're guessing you already
 have a working install of [Composer](http://getcomposer.org). Perfect, we will need it.
 
-Create a new API Platform project as usual:
-
-    composer create-project api-platform/api-platform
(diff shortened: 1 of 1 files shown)


ssmusoke:
While it may be better to link to the install documentation, putting the commands here helps the flow for the example rather than having to go somewhere first to get started

alanpoulain:
I think you misunderstood. We don't really support this command anymore.
And IMO it would be better to have a link, because the install documentation explains the possibilities and how it works.

ssmusoke:
Updated to use the example project name to match the setup instructions

deployment/heroku.md
```diff
@@ -15,9 +15,7 @@ If you don't already have one, [create an account on Heroku](https://signup.hero
 toolbelt](https://devcenter.heroku.com/articles/getting-started-with-php#set-up). We're guessing you already
 have a working install of [Composer](http://getcomposer.org). Perfect, we will need it.
 
-Create a new API Platform project as usual:
-
-    composer create-project api-platform/api-platform
+Create a new [API Platform project](distribution/index.md#using-symfony-flex-and-composer-advanced-users) which will be used in the rest of this example
```
alanpoulain: I would prefer if you changed this documentation to make it use the distribution.

dunglas:
Thanks @ssmusoke!

PR #760, opened 2019-03-13
Change doc to reflect new extensions priorities

soyuka:


core/extensions.md
```diff
@@ -114,13 +114,22 @@ services:
 
     'App\Doctrine\CurrentUserExtension':
         tags:
-            - { name: api_platform.doctrine.orm.query_extension.collection, priority: 9 }
+            - { name: api_platform.doctrine.orm.query_extension.collection }
```
dunglas: This part is optional now. Isn't it? We can keep it but we must tell that it's not need if using autoconfiguration (the default).

PR #763, opened 2019-03-15
remove useless example

juferchaud:
If we keep `method: 'GET'` the result will be an additional unexpected route.

In fact, serialization will work as expected, but a new route will be created (a classic collection route on subresource, wich is not expected, specially when you create a subresource path).

I hope i do not misunderstand the implementation.


soyuka:
thanks @juferchaud !

PR #767, opened 2019-03-21
about GraphQL

defro:
serialization groups and serialization group per operation not supported by GraphQL

### DIFF
--- a/core/serialization.md
+++ b/core/serialization.md
@@ -66,6 +66,8 @@ framework:
 
 ## Using Serialization Groups
 
+**This feature is not support by GraphQL query.**
+
 It is simple to specify what groups to use in the API system:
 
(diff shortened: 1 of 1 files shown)


alanpoulain:
It is? https://api-platform.com/docs/core/graphql/#serialization-groups

defro:
> It is? https://api-platform.com/docs/core/graphql/#serialization-groups

Hello @alanpoulain 
No it's not related to this group. It's about this 2 points:

1. https://api-platform.com/docs/core/serialization/#changing-the-serialization-context-dynamically
2. https://api-platform.com/docs/core/serialization/#changing-the-serialization-context-on-a-per-item-basis

alanpoulain:
All the API Platform documentation doesn't apply to GraphQL but I don't think indicating it everywhere it's the case is a good idea. Maybe we should use an icon?

defro:
Hello @alanpoulain 
Yes I would be a great idea to add icon when a feature is available for REST and/or GraphQL request.

dunglas:
Thanks helping improving our docs @defro. Closing for now

PR #787, opened 2019-04-11
Improve JWT docs

teohhanhui:
Stop recommending the use of FOSUserBundle, among other things.

Also see https://github.com/api-platform/demo/pull/111

### DIFF
--- a/core/jwt.md
+++ b/core/jwt.md
@@ -1,64 +1,142 @@
 # JWT Authentication
 
 > [JSON Web Token (JWT)](https://jwt.io/) is a JSON-based open standard ([RFC 7519](https://tools.ietf.org/html/rfc7519)) for creating access tokens that assert some number of claims. For example, a server could generate a token that has the claim "logged in as admin" and provide that to a client. The client could then use that token to prove that he/she is logged in as admin. The tokens are signed by the server's key, so the server is able to verify that the token is legitimate. The tokens are designed to be compact, URL-safe and usable especially in web browser single sign-on (SSO) context.
-> - [Wikipedia](https://en.wikipedia.org/wiki/JSON_Web_Token)
+>
+> ―[Wikipedia](https://en.wikipedia.org/wiki/JSON_Web_Token)
 
 API Platform allows to easily add a JWT-based authentication to your API using [LexikJWTAuthenticationBundle](https://github.com/lexik/LexikJWTAuthenticationBundle).
-To install this bundle, [just follow its documentation](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Resources/doc/index.md).
 
 ## Installing LexikJWTAuthenticationBundle
 
-`LexikJWTAuthenticationBundle` requires your application to have a properly configured user provider.
-You can either use the [Doctrine user provider](https://symfony.com/doc/current/security/user_provider.html#entity-user-provider) provided
-by Symfony (recommended), [create a custom user provider](https://symfony.com/doc/current/security/user_provider.html#creating-a-custom-user-provider)
-or use [API Platform's FOSUserBundle integration](fosuser-bundle.md).
+We begin by installing the bundle:
 
-Here's a sample configuration using the data provider provided by FOSUserBundle:
+    $ docker-compose exec php composer require jwt-auth
+
+Then we need to generate the public and private keys used for signing JWT tokens. If you're using the [API Platform distribution](../distribution/index.md), you may run this from the project's root directory:
+
+    $ docker-compose exec php sh -c '
+        set -e
+        apk add openssl
+        mkdir -p config/jwt
+        jwt_passhrase=$(grep ''^JWT_PASSPHRASE='' .env | cut -f 2 -d ''='')
+        echo "$jwt_passhrase" | openssl genpkey -out config/jwt/private.pem -pass stdin -aes256 -algorithm rsa -pkeyopt rsa_keygen_bits:4096
+        echo "$jwt_passhrase" | openssl pkey -in config/jwt/private.pem -passin stdin -out config/jwt/public.pem -pubout
+        setfacl -R -m u:www-data:rX -m u:"$(whoami)":rwX config/jwt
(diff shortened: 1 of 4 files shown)


core/jwt.md
```diff
@@ -4,61 +4,100 @@
...
+        users:
+            entity:
+                class: App\Entity\User
+                property: email
 
     firewalls:
-        login:
-            pattern:  ^/login
+        dev:
+            pattern: ^/(_(profiler|wdt)|css|images|js)/
```
GregoireHebert: I suggest to remove the trailing `/` as it would prevent to get to `/_profiler` directly

teohhanhui: The correct path is `/_profiler/`, but yeah, for the redirect.

We should probably change it in symfony/recipes too: https://github.com/symfony/recipes/blob/c466e7a6cfa12b61a8e73537df518b4008a8f1c7/symfony/security-bundle/3.3/config/packages/security.yaml#L7

teohhanhui: Also, if someone has an `Image` resource, then `/images/{id}` bypasses authentication... :trollface:

core/jwt.md
```diff
@@ -4,61 +4,100 @@
...
-            pattern:   ^/
-            provider: fos_userbundle
             stateless: true
             anonymous: true
+            provider: users
             guard:
                 authenticators:
                     - lexik_jwt_authentication.jwt_token_authenticator
+```
+
```
antograssiot: Maybe it would help to specify here that the user needs to add a route `config/routes.yaml` then for this to work properly:
```yaml
# api/routes.yaml
authentication_token:
    path: /authentication_token
    methods: [POST]

File: core/jwt.md [hunk 3]
```diff
@@ -4,61 +4,100 @@
...
+If you want to avoid loading the `User` entity from database each time a JWT token needs to be authenticated, you may consider using
+the [database-less user provider](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Resources/doc/8-jwt-user-provider.md) provided by LexikJWTAuthenticationBundle. However, it means you will have to fetch the `User` entity from the database yourself (probably through the Doctrine EntityManager) as needed.
```
alanpoulain:
```suggestion
the [database-less user provider](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Resources/doc/8-jwt-user-provider.md) provided by LexikJWTAuthenticationBundle. However, it means you will have to fetch the `User` entity from the database yourself (probably through the Doctrine EntityManager).
```

core/jwt.md
```diff
@@ -4,61 +4,100 @@
...
+
+### Adding Authentication to an API Which Uses a Path Prefix
```
alanpoulain:
```suggestion
### Adding Authentication to an API which Uses a Path Prefix
```

core/jwt.md
teohhanhui: I think it's safe to remove "css", "images", "js" part as it should never hit the PHP / Symfony app anyway, unless you're using PHP to dynamically generate those (probably a very bad idea in this modern era).

core/jwt.md
teohhanhui: Why? You don't always need to fetch the `User` from database. I don't want to give the wrong impression of always having to do that.

core/jwt.md
teohhanhui: It's correct as it is.

https://capitalizemytitle.com/

core/jwt.md
teohhanhui: But I'll move "as needed" to before the brackets. It probably reads better that way.

teohhanhui:
Also see https://github.com/api-platform/demo/pull/111

Jibbarth:
:+1: 

I just wondering how users can retrieve their token to use in swaggerUi ?

On my projects, I added a collectionOperation `login` on the User Entity, and I don't have to create a custom controller to declare the route. And I can also use the swaggerUi to get the token. WDYT @teohhanhui ?

```php
<?php

namespace App\Entity;

use Doctrine\ORM\Mapping as ORM;
use Symfony\Component\Security\Core\User\UserInterface;
use Symfony\Component\Serializer\Annotation\Groups;

/**
 * @ApiResource(
 *     collectionOperations={
 *         //...
 *         "login"={
 *              "method"="POST",
 *              "path"="/login",
 *              "denormalization_context"={"groups"={"login"}},
 *              "swagger_context"={"summary"="Connect user"}
 *         }
 *     }
 * )
 */
class User implements UserInterface
{

    // ...

    /**
     * @var string
     * @ORM\Column(type="string", length=255)
     * @Groups({"...", "login"})
     */
    private $email;

    /**
     * @var string
     * @ORM\Column(type="string")
     * @Groups({"...", "login"})
     */
    private $password;
}
```

teohhanhui:
@Jibbarth It should not be on the User resource.

teohhanhui:
@Jibbarth Actually, we already link to https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Resources/doc/index.md#1-obtain-the-token

So it's fine.

core/jwt.md
```diff
@@ -1,64 +1,142 @@
...
 ## Installing LexikJWTAuthenticationBundle
 
-`LexikJWTAuthenticationBundle` requires your application to have a properly configured user provider.
-You can either use the [Doctrine user provider](https://symfony.com/doc/current/security/user_provider.html#entity-user-provider) provided
-by Symfony (recommended), [create a custom user provider](https://symfony.com/doc/current/security/user_provider.html#creating-a-custom-user-provider)
-or use [API Platform's FOSUserBundle integration](fosuser-bundle.md).
+We begin by installing the bundle:
 
-Here's a sample configuration using the data provider provided by FOSUserBundle:
+    $ composer require jwt-auth
```
antograssiot: I think that usually we specify commands using the standard distribution but I'm fine like this...

```suggestion
    $ docker-compose exec php composer require jwt-auth
```

teohhanhui:
Thanks for the reviews, everyone. :tada: :rocket:

PR #791, opened 2019-04-18
Add YAML configuration example to specify the identifier

nicor42:


core/identifiers.md
```diff
@@ -28,6 +28,13 @@ final class Person {
     // ...
 }
 ```
+```yaml
```
antograssiot:
```suggestion

You can also use the YAML configuration format:

```yaml
```

teohhanhui:
I think it's better if we restructure this a bit. Having redundant documentation is confusing...

nicor42:
I checked, and there is indeed `identifier: true` in 2 locations

- https://api-platform.com/docs/core/identifiers/
- https://api-platform.com/docs/core/serialization/#entity-identifier-case

as a newbie on ApiPlatform, my opinion is : 

- is it not clear that there is a default identifier, and which one it is 
- it is not clear that there can be several identifiers, or that we can change the default behavior

and I think that those 2 points should be explained in `/docs/core/identifiers`

The page `/docs/core/serialization`, when talking about identifiers, should refer to information in `/docs/core/identifiers`.

If you agree, I could propose a modification, but I'm not sure I'm familiar enough with the doc.

NB : We could also have this example, to remove the default identifier : (if not set, there shall be 2 identifiers)
```yaml
App\Entity\Person:
    properties:
        id:
            identifier: false
        code:
            identifier: true
```

soyuka:
>   is it not clear that there is a default identifier, and which one it is
>   it is not clear that there can be several identifiers, or that we can change the default behavior
> and I think that those 2 points should be explained in /docs/core/identifiers

:+1:, also you example to change the default behavior could be useful but could also be confusing.

dunglas:
I agree too, can you do the changes in this PR? Thanks!

PR #794, opened 2019-04-19
Mercure: fix an error in an environment variable name

Deuchnord:
The current doc of Mercure's usage with API Platform states that we need to create three variables in the `.env` file in order to enable pushing to the Hub: `MERCURE_PUBLISH_URL`, `MERCURE_SUBSCRIBE_URL` and `MERCURE_JWT`. But the last one is not correct, and must actually be named `MERCURE_JWT_SECRET`.


teohhanhui:
Thanks @jewome62! :tada:

PR #800, opened 2019-04-26
fix: undefined $attributes variable

gnumoksha:


dunglas:
Only parenthesis are missing: `!($attributes = RequestAttributesExtractor::extractAttributes($request))`

gnumoksha:
@dunglas Well, my team and I don't use "assignment in conditions" because we think such a thing is error-prone and increase the required cognitive effort to read the code. In fact, we use a [phpcs rule](https://github.com/slevomat/coding-standard/#slevomatcodingstandardcontrolstructuresassignmentincondition) that blocks such thing.

I think a documentation must be as clear as possible thus I will stay with this proposal.

dunglas:
We use such assignations everywhere in API Platform and in the docs (because in cases like this one it’s more concise and not harder to understand). Please change as we try to make snippets consistents.

PR #803, opened 2019-05-06
Avoid reloading the entire app when authenticating

Fab1en:
The `window.location.replace('/');` line in `admin/src/authProvider.js` reloads the entire app when authentication is done. This is only necessary because the `localStorage.getItem('token')` is called once during Admin component startup and not every time a request is done. By moving this call inside `fetchHydra`, the token refresh is taken into account immediately, without needing a full app refresh.


dunglas:
ping @mauchede

vinceAmstoutz:

PR #805, opened 2019-05-08
Reuse the API entrypoint from .env

comxd:


admin/authentication-support.md
```diff
@@ -64,7 +64,7 @@ import { HydraAdmin, hydraClient, fetchHydra as baseFetchHydra } from '@api-plat
 import authProvider from './authProvider';
 import { Route, Redirect } from 'react-router-dom';
 
-const entrypoint = 'https://demo.api-platform.com'; // Change this by your own entrypoint
+const entrypoint = process.env.REACT_APP_API_ENTRYPOINT;
```
dunglas: Could be add a comment to explain what it is, for people not using the distribution.

PR #813, opened 2019-05-14
Choose visible fields

comxd:


comxd:
Do not merge yet, it seem to be currently not implemented. Sorry.

PR #815, opened 2019-05-15
initializing groups key, renaming class and small tweaks

weaverryan:
o/

A few things that I *think* might help:

1) You can only have one context builder per app (right?) so let's make that more clear by calling it `SerializerContextBuilder` instead of `BookContextBuilder`. The name led me to initially expect you to be able to have one context builder per resource.

2) And some minor fixes...

Cheers!

### DIFF
--- a/core/serialization.md
+++ b/core/serialization.md
@@ -420,11 +420,7 @@ final class SerializerContextBuilder implements SerializerContextBuilderInterfac
         $context = $this->decorated->createFromRequest($request, $normalization, $extractedAttributes);
         $resourceClass = $context['resource_class'] ?? null;
 
-        if (!isset($context['groups'])) {
-            $context['groups'] = [];
-        }
-
-        if ($resourceClass === Book::class && $this->authorizationChecker->isGranted('ROLE_ADMIN') && false === $normalization) {
+        if ($resourceClass === Book::class && isset($context['groups']) && $this->authorizationChecker->isGranted('ROLE_ADMIN') && false === $normalization) {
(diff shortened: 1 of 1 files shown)


antograssiot:
you can definitely have more than one SerializerContextBuilder per app @weaverryan 😄

weaverryan:
@antograssiot oh! How? When you decorate the `api_platform.serializer.context_builder` service, that replaces it. If we decorate it multiple times, doesn't only the last one win? So, only 1 is ever used?

Also, I don't understand why we need to set `autoconfigure: false` on this - I can't find any autoconfigure config for `SerializerContextBuilderInterface`.

antograssiot:
Indeed @weaverryan, `autoconfigure:false` is useless I think.

Then if you decorate multiple time, you'll get through all of them AFAIK. This is an example of what I'm using:
``` 
services:
    # default configuration for services in *this* file
    _defaults:
        autowire: true
        autoconfigure: true

    AppBundle\Serializer\ContextBuilder\EntityContextBuilder:
        decorates: 'api_platform.serializer.context_builder'
        decoration_priority: -1

    AppBundle\Serializer\ContextBuilder\CountryContextBuilder:
        decorates: 'api_platform.serializer.context_builder'

    AppBundle\Serializer\ContextBuilder\SalesOrderContextBuilder:
        decorates: 'api_platform.serializer.context_builder'
    
    # more...

```

Using `decoration_priority` you can even tweak the order they are called if needed...

dunglas:
@antograssiot it would be great to add your example to this docs entry!

PR #816, opened 2019-05-16
Removing check in normalizer to prevent multiple calls

weaverryan:
Hi again!

This could be me misunderstanding things, as usual :). I don't understand the `ALREADY_CALLED` check. In fact, if I'm returning a collection of books, I DO want the `normalize()` method to be called and multiple times and for my logic to be run each time. I tested this locally - by removing this check, I can give some books in my collection the `can_retrieve_book` group, but not others.

I'm sure that this logic was there at least originally for some reason. Is it still needed? If so, why?

Cheers!

### DIFF
--- a/core/serialization.md
+++ b/core/serialization.md
@@ -452,8 +452,8 @@ services:
             - { name: 'serializer.normalizer', priority: 64 }
 ```
 
-The Normalizer class is a bit harder to understand, because it must ensure that it is only called once and that there is no recursion.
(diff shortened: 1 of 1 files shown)


weaverryan:
Ok, NOW I understand the recursion problem - it happens when you `use NormalizerAwareTrait`, which you should. And I see now how using the context (which is attached to each object) solves this. Sorry for the noise!

PR #817, opened 2019-05-17
Adding missing namespace to controller

leberknecht:
Without this i'm getting `Controller "CreateBookPublication" does neither exist as service nor as class`, strictly following the example


dunglas:
It should work with the relative path. We need to investigate further.

PR #818, opened 2019-05-17
Update Xdebug to 2.7.2

comxd:
[2019-05-06] — Xdebug 2.7.2
Fixed Bugs
Fixed bug #1488: Rewrite DBGp 'property_set' to always use eval
Fixed bug #1586: error_reporting()'s return value is incorrect during debugger's 'eval' command
Fixed bug #1615: Turn off Zend OPcache when remote debugger is turned on
Fixed bug #1656: remote_connect_back alters header if multiple values are present
Fixed bug #1662: __debugInfo should not be used for user-defined classes


teohhanhui:
Thanks @comxd! :tada:

PR #821, opened 2019-05-20
Add a comment for the listFields option

comxd:
Currently the doc add an empty `listFields` option, this cause a blank List view without fields.


dunglas:
Thanks @comxd

Issue #814, opened 2019-05-15
Document original data for the access expression language 


soyuka:
Document original data for the access expression language from https://github.com/api-platform/core/pull/2779

PR #823, opened 2019-05-21
Document the `previous_object`added in Expression Language

antograssiot:
Add documentation for api-platform/core#2779 and api-platform/core#2811

fixes https://github.com/api-platform/docs/issues/814

### DIFF
--- a/core/security.md
+++ b/core/security.md
@@ -30,7 +30,8 @@ use Symfony\Component\Validator\Constraints as Assert;
  *         "post"={"access_control"="is_granted('ROLE_ADMIN')"}
  *     },
  *     itemOperations={
- *         "get"={"access_control"="is_granted('ROLE_USER') and object.owner == user"}
+ *         "get"={"access_control"="is_granted('ROLE_USER') and object.owner == user"},
+ *         "put"={"access_control"="is_granted('ROLE_USER') and previous_object.owner == user"},
  *     }
  * )
  * @ORM\Entity
@@ -60,7 +61,7 @@ class Book
      * @ORM\ManyToOne(targetEntity=User::class)
      */
     public $owner;
-    
+
     // ...
 }
 ```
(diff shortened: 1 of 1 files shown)


core/security.md
```diff
@@ -60,14 +61,15 @@ class Book
...
-    
+
     // ...
 }
 ```
 
 This example is only going to allow fetching the book related to the current user. If the user tries to fetch a book which is not
 linked to his account, it will not return the resource. In addition, only admins are able to create books which means
 that a user could not create a book.
+Additionally, in some cases you need to perform security checks on the original data. For example here, only the actual owner should be allowed to edit its book. In those cases, you can use the `previous object` variable that will contain the object that was read from your datasource.
```
dunglas:
```suggestion
Additionally, in some cases you need to perform security checks on the original data. For example here, only the actual owner should be allowed to edit its book. In those cases, you can use the `previous_object` variable that will contain the object that was read from your datasource.
```

core/security.md
```diff
@@ -30,7 +30,8 @@ use Symfony\Component\Validator\Constraints as Assert;
  *         "post"={"access_control"="is_granted('ROLE_ADMIN')"}
  *     },
  *     itemOperations={
- *         "get"={"access_control"="is_granted('ROLE_USER') and object.owner == user"}
+ *         "get"={"access_control"="is_granted('ROLE_USER') and object.owner == user"},
+ *         "put"={"access_control"="is_granted('ROLE_USER') and previous_object.owner == user"}
```
teohhanhui: Missing trailing comma.

core/security.md
```diff
@@ -60,14 +61,15 @@ class Book
...
-    
+
     // ...
 }
 ```
 
 This example is only going to allow fetching the book related to the current user. If the user tries to fetch a book which is not
 linked to his account, it will not return the resource. In addition, only admins are able to create books which means
 that a user could not create a book.
+Additionally, in some cases you need to perform security checks on the original data. For example here, only the actual owner should be allowed to edit its book. In those cases, you can use the `previous_object` variable that will contain the object that was read from your datasource.
```
teohhanhui: Needs an empty blank line before to start a new paragraph. (Markdown syntax)

```suggestion
Additionally, in some cases you need to perform security checks on the original data. For example here, only the actual owner should be allowed to edit their book. In these cases, you can use the `previous_object` variable which contains the object that was read from the data provider.
```

teohhanhui:
Thanks @antograssiot! :tada:

PR #825, opened 2019-05-21
Fixed `no PropertyInfo extractor` problem of docs

gustawdaniel:
Problem is described here

> https://github.com/api-platform/core/issues/2563

It occurs when you not define properties in Entity.

Additionally added `@ApiProperty(identifier=true)` because of error "No identifiers defined for resource of type \"App\\Entity\\Book\""

Also removed services registration because of symfony has now autowiring by default.

Added getter and setter in DTO classes because of it allow to generate models in docs


core/dto.md
```diff
@@ -23,6 +24,12 @@ use App\Dto\BookOutput;
  */
 final class Book
 {
+    /**
+     * At least one property should be identifier
+     * @ApiProperty(identifier=true)
```
teohhanhui: It makes much more sense to use `$isbn` as identifier.

gustawdaniel: I agree.

core/dto.md
```diff
@@ -46,7 +53,12 @@ We have the following `BookInput`:
 namespace App\Dto;
 
 final class BookInput {
-  public $isbn;
+    public $isbn;
+
+    /*
+     * Setter is required to generate documentation of example model
+     */
+    public function setIsbn(string $isbn) { $this->isbn = $isbn; }
```
dunglas: It shouldn't be the case. Just adding the following PHPDoc on the property should be enough:
```
/**
  * @var string The ISBN number
  */
```

Can you test this please?

core/dto.md
```diff
@@ -60,36 +72,46 @@ namespace App\DataTransformer;
 
 use ApiPlatform\Core\DataTransformer\DataTransformerInterface;
 use App\Dto\BookInput;
+use App\Entity\Book;
 
 final class BookInputDataTransformer implements DataTransformerInterface
 {
     /**
```
dunglas: You should just totally remove this docblock in the example. It brings nothing.

core/dto.md
```diff
@@ -60,36 +72,46 @@ namespace App\DataTransformer;
...
+     * @param string $to
+     * @param array $context
+     * @return Book
      */
     public function transform($data, string $to, array $context = [])
     {
-        $book = new Book();
-        $book->isbn = $data->isbn;
-        return $book;
+        if($data instanceof BookInput) {
```
dunglas: What if it's not a BookInput? Also, you should use a guard clause, and fix CS.

core/dto.md
```diff
@@ -60,36 +72,46 @@ namespace App\DataTransformer;
...
+            $book = new Book();
+            $book->isbn = $data->isbn;
+            $book->name = strrev($data->isbn); // there can be called service that calculate lacking fields, `strrev` is used for example
+            return $book;
+        }
     }
 
     /**
-     * {@inheritdoc}
+     * BookInput -> Book
```
dunglas: This dockblock brings nothing, and can be confusing, please remove it

core/dto.md
```diff
@@ -128,26 +155,39 @@ use App\Entity\Book;
 final class BookOutputDataTransformer implements DataTransformerInterface
 {
     /**
-     * {@inheritdoc}
+     * Convert Book into BookOutput
```
dunglas: To remove

core/dto.md
gustawdaniel: Yes. I tested. It works exactly like yo wrote.

core/dto.md
gustawdaniel: What is CS? I can simply remove this `if`

teohhanhui: CS = coding standards / coding style. There are multiple CS issues in your changes. Here, for example, it must be:

```php
if (...) {
```

not:

```php
if(...) {
```

smxsm:
Why isn't this merged? The official docs are still wrong and the code there is still throwing this error ...

.gitignore
```diff
@@ -0,0 +1 @@
+.idea
```
alanpoulain: Can you remove this file?

alanpoulain:
It should be rebased and target 2.5 now.

core/dto.md
```diff
@@ -23,6 +24,12 @@ use App\Dto\BookOutput;
  */
 final class Book
 {
+    public $name;
```
alanpoulain: New line?

core/dto.md
```diff
@@ -46,7 +53,10 @@ We have the following `BookInput`:
 namespace App\Dto;
 
 final class BookInput {
```
alanpoulain: New line?

core/dto.md
alanpoulain: Has not been done.

gustawdaniel:
I only wanted improve documentation. 

Key point of this pull request was prevent from error `no PropertyInfo extractor`, no spaces after `if`, new lines and `.idea` in `.gitignore`.

I give up and closing this pull request.

PR #830, opened 2019-06-02
[GraphQL] Custom types, queries and mutations

alanpoulain:


core/graphql.md
```diff
@@ -106,7 +106,168 @@ To get the next page, you would add the `endCursor` from the current page as the
...
+{
+    // ...
+}
+```
+
+As you can see, it's possible to define your own arguments for your custom queries.
+They are following the GraphQL type system.
+If you don't define the `args` property, it will be the default ones (for example `id` for an item).
+
+You can also note that:
```
soyuka: `Note also that:`?

core/graphql.md
```diff
@@ -106,7 +106,168 @@ To get the next page, you would add the `endCursor` from the current page as the
 When the property `hasNextPage` of the `pageInfo` field is false, you've reached the last page.
 If you move forward, you'll end up having an empty result.
 
-## Mutations
+### Custom queries
```
dunglas:
```suggestion
### Custom Queries
```

core/graphql.md
```diff
@@ -106,7 +106,168 @@ To get the next page, you would add the `endCursor` from the current page as the
...
+use App\Model\Book;
+
+final class BookCollectionResolver implements QueryCollectionResolverInterface
+{
+    /**
+     * @param Book[] $collection
+     *
+     * @return Book[]
+     */
+    public function __invoke($collection, array $context)
```
dunglas:
```suggestion
    public function __invoke(iterable $collection, array $context): iterable
```

(not sure if it works)

core/graphql.md
```diff
@@ -129,6 +290,103 @@ mutation DeleteBook($id: ID!, $clientMutationId: String!) {
 }
 ```
 
+### Custom mutations
```
dunglas:
```suggestion
### Custom Mutations
```

core/graphql.md
```diff
@@ -129,6 +290,103 @@ mutation DeleteBook($id: ID!, $clientMutationId: String!) {
...
+
+Creating custom mutations is like creating the [custom queries](#custom-queries).
```
dunglas:
```suggestion
Creating custom mutations is like creating [custom queries](#custom-queries).
```

core/graphql.md
```diff
@@ -129,6 +290,103 @@ mutation DeleteBook($id: ID!, $clientMutationId: String!) {
...
+use App\Model\Book;
+
+final class BookMutationResolver implements MutationResolverInterface
+{
+    /**
+     * @param Book|null $item
+     *
+     * @return Book
+     */
+    public function __invoke($item, array $context)
```
dunglas: Same here, `iterable`?

core/graphql.md
```diff
@@ -410,6 +668,178 @@ In this case, the REST endpoint will be able to get the two attributes of the bo
 The GraphQL endpoint will be able to query only the name. It will only be able to create a book with an author.
 When doing this mutation, the author of the created book will not be returned (the name will be instead).
 
+## Custom types
```
dunglas:
```suggestion
## Custom Types
```

core/graphql.md
```diff
@@ -410,6 +668,178 @@ In this case, the REST endpoint will be able to get the two attributes of the bo
...
+    App\Type\Definition\DateTimeType:
+        tags:
+            - { name: api_platform.graphql.type }
+```
+
+Your custom type is now registered and is available in the `TypesContainer`.
+
+To use it please [modify the extracted types](#modify-the-extracted-types) or use it directly in [custom queries](#custom-queries) or [custom mutations](#custom-mutations).
+
+## Modify the extracted types
```
dunglas:
```suggestion
## Modifying the Extracted Types
```

core/graphql.md
alanpoulain: Maybe it could be but we should change the code of the interface instead (could be done later?).

core/graphql.md
alanpoulain: For the context? I think having an array is enough.

milosa:
I have been trying to add a custom query the way it is described in this PR.
I added this to my Product entity:

```
 * @ApiResource(graphql={
 *     "collectionQuery"={
 *         "collection_query"=ProductCollectionResolver::class
 *     }
 *     })
```

resolver:
```
    /**
     * @param Product[] $collection
     * @param array $context
     * @param ProductRepository $productRepository
     * @return Product[]
     */
    public function __invoke($collection, array $context, ProductRepository $productRepository)
    {
        return $productRepository->findProductsByTagCollection(1);
    }
```

I expected this to just add an extra query, but what happens is that it completely removes all the default queries and mutations (create, update, delete) related to the Product entity and the inputs and payloads for them.

Maybe I'm doing something wrong?

alanpoulain:
Yes you need to add them too, like REST custom actions: https://api-platform.com/docs/core/operations/#recommended-method

PR #837, opened 2019-06-10
Update Traefik according to api-platform V2.4 docker-compose declaration

Darkweak:
Implements  into traefik doc file
* mercure and admin declaration
* going further to define all in .env file

### DIFF
--- a/deployment/traefik.md
+++ b/deployment/traefik.md
@@ -26,7 +26,7 @@ The API DNS will be specified with `traefik.frontend.rule=Host:your.host` (here 
 
 
 ```yaml
-# api-platform-app/docker-compose.yml
+# docker-compose.yml
 version: '3.4'
 
 x-cache:
@@ -181,10 +181,10 @@ If your network is of type B, it may conflict with the Traefik sub-network.
 
 ## Going Further
 
-You may want multiple api-platform instances on same server. To deal with it, you'll have to externalize first traefik to another docker-compose file anywhere on your server, but as traefik will listen on external 80 & 443 ports you can run only 1 traefik instance with this configuration per server  
-Then you'll have something like this example :  
+As this Traefik configuration listens on 80 and 443 ports, you can run only 1 Traefik instance per server. However, you may want to run multiple API Platform projects on same server. To deal with it, you'll have to externalize the Traefik configuration to another `docker-compose.yml` file, anywhere on your server.
+Here is a working example:  
 ```yaml
-# traefik/docker-compose.yml
+# /somewhere/docker-compose.yml
 version: '3.4'
 
 services:
@@ -197,7 +197,7 @@ services:
       - "8080:8080"
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
-# Uncomment these following lines to enable toml files & let's encrypt certificates, refer to traefik doc : https://docs.traefik.io/user-guide/docker-and-lets-encrypt/
+# load a TOML configuration file and to generate Let's Encrypt certificated as explained in https://docs.traefik.io/user-guide/docker-and-lets-encrypt/
 #      - ./traefik.toml:/traefik.toml
 #      - ./acme.json:/acme.json
     networks:
@@ -210,9 +210,9 @@ networks:
   # Add other networks here
 ```
 
-After that, you must define your api-platform docker-compose file  
-```yaml
-# api-platform-app/docker-compose.yml
+Then update the `docker-compose.yaml` file belonging to your API Platform projects:  
+```patch
+# docker-compose.yml
 version: '3.4'
 
 x-cache:
@@ -222,22 +222,22 @@ x-cache:
     - ${CONTAINER_REGISTRY_BASE}/nginx
     - ${CONTAINER_REGISTRY_BASE}/varnish
 
-x-network:
-  &network
-  networks:
-    - api_platform_network
++x-network:
++  &network
++  networks:
++    - api_platform_network
 
 services:
 # Uncomment these lines only if you want to run one api-platform instance using traefik
-#  traefik:
-#    image: traefik:latest
-#    command: --api --docker
-#    ports:
-#      - "80:80"
-#      - "443:443"
-#    volumes:
-#      - /var/run/docker.sock:/var/run/docker.sock
-#    <<: *network
+-  traefik:
+-    image: traefik:latest
+-    command: --api --docker
+-    ports:
+-      - "80:80"
+-      - "443:443"
+-    volumes:
+-      - /var/run/docker.sock:/var/run/docker.sock
+-    <<: *network
 
   php:
     image: ${CONTAINER_REGISTRY_BASE}/php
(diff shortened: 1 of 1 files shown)


deployment/traefik.md
```diff
@@ -12,29 +12,38 @@ Override ports and add labels to tell Traefik to listen on the routes mentionned
...
-      - "traefik.frontend.rule=Host:api.localhost"
+      - traefik.frontend.rule=Host:api.localhost
 ``` 
 The API DNS will be specified with `traefik.frontend.rule=Host:your.host` (here api.localhost)  
 
 `--traefik.port=3000` The port specified to Traefik will be exposed by the container (here the React app exposes the 3000 port)  
 
 
 ```yaml
+# api-platform-app/docker-compose.yml
```
dunglas:
```suggestion
# docker-compose.yml
```

deployment/traefik.md
```diff
@@ -120,3 +178,195 @@ If you do that, you'll have to update the `CORS_ALLOW_ORIGIN` environment variab
 ## Known Issues
 
 If your network is of type B, it may conflict with the Traefik sub-network.
+
+## Going Further
+
+You may want multiple api-platform instances on same server. To deal with it, you'll have to externalize first traefik to another docker-compose file anywhere on your server, but as traefik will listen on external 80 & 443 ports you can run only 1 traefik instance with this configuration per server
```
dunglas:
```suggestion
As this Traefik configuration listens on 80 and 443 ports, you can run only 1 Traefik instance per server. However, you may want to run multiple API Platform projects on same server. To deal with it, you'll have to externalize the Traefik configuration to another `docker-compose.yml` file, anywhere on your server.
```

deployment/traefik.md
```diff
@@ -120,3 +178,195 @@ If you do that, you'll have to update the `CORS_ALLOW_ORIGIN` environment variab
 ## Known Issues
 
 If your network is of type B, it may conflict with the Traefik sub-network.
+
+## Going Further
+
+You may want multiple api-platform instances on same server. To deal with it, you'll have to externalize first traefik to another docker-compose file anywhere on your server, but as traefik will listen on external 80 & 443 ports you can run only 1 traefik instance with this configuration per server  
+Then you'll have something like this example :
```
dunglas:
```suggestion
Here is a working example:  
```

deployment/traefik.md
```diff
@@ -120,3 +178,195 @@ If you do that, you'll have to update the `CORS_ALLOW_ORIGIN` environment variab
 ## Known Issues
 
 If your network is of type B, it may conflict with the Traefik sub-network.
+
+## Going Further
+
+You may want multiple api-platform instances on same server. To deal with it, you'll have to externalize first traefik to another docker-compose file anywhere on your server, but as traefik will listen on external 80 & 443 ports you can run only 1 traefik instance with this configuration per server  
+Then you'll have something like this example :  
+```yaml
+# traefik/docker-compose.yml
```
dunglas:
```suggestion
# /somewhere/docker-compose.yml
```

deployment/traefik.md
```diff
@@ -120,3 +178,195 @@ If you do that, you'll have to update the `CORS_ALLOW_ORIGIN` environment variab
...
+  traefik:
+    image: traefik
+    command: --api --docker
+    ports:
+      - "80:80"
+      - "443:443"
+      - "8080:8080"
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+# Uncomment these following lines to enable toml files & let's encrypt certificates, refer to traefik doc : https://docs.traefik.io/user-guide/docker-and-lets-encrypt/
```
dunglas:
```suggestion
# Uncomment these following lines to load a TOML configuration file and to generate Let's Encrypt certificated as explained in https://docs.traefik.io/user-guide/docker-and-lets-encrypt/
```

deployment/traefik.md
```diff
@@ -120,3 +178,195 @@ If you do that, you'll have to update the `CORS_ALLOW_ORIGIN` environment variab
...
+      - api_platform_network
+    # Add other networks here
+
+networks:
+  api_platform_network:
+    external: true
+  # Add other networks here
+```
+
+After that, you must define your api-platform docker-compose file
```
dunglas:
```suggestion
Then update the `docker-compose.yaml` file belonging to your API Platform projects:

```

deployment/traefik.md
```diff
@@ -120,3 +178,195 @@ If you do that, you'll have to update the `CORS_ALLOW_ORIGIN` environment variab
...
+
+networks:
+  api_platform_network:
+    external: true
+  # Add other networks here
+```
+
+After that, you must define your api-platform docker-compose file  
+```yaml
+# api-platform-app/docker-compose.yml
```
dunglas:
```suggestion
# docker-compose.yml
```

deployment/traefik.md
```diff
@@ -120,3 +178,195 @@ If you do that, you'll have to update the `CORS_ALLOW_ORIGIN` environment variab
...
+version: '3.4'
```
dunglas: I suggest to use the patch format instead:

```
```patch
# docker-compose.yml
-  traefik:
-    image: traefik:latest
-    command: --api --docker
-    ports:
-      - "80:80"
-      - "443:443"
-    volumes:
-      - /var/run/docker.sock:/var/run/docker.sock
-    <<: *network 
```
```

File: deployment/traefik.md [hunk 9]
```diff
@@ -120,3 +178,195 @@ If you do that, you'll have to update the `CORS_ALLOW_ORIGIN` environment variab
...
+
+volumes:
+  db-data: {}
+
+networks:
+  api_platform_network:
+    external: true
+```
+
+Then as you can see, there is some env vars must be defined, here is an example of .env file :
```
dunglas:
```suggestion
Finally, some environment variables must be defined, here is an example of a `.env` file to set them:  
```

deployment/traefik.md
```diff
@@ -120,3 +178,195 @@ If you do that, you'll have to update the `CORS_ALLOW_ORIGIN` environment variab
...
+DOMAIN_NAME=localhost
+HTTP_OR_SSL=http://
+DB_NAME=api-platform-db-name
+DB_PASS=YouMustChangeThisPassword
+DB_USER=api-platform
+JWT_KEY=!UnsecureChangeMe!
+SUBDOMAINS_LIST=(admin|api|cache|mercure|www)
+```
+
+This way let you configure your main variables into one single file
```
dunglas:
```suggestion
This way, you can configure your main variables into one single file.
```

Darkweak:
ping @dunglas Fixes are done

teohhanhui:
Is there a reason why we don't just use Traefik in api-platform/api-platform distribution? Or does it add much complexity?

Darkweak:
@dunglas told me it wasn't necessary to including this on distribution, so it's just in doc part

teohhanhui:
I think it'd be desirable to include Traefik out of the box in the distribution. WDYT @dunglas?

But I'm not sure how feasible it is to have the hostnames be resolved correctly, out of the box.

teohhanhui:
If only Docker supports zeroconf: https://en.wikipedia.org/wiki/Zero-configuration_networking 😂

teohhanhui:
https://github.com/containous/traefik/issues/4092

dunglas:
Traefik introduces a one more dependency, and some complexity. Also, from my experience, it doesn't play very well with Kubernetes (it's supported, but it's a bit hacky, especially with Helm). I would prefer to keep Traefik setup as a docs entry for now

Darkweak:
So, just merge this, at the moment we won't implement it into API P

dunglas:
Thanks @Darkweak

PR #838, opened 2019-06-20
Reduce visibility of FOSUserBundle

teohhanhui:
Don't promote it


dunglas:
I would prefer to add a warning on the FOSUser integration page than removing this link.

teohhanhui:
I think it's bad to highlight it as one of the "features" of API Platform.

teohhanhui:
Maybe we can move it further down the page?

teohhanhui:
For example, we don't mention NelmioApiDocBundle integration as a feature. We should give FOSUserBundle the same treatment.

dunglas:
Thanks @teohhanhui. It will still appear in the menu so it's good.

PR #839, opened 2019-06-22
Update import of `history` package

glauberm:
Update import of `history` package to avoid the following error:
```
Warning: Please use `require("history").createBrowserHistory` instead of `require("history/createBrowserHistory")`. Support for the latter will be removed in the next major release.
```


dunglas:
Thanks @glauberm

PR #840, opened 2019-06-24
Wrong injected service referenced

yivi:
Text reference to injected service ($myService) doesn't match the service injected in the code snippet. (BookPublishingHandler $bookPublishingHandler)


dunglas:
Good catch! Thanks @yivi

PR #841, opened 2019-06-24
Event: GetResponseForControllerResultEvent is deprecated

jb-reynaud:
## Why
Since Symfony 4.3, this class is deprecated.
source: https://github.com/symfony/symfony/blob/4.3/src/Symfony/Component/HttpKernel/Event/GetResponseForControllerResultEvent.php#L18


dunglas:
Thanks @jb-reynaud

PR #843, opened 2019-07-01
Update troubleshooting.md

elodieirdor:
Add a little section about a trouble I ran to. Maybe it can help !

### DIFF
--- a/extra/troubleshooting.md
+++ b/extra/troubleshooting.md
@@ -95,3 +95,9 @@ server {
 ```
 
 You then need to rebuild your containers by running `docker-compose build`.
+
+## "connect() failed (111: Connection refused) while connecting to upstream" 502 error
+
+Some of your API calls fail with a 502 error and the logs for the api container shows the following error message `connect() failed (111: Connection refused) while connecting to upstream`.
+
+This can be due to the api console which is not executable. Just run `chmod 755 api/bin/console`.


dunglas:
It's probably because you downloaded the `.zip`archive instead of the `.tar.gz`. Some Zip unarchiver doesn't preserve properly the permissions during the extraction process, it's why we suggest to always prefer the zip.

Maybe could you also mention this in the section?

dunglas:
And thanks for contributing btw!

teohhanhui:
While the solution helped in your case, that error message could mean anything. So I'm not really sure we should present it this way...

elodieirdor:
It was a new installation from the github template.

dunglas:
Hum that's weird. How did you retrieved the template's code? Using simply `git clone`? What is your operating system?

teohhanhui:
Yes, the GitHub template feature does not preserve the executable bit, unfortunately. Strangely, a normal fork does.

teohhanhui:
https://github.com/api-platform/api-platform/issues/1164#issuecomment-502753250

elodieirdor:
Indeed @teohhanhui that was my problem!

elodieirdor:
Should I close this PR or do you want me to update the section and mention that the problem occured only with the github template?

PR #845, opened 2019-07-02
Update authentication-support.md to fix bug in getParameters

IkeLutra:
`getParameters` on each resource is called in `componentDidMount` of `HydraAdmin`.

The default method doesn't send the `Authorization` header. Here is an adjustment to the sample code to do this.

Let me know if there is a better way to solve this

### DIFF
--- a/admin/authentication-support.md
+++ b/admin/authentication-support.md
@@ -75,7 +75,17 @@ const apiDocumentationParser = entrypoint =>
   parseHydraDocumentation(entrypoint, {
     headers: new Headers(fetchHeaders),
   }).then(
-    ({ api }) => ({ api }),
+    ({ api }) => {
+      api.resources = api.resources.map(resource => {
+        resource.getParameters = () => {
+          return getParameters(resource, {
+            headers: new Headers(fetchHeaders)
+          });
+        };
+        return resource;
+      });
+      return { api }
+    },
     result => {
       const { api, status } = result;
 


Vultures:
@IkeLutra The code worked fine for me. Thanks.
Please add the following import since it is throwing an error without it:
`import getParameters from '@api-platform/api-doc-parser/lib/hydra/getParameters';`

admin/authentication-support.md
```diff
@@ -75,7 +75,17 @@ const apiDocumentationParser = entrypoint =>
   parseHydraDocumentation(entrypoint, {
     headers: new Headers(fetchHeaders),
   }).then(
-    ({ api }) => ({ api }),
+    ({ api }) => {
```
Darkweak: At the top of the document import such as @Vultures said
```
import getParameters from '@api-platform/api-doc-parser/lib/hydra/getParameters';
```

then declare common headers just after fetchHeaders declaration
```
const fetchHeaders = {'Authorization': `Bearer ${localStorage.getItem('token')}`};
const headers = new Headers(fetchHeaders);
```

After that, you'll be able to reduce your fix like that
```suggestion
      ({ api }) => {
      api.resources = api.resources.map(resource => {
        resource.getParameters = () => getParameters(resource, { headers });
        return resource;
      });
      return { api }
    }
```

You did a good work then 👍

PR #850, opened 2019-07-22
fix enum type without properties

yvalentin:
To corresponding to the example given on the related PHP class part, the ```enum``` type ```OfferItemCondition``` needs to have:
```yaml
  properties: {}
```


dunglas:
Thanks @yvalentin

PR #859, opened 2019-08-19
Fix #1646

vincentchalamon:


core/security.md
```diff
@@ -114,17 +114,62 @@ Alternatively, using YAML:
...
+            security: 'is_granted("ROLE_USER") and object.owner == user'
+            security_message: 'Sorry, but you are not the book owner.'
+    # ...
+```
+
+## Execute security after denormalization
+
+The "security" attribute is executed before the object denormalization. For some cases, it might be useful to execute
+a security after the denormalization.
+To do so, prefer using "late\_security", which allows you to use the "previous\_object" as the denormalized object:
```
antograssiot: That seems, strange, I would expect that in a late security, `object` would be the denormalized object and `previous_object` the one before denormalization.

vincentchalamon: You're right, I'll fix that. BTW, I forgot to rename the `late\_security` to `security\_post\_denormalize`

PR #860, opened 2019-08-22
Don't suggest symfony/web-server-bundle

chalasr:


dunglas:
Thanks (for the record, we removed this line because WebServerBundle is now deprecated)

PR #862, opened 2019-08-25
Important "preface" for GraphQL about property-types

oleg-andreyev:


core/graphql.md
```diff
@@ -10,6 +10,12 @@ API Platform creates a REST API by default. But you can choose to enable GraphQL
 
 Once enabled, you have nothing to do: your schema describing your API is automatically built and your GraphQL endpoint is ready to go!
 
+## Preface
+Because GraphQL is strongly typed it is important to define type of your properties and insure that they are readable.
```
alanpoulain: I don't think it's necessary to talk about the `SchemaBuilder` class directly. Is there a particular aim?

core/graphql.md
```diff
@@ -129,6 +135,19 @@ mutation DeleteBook($id: ID!, $clientMutationId: String!) {
 }
 ```
 
+if you publish a book
```
alanpoulain: Why adding this?

core/graphql.md
```diff
@@ -169,6 +188,20 @@ class Offer
 }
 ```
 
+```yaml
```
alanpoulain: Please add a sentence before.

core/graphql.md
oleg-andreyev: It can be removed. I just wanted to point other developers where to look if they need to debug something. I’ve spent few hours understanding why my fields aren’t returned and after debugging SchemaBuilder I’ve figured out that property must be typed

alanpoulain: In master, the `SchemaBuilder` is no more the only class to debug so I'm not sure we should add it here.

oleg-andreyev: Hmm. I can change target branch to 2.4.x if it still accepts PRs, otherwise I'll review master and rephrase it.

alanpoulain: It accepts PRs, but the 2.5 will be released soon so it would be probably better to target master.

alanpoulain:
Closing this for now. Feel free to reopen it.

PR #876, opened 2019-09-23
Use retrieve documentation

dunglas:


alanpoulain:
Is there a difference? `bin/check-outline` only displays some warnings.

dunglas:
That's the goal! Warnings must be displayed (and fixed).

PR #877, opened 2019-09-24
Improve code to be consistent with arraycollection

ismail1432:


dunglas:
Hi! Thanks for the contribution, but can you explain a bit more what's the benefit? AFAIK, `ArrayCollection` implements the `\ArrayAccess` interface.

ismail1432:
Thanks for your great work !
 There is no big benefit except that most of the time when you see an example of an addX you see it with the add method instead the []

alanpoulain:
I think it's fine as is. Closing.

PR #886, opened 2019-09-26
Document disabling item operation

soyuka:


core/operations.md
```diff
@@ -134,6 +134,31 @@ Or the XML configuration format:
 API Platform Core is smart enough to automatically register the applicable Symfony route referencing a built-in CRUD action
 just by specifying the method name as key, or by checking the explicitly configured HTTP method.
 
+If you want to completely remove item operations, you have to declare a default GET operation so that the resource can still be identified by an IRI. For example:
```
teohhanhui:
```suggestion
If you do not want to allow access to the resource item (i.e. you don't want a `GET` item operation), instead of omitting it altogether, you should instead declare a `GET` item operation which returns HTTP 404 (Not Found), so that the resource item can still be identified by an IRI. For example:
```

core/operations.md
```diff
@@ -134,6 +134,31 @@ Or the XML configuration format:
...
+ *     itemOperations={
+ *         "get"={
+ *             "controller"=NotFoundAction::class,
+ *             "read"=false,
+ *             "output"=false,
+ *         },
+ *     },
+ * )
+ */
+ class Book {}
```
teohhanhui:
```suggestion
 class Book {
 }
```

core/operations.md
```diff
@@ -134,6 +134,31 @@ Or the XML configuration format:
 API Platform Core is smart enough to automatically register the applicable Symfony route referencing a built-in CRUD action
 just by specifying the method name as key, or by checking the explicitly configured HTTP method.
 
+If you want to completely remove item operations, you have to declare a default GET operation so that the resource can still be identified by an IRI. For example:
+
+```
+<?php
+
```
teohhanhui: Missing namespace...

PR #887, opened 2019-09-27
Use yaml const to not create a deprecation message

darthf1:
`Using a string "HTTP_BAD_REQUEST" as a constant of the "Symfony\Component\HttpFoundation\Response" class is deprecated since API Platform 2.1 and will not be possible anymore in API Platform 3. Use the Symfony's custom YAML extension for PHP constants instead (i.e. "!php/const Symfony\Component\HttpFoundation\Response::HTTP_BAD_REQUEST").`


soyuka:
thanks @darthf1

PR #888, opened 2019-09-27
Add extending docs, improve security docs and getting started

dunglas:
Partial alternative to #885. Closes https://github.com/api-platform/docs/pull/866.

PR #889, opened 2019-09-27
Review messenger related documentation

soyuka:
Related: https://github.com/api-platform/core/issues/3065
 @asimonf do you see something we should clarify a bit more or that's missing here? Thanks!


core/messenger.md
```diff
@@ -29,22 +29,21 @@ use Symfony\Component\Validator\Constraints as Assert;
...
-final class ResetPasswordRequest
+final class Book
 {
     /**
      * @var string
      *
      * @Assert\NotBlank
      */
-    public $username;
+    public title;
```
ArnoudThibaut:
```suggestion
    public $title;
```

asimonf:
> Related: [api-platform/core#3065](https://github.com/api-platform/core/issues/3065)
> @asimonf do you see something we should clarify a bit more or that's missing here? Thanks!

Thanks for asking. 

If I'm not mistaken, given that the example is now a Book resource, it is entirely possible that someone uses that example as a base with a Doctrine entity that has relations with other entities. In such cases, if the native serializer is used for the message (`Symfony\Component\Messenger\Transport\Serialization\PhpSerializer`), those fields won't be properly merged to the object manager on the other side of the transport. Bare in mind that this serializer is the default since Symfony 4.3. 

I think what's missing is either a mention that the above is expected behavior and the user should manually merge those fields back to the object manager in the message handler, or, make a mention that tells the user to change the native serializer to the default one at the cost of performance to avoid having to merge those entities and handle special cases. The latter would be possible using the following inside the `messenger.yaml` file.

```yaml
framework:
    messenger:
        serializer:
            default_serializer: messenger.transport.symfony_serializer
            symfony_serializer:
                format: json
                context: { }
```

I don't know if the above will work just fine when using DTOs, though, so maybe just a mention, for now, that entities should be manually merged back?

Maybe a link to the documentation on Symfony's Messenger page would be appropriate. Also, maybe a mention that while functionally stable, this is still experimental and the API might change in future releases.

soyuka:
Thank you very much for your input!

> the user should manually merge those fields back to the object manager in the message handler, or, make a mention that tells the user to change the native serializer to the default one at the cost of performance to avoid having to merge those entities and handle special cases

Indeed, I'll definitely propose both solutions IMO both are valid. I'll also add a note that DTO will not work with API Platform's serializer because our serializer is working only with Resources tagged as `ApiResource`, they need to have an IRI (ie at least one GET operation in the Router).

@teohhanhui don't hesitate to comment if you think we should add something else!

core/messenger.md
```diff
@@ -98,20 +95,39 @@ That's all!
...
+framework:
+    messenger:
+        serializer:
+            default_serializer: messenger.transport.symfony_serializer
+            symfony_serializer:
+                format: json
+                context: { }
+```
+
+Note that when doing so it'll have a negative impact on performances. Also, if API Platform's serializer is used, it will work only for resources tagged as `ApiResource` and not for DTOs. Indeed, our normalizer expects resources to have an IRI, in other words at least one `GET` operation declared in symfony's `RouteCollection` (see also [](./operations.md)).
```
teohhanhui: > Also, if API Platform's serializer is used, it will work only for resources tagged as `ApiResource` and not for DTOs.

I'm not sure I understand what you mean here.

soyuka: I'm trying to say that our normalizer doesn't support DTOs because they're not resources.

asimonf:
@teohhanhui @soyuka 

I have a question regarding the latter part of the doc. What happens if, instead of implementing a `POST`, a `PUT` is implemented using a DTO. This would mean that the user is requesting an edit rather than creating a new resource. Is a data transformer called? If it is, when? Would there be a way to have access to the original object to perform the actual transformation? It isn't obvious to me, as an outsider, where would the Data Transformer enter into this, if it enters at all.

soyuka:
Yes it will, and you'll have access to the Resource that needs to be updated see https://api-platform.com/docs/core/dto/#updating-a-resource-with-a-custom-input. 
The call to messenger will happen later when the resource should've been written (WriteListener, better docs about extension points soon https://github.com/api-platform/docs/pull/888).

Now the above is true with `messenger=true` and an `input` class. 

If you set `messenger=input` the Input will not be transformed because it'll use [this data transformer](https://github.com/api-platform/core/blob/5503d65e8ad6535539ac4f6b1bcc89962457bc98/src/Bridge/Symfony/Messenger/DataTransformer.php) that just returns the input which will then be send to messenger. 

Related issue that'll improve resources update with messenger: https://github.com/api-platform/core/issues/3082#issuecomment-535575673 (UpdateStamp implementation).

core/messenger.md
```diff
@@ -29,22 +29,21 @@ use Symfony\Component\Validator\Constraints as Assert;
 
 /**
  * @ApiResource(
- *     messenger=true,
  *     collectionOperations={
+ *         "output"=false,
```
dunglas: It will not work. Should be either as an attribute of `post` (preferred) or at the root of the annotation.

core/messenger.md
```diff
@@ -29,22 +29,21 @@ use Symfony\Component\Validator\Constraints as Assert;
...
  *         "post"={"status"=202}
  *     },
- *     itemOperations={},
- *     output=false
+ *     messenger=true,
```
dunglas: As only the `POST` operation is registered, wouldn't it be more explicit to set this attribute at the operation level too?

core/messenger.md
```diff
@@ -53,40 +52,38 @@ Alternatively, you can use the YAML configuration format:
...
     attributes:
         messenger: true
+    collectionOperations:
+      post:
         output: false
+        status: 202
 ```
 
-Because the `messenger` attribute is `true`, when a `POST` is handled by API Platform, the corresponding instance of the `ResetPasswordRequest` will be dispatched.
+Because the `messenger` attribute is `true`, when a write operation (`PUT`, `DELETE`, `POST`) is handled by API Platform, the corresponding instance of the `Book` will be dispatched.
```
dunglas:
```suggestion
Because the `messenger` attribute is `true`, when a write operation (`PUT`, `DELETE`, `POST`) is handled by API Platform, after [deserialization](serialization.md) the hydrated instance of the `Book` object will be dispatched to [the default message bus](https://symfony.com/doc/current/messenger.html#dispatching-the-message).
```

core/messenger.md
```diff
@@ -53,40 +52,38 @@ Alternatively, you can use the YAML configuration format:
...
 
-For this example, only the `POST` operation is enabled.
-We use the `status` attribute to configure API Platform to return a [202 Accepted HTTP status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202).
+For the `POST` operation here, We use the `status` attribute to configure API Platform to return a [202 Accepted HTTP status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202).
```
dunglas:
```suggestion
For the `POST` operation here, we use the `status` attribute to return a [202 Accepted HTTP status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202) in case of success.
```

core/messenger.md
```diff
@@ -98,20 +95,39 @@ That's all!
...
 API Platform automatically uses the `Symfony\Component\Messenger\Stamp\HandledStamp` when set.
 It means that if you use a synchronous handler, the data returned by the `__invoke` method replaces the original data.
 
-## Detecting Removals
+### Detecting Removals
 
 When a `DELETE` operation occurs, API Platform automatically adds a `ApiPlatform\Core\Bridge\Symfony\Messenger\RemoveStamp` ["stamp"](https://symfony.com/doc/current/components/messenger.html#adding-metadata-to-messages-envelopes) instance to the "envelope".
 To differentiate typical persists calls (create and update) and removal calls, check for the presence of this stamp using [a custom "middleware"](https://symfony.com/doc/current/components/messenger.html#adding-metadata-to-messages-envelopes).
 
+### Working with Doctrine's entity manager
```
dunglas:
```suggestion
### Working with Doctrine's Entity Manager
```

core/messenger.md
```diff
@@ -98,20 +95,39 @@ That's all!
...
+
+Since Symfony 4.3, the messenger is using the native serializer (`Symfony\Component\Messenger\Transport\Serialization\PhpSerializer`). If you're using Doctrine entities as the Message, note that relation fields will not be in sync with Doctrine's Object Manager and you will have to merge those by hand in the message handler.
```
dunglas:
```suggestion
When using async handlers, by default the Messenger component uses the native `serialize()` PHP function to serialize the content of the payload to transfer to the handler. If you're using a Doctrine entity as a message, it will not be in sync with the Doctrine's Object Manager when received by the handler. You'll have [to merge the entity](https://www.doctrine-project.org/projects/doctrine-orm/en/current/reference/working-with-objects.html#merging-entities) manually.
```

core/messenger.md
```diff
@@ -98,20 +95,39 @@ That's all!
...
+Another solution is to switch the messenger's serializer to one supporting the format of your choice, for example:
```
dunglas: I don't understand how serializing in JSON instead of using the native format fix prevent to have to merge the entity. I would drop the rest of this section entirely.

core/messenger.md
```diff
@@ -120,66 +136,85 @@ Set the `messenger` attribute to `input`, and API Platform will automatically di
...
  * @ApiResource(
  *     collectionOperations={
- *         "post"={"status"=202}
- *     },
- *     itemOperations={},
- *     messenger="input",
- *     input=ResetPasswordRequest::class,
- *     output=false
+ *         "post",
+ *         "get",
```
dunglas: should be first

core/messenger.md
```diff
@@ -120,66 +136,85 @@ Set the `messenger` attribute to `input`, and API Platform will automatically di
...
+
+final class PasswordResetRequest
 {
     /**
-     * @var string
-     * @Assert\NotBlank
+     * @Groups({"user_password_reset_request"})
      */
-    public $var;
+    private $email;
```
dunglas: Why not using a public properties? What's the benefit of using serialization groups here? It makes the config super verbose for no benefits.

core/messenger.md
```diff
@@ -120,66 +136,85 @@ Set the `messenger` attribute to `input`, and API Platform will automatically di
...
+    public function getEmail(): string
+    {
+        return $this->email;
+    }
 }
 ```
 
-For this example, only the `POST` operation is enabled on `/users`.
-As above, we use the `status` attribute to configure API Platform to return a [202 Accepted HTTP status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202).
+For the `POST` operation here, We use the `status` attribute to configure API Platform to return a [202 Accepted HTTP status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202).
```
dunglas: duplicated, or keep "as above" maybe?

core/messenger.md
asimonf: It's not about the format per se. It's because, by falling back to the default Symfony serializer, the Api-Platform item normalizer ends up being called and, since this object is a resource, that normalizer does the job of merging the entities back to the Object Manager. 

This was the default behavior before 4.3, which is why I suggested it should be mentioned somewhere.

asimonf: I would leave the section and maybe add an extra caveat since it is entirely possible that a user reads the Symfony documentation and decides to use json for interoperability reasons and then don't realize why the behavior is different.

Steveb-p:
If I can voice my opinion.

I've recently tried using messenger integration in my app and was surprised by the fact that enabling it causes my entity (which I used as input) to stop being saved to database. I'd love it if somewhere in documentation there would be a part warning that this works as a Data Persister and replaces the Doctrine one.

I now know that they are chained and the first supporting persister works, but it baffled me at the time. I couldn't figure out why enabling messenger integration for an entity caused my serializer to fail when trying to generate an IRI for the entity.

vinceAmstoutz:

PR #906, opened 2019-10-03
Fix syntax highlight in example

lex111:
:open_mouth:


alanpoulain:
Good catch, thank you.

PR #907, opened 2019-10-03
Add precisions on the custom permission

rgazelot:
It took me a while to figure out that I had to give the object into my `is_granted()` expression. Maybe it could be added to the documentation to save time for new users like me.


dunglas:
Thanks for the clarification @rgazelot!

PR #910, opened 2019-10-04
Document context aware data persister interface fix #902

### LINKED ISSUE #902: Document the ContextAwareDataPersisterInterface interface
https://github.com/api-platform/core/blob/master/src/DataPersister/ContextAwareDataPersisterInterface.php should be used in https://github.com/api-platform/docs/blob/2.5/core/data-persisters.md

soyuka:
fix #902


dunglas:
Thanks Antoine!

PR #903, opened 2019-10-01
Improve docs

lex111:
Varios docs improvements.

P.S. Have I made changes to the proper branch?

PR #911, opened 2019-10-04
Do not use stylized form of "NGINX"

teohhanhui:
Problem introduced in #903

See https://github.com/api-platform/docs/pull/903#pullrequestreview-296035235

Note: The "official form" is "nginx": https://nginx.org/en/

### DIFF
--- a/core/push-relations.md
+++ b/core/push-relations.md
@@ -29,9 +29,12 @@ class Book
 }
 ```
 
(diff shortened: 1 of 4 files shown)


alanpoulain:
It's not really clear.

teohhanhui:
If it can be written as both `nginx` and `NGINX`, then I think we should prefer `nginx` as it's more friendly (doesn't feel like shouting) and easier to type.

soyuka:
> It's not really clear.

He just wants a t-shirt so badly.

/s

Issue #884, opened 2019-09-24
[UX] New documentation hierarchy suggestion


bpolaszek:
Hey there,

_API-Platform_ is an ecosystem full of features and tools, but its documentation is growing fast, and it begins to lack of an appropriate structure to keep things tidy and clear.

I have identified several drawbacks on the current documentation that could lead astray newcomers:

* _The Api Component_ sublevels are far too many, and don't share the same level of importance (i.e _NelmioApiDocBundle_ being sibling of _Operations_, for instance)

* Starting by the _Distribution_ makes look like Api-Platform can't work without a proper Docker setup and all the things (_mercure_, _h2 push_, _varnish_, _let's encrypt_, _postgresql_, ...), which isn't true. 
With the advent of _Flex_, everything is done in Symfony to make applications _from simple to advanced_: you only `composer req` the components you need.
Beside, if we can assume the user has a minimal knowledge of Symfony (it is used under the hood along with _Doctrine_, share the same concepts, components and structure), _Api-Platform_ can work without _Docker_, so we shouldn't add _Docker_ into the learning curve if it's not mandatory.
As a consequence, I think the distribution should not be the very starting point, but exposed as a great tool to get a full-featured environment. 
The starting point could be a simple `composer req api && bin/console server:start`, to focus on the API itself instead of its environment.
In the meantime, example commands should not be prefixed by `docker-compose exec`.

* Similarly, some components (_schema generator_, _client generator_, ...) are useful tools but may not deserve a top-level position in the documentation hierarchy (generators being useful doesn't mean they are necessary). It make them look like they're completely part of Api-Platform and you can't work without them, which is wrong. At least it makes you focus on the stack instead of the core.
They could be grouped in a sub level and/or expanded in a separate documentation.

* Some topics cover easy setup as well as advanced coding and configuration. While this might be a good idea at first glance, most of the time it leads to extremely scrollable pages and things the user doesn't want to read at this moment. It also makes the information hard to find when you're looking for something. Hacks and tricks should be moved away in a specific section, in my opinion (and linked from the main topics they're related to).

* The documentation sections hierarchy should target newcomers _at first_, and go to experienced users _at the end_.

* An _F.A.Q._ or _How do I?_ section is really missing. Having such a section could also lead some developers to share their own recipes and help the community.

------------------

#### Read -> Install -> Configure -> Play -> Code -> Test -> Deploy -> Tune -> Hack

[...]

PR #913, opened 2019-10-04
Document file upload handling for GraphQL

lukasluecke:
This should cover what's needed to use file uploads with a GraphQL API (and with a link to a client implementation for Apollo).

### DIFF
--- a/core/file-upload.md
+++ b/core/file-upload.md
@@ -45,6 +45,7 @@ namespace App\Entity;
 use ApiPlatform\Core\Annotation\ApiProperty;
 use ApiPlatform\Core\Annotation\ApiResource;
 use App\Controller\CreateMediaObjectAction;
+use App\Resolver\CreateMediaObjectResolver;
 use Doctrine\ORM\Mapping as ORM;
 use Symfony\Component\HttpFoundation\File\File;
 use Symfony\Component\Serializer\Annotation\Groups;
@@ -82,6 +83,15 @@ use Vich\UploaderBundle\Mapping\Annotation as Vich;
  *     },
  *     itemOperations={
  *         "get",
+ *     },
+ *     graphql={
+ *         "upload"={
+ *             "mutation"=CreateMediaObjectResolver::class,
+ *             "deserialize"=false,
+ *             "args"={
+ *                 "file"={"type"="Upload!", "description"="The file to upload"}
+ *             }
+ *         },
  *     },
  * )
  * @Vich\Uploadable
@@ -129,8 +139,9 @@ class MediaObject
 
 ## Handling File Upload
 
-At this point, the entity is configured, but we still need to write the action
-that handles the file upload.
+At this point, the entity is configured, but we still need to write the handler for the uploaded file.
+
+### REST
 
 ```php
 <?php
@@ -150,6 +161,32 @@ final class CreateMediaObjectAction
         if (!$uploadedFile) {
             throw new BadRequestHttpException('"file" is required');
         }
+
+        $mediaObject = new MediaObject();
+        $mediaObject->file = $uploadedFile;
+
+        return $mediaObject;
+    }
+}
+```
+
+### GraphQL
+
+```php
+<?php
+// api/src/Resolver/CreateMediaObjectResolver.php
+
+namespace App\Resolver;
+
+use ApiPlatform\Core\GraphQl\Resolver\MutationResolverInterface;
+use App\Entity\MediaObject;
+use Symfony\Component\HttpFoundation\File\UploadedFile;
+
+final class CreateMediaObjectResolver implements MutationResolverInterface
+{
+    public function __invoke($item, array $context): MediaObject
+    {
+        $uploadedFile = $context['args']['input']['file'];
 
         $mediaObject = new MediaObject();
         $mediaObject->file = $uploadedFile;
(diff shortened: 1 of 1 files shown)


lukasluecke:
I'm not sure if we should split this whole "page" between REST and GraphQL, as to not mix them in the resource annotation, and then having to cover both for every step of the process.

lukasluecke:
Also the EventSubscriber for setting the resolved URL will not work with GraphQL, but I think we can fix that separately, by just replacing the whole thing with a custom normalizer (that should work for both REST and GraphQL, right?).

alanpoulain:
Thank you for working on this @lukasluecke!
For the other GraphQL features, we have already split the whole page between REST and GraphQL. We only reference the REST part when the code is exactly the same. So I think we should do the same for this feature.

core/file-upload.md
```diff
@@ -159,6 +170,32 @@ final class CreateMediaObjectAction
...
+
+namespace App\Resolver;
+
+use ApiPlatform\Core\GraphQl\Resolver\MutationResolverInterface;
+use App\Entity\MediaObject;
+use Symfony\Component\HttpFoundation\File\UploadedFile;
+
+final class CreateMediaObjectResolver implements MutationResolverInterface
+{
+    public function __invoke($item, array $context): MediaObject
```
alanpoulain: Missing `@param`.

core/file-upload.md
```diff
@@ -159,6 +170,32 @@ final class CreateMediaObjectAction
...
+    {
+        $uploadedFile = $context['args']['input']['file'];
+
```
alanpoulain: Please add a comment to explain the file should be processed here.

core/file-upload.md
```diff
@@ -159,6 +170,32 @@ final class CreateMediaObjectAction
...
+        $mediaObject = new MediaObject();
+        $mediaObject->file = $uploadedFile;
```
alanpoulain: Can we return the file directly like this?

lukasluecke:
@alanpoulain Regarding the review comments - currently my examples match the existing REST code pretty closely, so if we want to improve these, we should do it for both.

core/file-upload.md
lukasluecke: Could you explain what you mean?

core/file-upload.md
lukasluecke: What "processing" do you mean? There's nothing that **must** be done here for this to work.

core/file-upload.md
lukasluecke: See my comment below - it's the same as the existing REST example.

core/file-upload.md
alanpoulain: But if you upload a file, you need to do something with it, right? The comment tells you where to do it.

core/file-upload.md
alanpoulain: The REST `__invoke` doesn't need a PHPDoc (since it's already typehinted), but it's useful here.

core/file-upload.md
alanpoulain: I'm not sure the `UploadedFile` is easily serializable.

core/file-upload.md
lukasluecke: Alright, makes sense. We do not use the $item here, but I'll describe that as well 👍

core/file-upload.md
lukasluecke: That's what the VichUploaderBundle handles - you just set the file on the `MediaObject`, and everything else happens automatically. We do not do any processing in the existing REST example either.

core/file-upload.md
lukasluecke: Well it's exactly the same in the existing REST example, so I'd believe it works - otherwise you would probably configure that in the `MediaObject` resource serialization config, no?

core/file-upload.md
alanpoulain: Oh OK! @mahmoodbazdar didn't use the `VichUploaderBundle` in the tests so I was a little confused. Did you try the GraphQL file upload with this bundle? Is it working well?

core/file-upload.md
alanpoulain: OK I understand: it's because the property `file` is not serialized. Only `contentUrl` is.

core/file-upload.md
lukasluecke: No worries 😉 Yeah we‘re actively using it like this in our project, it‘s working fine (pretty much exactly the same like REST).

alanpoulain: Nice!

lukasluecke:
@alanpoulain Should I split it into a separate page as part of this PR, or do we want to handle that afterwards? So we‘d just add a „Handling File Upload“ page under the GraphQL section? And maybe add a link in the „original“ for discoverability?

alanpoulain:
The GraphQL page is rapidly growing, so I was thinking of splitting it into multiple pages, but it will probably be done after https://github.com/api-platform/docs/issues/884 (\cc @bpolaszek).

In the meantime, you should add a section `Handling File Upload` in the GraphQL page with the GraphQL specific code and reference the REST part when the code is the same. We should avoid duplication as much as possible.

In the future, it will be great to have the same page for all the possible configurations (REST/GraphQL/ORM/MongoDB ODM) with some selectors like the selector for the version, but it's a lot of work...

alanpoulain:
Any update on this @lukasluecke?

lukasluecke:
@alanpoulain Will try to finish this tomorrow 🙂

alanpoulain:
@lukasluecke I can take this over if you're okay with it.

alanpoulain:
Closed in favor of: https://github.com/api-platform/docs/pull/973.
Thank you @lukasluecke!

PR #916, opened 2019-10-05
Use decorated serializer to resolve url of uploaded files

lukasluecke:
This matches the recommended approach of using our extension points, and also works for GraphQL

### DIFF
--- a/core/file-upload.md
+++ b/core/file-upload.md
@@ -164,64 +164,65 @@ final class CreateMediaObjectAction
 Returning the plain file path on the filesystem where the file is stored is not useful for the client, which needs a
 URL to work with.
 
-An [event subscriber](events.md#custom-event-listeners) could be used to set the `contentUrl` property:
+We will be [decorating a serializer](serialization.md#decorating-a-serializer-and-adding-extra-data) to resolve the `contentUrl` property:
 
 ```php
 <?php
-// api/src/EventSubscriber/ResolveMediaObjectContentUrlSubscriber.php
+// api/src/Serializer/MediaObjectContentUrlNormalizer.php
 
-namespace App\EventSubscriber;
+namespace App\Serializer;
 
-use ApiPlatform\Core\EventListener\EventPriorities;
-use ApiPlatform\Core\Util\RequestAttributesExtractor;
 use App\Entity\MediaObject;
-use Symfony\Component\EventDispatcher\EventSubscriberInterface;
-use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\HttpKernel\Event\GetResponseForControllerResultEvent;
-use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+use Symfony\Component\Serializer\SerializerAwareInterface;
+use Symfony\Component\Serializer\SerializerInterface;
 use Vich\UploaderBundle\Storage\StorageInterface;
 
-final class ResolveMediaObjectContentUrlSubscriber implements EventSubscriberInterface
+final class MediaObjectContentUrlNormalizer implements NormalizerInterface, DenormalizerInterface, SerializerAwareInterface
 {
+    private $decorated;
     private $storage;
 
-    public function __construct(StorageInterface $storage)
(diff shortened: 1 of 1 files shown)


core/file-upload.md
```diff
@@ -164,64 +164,65 @@ final class CreateMediaObjectAction
...
+    public function supportsDenormalization($data, $type, $format = null)
+    {
+        return $this->decorated->supportsDenormalization($data, $type, $format);
+    }
 
-        foreach ($mediaObjects as $mediaObject) {
-            if (!$mediaObject instanceof MediaObject) {
-                continue;
-            }
+    public function denormalize($data, $class, $format = null, array $context = [])
```
alanpoulain: I don't think we need to denormalize a `MediaObject`. You can only use the `NormalizerInterface` IMO.

core/file-upload.md
```diff
@@ -164,64 +164,65 @@ final class CreateMediaObjectAction
...
+    {
+        return $this->decorated->denormalize($data, $class, $format, $context);
+    }
 
-            $mediaObject->contentUrl = $this->storage->resolveUri($mediaObject, 'file');
+    public function setSerializer(SerializerInterface $serializer)
+    {
+        if($this->decorated instanceof SerializerAwareInterface) {
```
alanpoulain:
```suggestion
        if ($this->decorated instanceof SerializerAwareInterface) {
```
I know this CS is in the other part as well.

core/file-upload.md
lukasluecke: You're probably right - I just took the example from the Serialization docs and modified it accordingly. Will update this 👍

core/file-upload.md
```diff
@@ -164,64 +164,65 @@ final class CreateMediaObjectAction
...
-use Symfony\Component\HttpKernel\Event\GetResponseForControllerResultEvent;
-use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+use Symfony\Component\Serializer\SerializerAwareInterface;
+use Symfony\Component\Serializer\SerializerInterface;
 use Vich\UploaderBundle\Storage\StorageInterface;
 
-final class ResolveMediaObjectContentUrlSubscriber implements EventSubscriberInterface
+final class MediaObjectContentUrlNormalizer implements NormalizerInterface, DenormalizerInterface, SerializerAwareInterface
```
teohhanhui: Just name it `MediaObjectNormalizer` :smile:

core/file-upload.md
teohhanhui: The problem is, if we're declaring this as a decorating service in Symfony DI, then it should implement the same interfaces. I'd suggest that we do **NOT** declare this as a decorating service (i.e. no `decorates`).

This way we can also remove `SerializerAwareInterface`.

lukasluecke: @teohhanhui Sounds good - does that change the behavior in any way? Is there any additional configuration required if you do not decorate the existing service?

core/file-upload.md
```diff
@@ -164,64 +164,65 @@ final class CreateMediaObjectAction
 Returning the plain file path on the filesystem where the file is stored is not useful for the client, which needs a
 URL to work with.
 
-An [event subscriber](events.md#custom-event-listeners) could be used to set the `contentUrl` property:
+We will be [decorating a serializer](serialization.md#decorating-a-serializer-and-adding-extra-data) to resolve the `contentUrl` property:
```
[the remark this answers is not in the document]
lukasluecke: @teohhanhui What do you mean? I took both the title and the link from the existing documentation.

(If we move away from the decoration, then this section has to be updated anyway.)

core/file-upload.md
lukasluecke: I tried to match the previous name somewhat, but shortening it is fine with me.

core/file-upload.md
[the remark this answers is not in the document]
dunglas: It could be renamed, but in another PR as here it's just a link.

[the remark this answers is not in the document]
dunglas: Let's rename then!

core/file-upload.md
dunglas: Let's rename then!

alanpoulain:
Superseded by https://github.com/api-platform/docs/pull/1339. Thank you @lukasluecke.

PR #919, opened 2019-10-08
Update authentication support

tienvx:
This PR do 2 things:
* Add example for handling authentication for api-platform/admin#237
* Remove old data provider guide. I think most of them are covered in https://api-platform.com/docs/admin/customizing/


dunglas:
Thank you very much @tienvx!

PR #920, opened 2019-10-08
Update File Upload page to match OAS3 specs

wassafr:
Current version of this page matches OAS2 specifications while the API Platform Distribution uses OAS3.


dunglas:
Thanks!

Issue #807, opened 2019-05-09
Default configuration is wrong


auipga:
If the values provided as default (see [2.4/core/configuration.md](https://github.com/api-platform/docs/blob/2.4/core/configuration.md)) really are the defaults, then I expect them to work.

The following defaults don't work:
```yaml
api_platform:
    title: ''                         # cannot contain an empty value
    description: ''                   # cannot contain an empty value
    collection:
        pagination:
            maximum_items_per_page: ~ # Expected int, but got NULL
    http_cache:
        max_age: ~                    # Expected int, but got NULL
        shared_max_age: ~             # Expected int, but got NULL
```

PR #921, opened 2019-10-09
Update configuration.md default to correct work

edamov:
Provided default values should work #807


dunglas:
Thanks @edamov. Close #807.

auipga:
Thank you

PR #929, opened 2019-10-11
Update getting-started.md

abdounikarim:
Hello,

I just fixed the typo for the nelmio_cors.yaml instead of nelmio-cors.yaml

Cheers 😉


dunglas:
Good catch, thanks!

PR #939, opened 2019-10-15
Update extensions.md

abdounikarim:
Hello,

I think the last condition brokes everything. I let you check because i'm not sure. 
And i just saw the tuto on SymfonyCasts, and Ryan don't use it.

Cheers 😉

| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #... 
| License       | MIT
| Doc PR        | api-platform/docs#...


dunglas:
It allows to use to not set the `WHERE` clause when no user is logged in (in CLI for instance). It looks legit to me.

PR #941, opened 2019-10-17
Added MIT license

mvegter:
Added MIT license to match the other repositories.


dunglas:
Thanks but the documentation is not under MIT license (it makes non sense for other things than software). As stated on the websitec the documentation is under CC-BY 3.0. Feel free to open a PR adding a readme including the license.

PR #948, opened 2019-10-22
Fix command to install Graphql

florent-andre:
* the second part of the command was executed on the host and not in the image, thus lead to errors.

| Q             | A
| ------------- | ---
| Doc fix?      | yes


dunglas:
Thanks Florent!

PR #949, opened 2019-10-23
Fix : patch_formats isn't an avaible option

Zowac:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| License       | MIT


Zowac:
> It is an available option. The current docs look good to me.

The option doesn't appear : https://github.com/api-platform/docs/blob/2.5/core/configuration.md

And when i tried to used it, i get an error

soyuka:
> It is an available option. The current docs look good to me.

PR #960, opened 2019-10-26
Document per operation DTO usage

julienfalque:
The documentation about using DTOs only shows `input` and `output` attributes at the resource configuration level. This is a bit misleading because one might think only one input DTO and only one output DTO are supported, while they can be different for each operation.


dunglas:
Nice one, thanks!

PR #981, opened 2019-11-07
Prevent uploaded files from being overwritten

jaapjan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | -
| License       | MIT
| Doc PR        | api-platform/docs#...

When someone uploads a file with an already existing name it overwrites the existing file. Easily prevented by providing this configuration.

### DIFF
--- a/core/file-upload.md
+++ b/core/file-upload.md
@@ -26,6 +26,7 @@ vich_uploader:
         media_object:
             uri_prefix: /media
             upload_destination: '%kernel.project_dir%/public/media'
+            # Will rename uploaded files using a uniqueid as a prefix.
             namer: Vich\UploaderBundle\Naming\OrignameNamer
 ```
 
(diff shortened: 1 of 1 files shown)


core/file-upload.md
```diff
@@ -26,6 +26,7 @@ vich_uploader:
         media_object:
             uri_prefix: /media
             upload_destination: '%kernel.project_dir%/public/media'
+            namer: Vich\UploaderBundle\Naming\OrignameNamer
```
soyuka: Could you add a comment on top of that explaining what it does?

jaapjan: Sure! Done ✅ Please let me know if it's ok like this.

soyuka:
thanks @jaapjan !

PR #995, opened 2019-11-24
added note about NODE_TLS_REJECT_UNAUTHORIZED to Troubleshooting

oleg-andreyev:
When running API Platform on development servers without valid TLS `client-generator` cannot generate client/interfaces because fetch is rejected.


soyuka:
Thanks @oleg-andreyev

PR #997, opened 2019-11-26
Add missing back-ticks to allow correct formatting

BonnieDoug:
Add missing back-ticks to allow correct formatting.


soyuka:
Thanks @BonnieDoug I think that @alOneh is working on this in https://github.com/api-platform/docs/pull/993 I'll let him fix that there.

PR #999, opened 2019-11-27
[Messenger] Add information about multiple handlers return value

fredericbarthelet:
When using Messenger integration, the current documentation specifies :
- use `output=false` to return blank response
- remove `output=false` if you want to return Data returned by your MessageHandler

However, no information is given on situation where multiple handlers are handling the same message.
- if the last handler returns void and resource configuration specifies that an output must be generated, a `No item route associated with the type` error is thrown due to the absence of any `GET` item operation configuration on the resource
- Ensuring that the handler returning data to be returned in the HTTP response process the message last solves this issue


soyuka:
thanks @fredericbarthelet !

PR #1004, opened 2019-11-30
file-upload-and-linking-admin-panel 

yudhir:
Use decorating your dataprovider feature of React Admin to upload a Multi-part File upload.

### DIFF
--- a/admin/managing-images-files.md
+++ b/admin/managing-images-files.md
@@ -1,0 +1,92 @@
+# Uploading Images and Linking to Other Entities
+
+Api-platform Core has file uploading support as shown in [Handling File Upload](https://api-platform.com/docs/master/core/file-upload/#handling-file-upload). With React admin's excellent [customization capabilities](customizing.md) we can add file and Image support in the admin panel . 
+
+## Decorating your Data Provider to add upload features
+
+Instead of writing your own Data Provider, you can enhance the capabilities of an existing data provider. You can even restrict the feature only for a given resource (entity) . 
+
+```javascript
+//in addUploadFeature.js
+
+//retreiving URL and Headers for AJAX request using fetch api . 
+const entrypoint = "http://localhost:8000/api";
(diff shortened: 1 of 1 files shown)


admin/managing-images-files.md
```diff
@@ -0,0 +1,91 @@
...
+const addUploadCapabilities = requestHandler => {
+    return (type, resource, params) => {
+
+        // Upload image to image_objects endpoint and link back the image
+        if (type === 'CREATE' && resource === 'books') {
+
+            if (params.data.file.rawFile instanceof File) {
+                const body = new FormData();
+                body.append("file", params.data.file.rawFile);
+                return (fetch(`${entrypoint}/image_objects`, {
```
dunglas, line 25: Maybe could we use `async`/`await` to simplify the code a bit?

dunglas (commented):
Thanks! We definitely need to have docs for this. I made some tiny modifications.

alanpoulain:
I think we don't need this anymore with https://api-platform.com/docs/admin/file-upload/.

admin/managing-images-files.md
yudhir: Sure .

yudhir:
Yeah , if it does the same thing , we don't need it .

alanpoulain:
You can use it to upload an entity with its related file too (it needs to be handled a little differently in the back but it is feasible: I've done it recently).
It will avoid to make 2 requests (and handle correctly the errors).

yudhir:
> 
> 
> You can use it to create an entity and upload its related file at the same time too (it needs to be handled a little differently in the back but it is feasible: I've done it recently).
> It will avoid to make 2 requests (and handle correctly the errors).

That could be added in the admin-docs as an example . It meets many practical usecases.

alanpoulain:
Yes it would be a nice addition indeed.
I don't know if I will have the time to do it, but you need two things:
- a custom `MultipartDecoder`
- a custom denormalizer
With this, a `POST` endpoint can be used with `multipart/form-data` (file + data).

PR #1006, opened 2019-12-01
Add extra tip to demo body of first `Review`

ralbear:
Because this explanation is in the introduction part of the guide, i considerate useful have extra tip to prevent people get stuck on the **IRI** error you get when use the original content and uses symfony flex.

### DIFF
--- a/distribution/index.md
+++ b/distribution/index.md
@@ -471,7 +471,7 @@ Now, add a review for this book using the `POST` operation for the `Review` reso
     "publicationDate": "September 21, 2016"
 }
 ```
-**Tip:** If you are using symfony flex, the content of the key `book` should be `"/api/books/1"`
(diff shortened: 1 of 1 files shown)


distribution/index.md
```diff
@@ -471,6 +471,7 @@ Now, add a review for this book using the `POST` operation for the `Review` reso
     "publicationDate": "September 21, 2016"
 }
 ```
+**Tip:** If you are using symfony flex, the content of the key `book` should be `"/api/books/1"`
```
GregoireHebert, line 474: It's not related to flex, but the prefix you configured in your routing.
I think this addition can bring confusion.

dunglas, line 474:
```suggestion
**Note:** If you are using have installed API Platform in an existing project using `composer`, the content of the key `book` must be `"/api/books/1"`
```

PR #1012, opened 2019-12-24
Fix wrong operation name for subresource operation

teohhanhui:
Currently the route name is used as the operation name, which is not ideal, but it is what it is.


soyuka:
thanks @teohhanhui

Charl13:
@teohhanhui @soyuka Currently at 2.6.3 this change in documentation doens't reflect functionality. In other words, `answer_get_subresource` works and `api_questions_answer_get_subresource` does not.

PR #1013, opened 2019-12-26
Add missing method attribute for YAML example

MrNicodemuz:
The current documentation mentions that:
```In the previous examples, the method attribute is mandatory, because the operation name doesn't match a supported HTTP method.```

Yet, the method attribute was missing in the YAML example.


teohhanhui:
Thanks @MrNicodemuz! :tada:

PR #1016, opened 2020-01-06
Fix the list of services and their ports on the Distribution page

Deuchnord:
The list of services was out of date:

- Add Vulcain in the services list
- Update the opened ports

### DIFF
--- a/distribution/index.md
+++ b/distribution/index.md
@@ -82,7 +82,6 @@ This starts the following services:
 | api      | The HTTP server for the API (NGINX)                               | n/a                                                         | all                                                |
 | vulcain  | The [Vulcain](https://vulcain.rocks) gateaway                     | 8443                                                        | all (prefer using a managed service in prod)       |
 | mercure  | The Mercure hub, [for real-time capabilities](../core/mercure.md) | 1337                                                        | all (prefer using the managed version in prod)     |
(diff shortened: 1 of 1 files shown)


Deuchnord:
> I don't understand what the `prod` environment means in this table.

No sure to understand what you mean, but as I understand the table, the _Environment_ column tells on what environments the service should be run, isn't it?

> We should also remove the `h2-proxy` row.

Good catch! :+1:

alanpoulain:
It seems to be the case but it's not very clear IMO.

alanpoulain:
Thank you @Deuchnord :slightly_smiling_face:

distribution/index.md
```diff
@@ -77,11 +77,11 @@ This starts the following services:
...
-| admin    | A development server for the admin                                | 81                                                          | dev (use a static website hosting service in prod) |
-| api      | The HTTP server for the API (NGINX)                               | 8080                                                        | all                                                |
+| client   | A development server for the Progressive Web App                  | 443                                                         | dev (use a static website hosting service in prod) |
+| admin    | A development server for the admin                                | 444                                                         | dev (use a static website hosting service in prod) |
+| api      | The HTTP server for the API (NGINX)                               | n/a                                                         | all                                                |
+| vulcain  | The [Vulcain](https://vulcain.rocks) gateaway                     | 8443                                                        | all (prefer using a managed service in prod)       |
```
teohhanhui: > (prefer using a managed service in prod)

Don't think there is one? :see_no_evil:

And considering that it has to be run with a specific `UPSTREAM` value, not sure if it's possible to have a shared service...

PR #1017, opened 2020-01-06
Update filters.md

luca-nardelli:
Today I digged through the sources of ExistsFilter because I had to understand whether it worked for collections or not, and I could not properly understand it from the docs. This PR adds an entry to the docs to clarify that behavior.

### DIFF
--- a/core/filters.md
+++ b/core/filters.md
@@ -405,6 +405,10 @@ Given that the collection endpoint is `/offers`, you can filter offers on nullab
 
 It will return all offers where `transportFees` is not `null`.
 
+#### Exists Filter and toMany relations
+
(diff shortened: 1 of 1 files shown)


luca-nardelli:
It makes sense, should I close this PR?

GregoireHebert:
@alanpoulain 
it seems that at the moment it works in MongoDB, since an empty embedded document is just a field that is ignored. A proposal has been made to replace it by an empty [] or {}, but it's a BC break so it's not merged, and won't be shortly without a BC. 

https://github.com/doctrine/mongodb-odm/pull/1624

I think we can close this PR.

PR #1018, opened 2020-01-14
Update authentication-support.md

e-belair:
Help user to use the good React version of docs


alanpoulain:
API Platform and API Platform Admin are not directly related. Moreover API Platform Admin 2 is now using react-admin 3.

alanpoulain:
I find it strange to be written here. If we do it like this, we should do it at every place where react-admin documentation is referenced and I don't think we should.
Moreover I think it's pretty common for a documentation to have a selector for a version.

PR #1019, opened 2020-01-14
add comment to use openapi_context

amine2z:
add comment to use openapi_context attribute when using openapi V3 specification


soyuka:
thanks @amine2z

PR #1028, opened 2020-01-30
Add example of decorating the built-in data persisters

jamesisaac:
This use-case is alluded to throughout the docs, but I thought it would be nice to have a concrete example, as I imagine hooking in custom business logic is a very common need.


core/data-persisters.md
```diff
@@ -70,3 +70,73 @@ services:
 ```
 
 Note that if you don't need any `$context` in your data persister's methods, you can implement the [`DataPersisterInterface`](https://github.com/api-platform/core/blob/master/src/DataPersister/DataPersisterInterface.php) instead.
+
+## Decorating the Built-In Data Persisters
+
+If you want to execute custom business logic before, during or after peristence, this can be achieved by [decorating](https://symfony.com/doc/current/service_container/service_decoration.html) the built-in data persisters.
```
alanpoulain:
```suggestion
If you want to execute custom business logic before or after peristence, this can be achieved by [decorating](https://symfony.com/doc/current/service_container/service_decoration.html) the built-in data persisters.
```

core/data-persisters.md
```diff
@@ -70,3 +70,73 @@ services:
...
+}
+```
+
+Even with service autowiring and autoconfiguration enabled, you must still configure the decoration:
+
+```yaml
+# api/config/services.yaml
+services:
+    # ...
+    'App\DataPersister\UserDataPersister':
```
alanpoulain:
```suggestion
    App\DataPersister\UserDataPersister:
```

alanpoulain:
Nice addition, thank you! You even considered REST and GraphQL :slightly_smiling_face:

jamesisaac:
@alanpoulain Thanks for your quick feedback!

For the 2 points you noted, I wasn't sure what the recommended style/language was in API Platform, so I was basing it off other nearby parts of the docs :wink: so I've made your recommended changes there too for consistency.

> Nice addition, thank you! You even considered REST and GraphQL 🙂

Of course!  I love how this abstraction means it's just one extra line to support both types of API :grin:

alanpoulain:
I understand why you did this, but I don't think the documentation is consistent enough to use it as reference :smile:

soyuka:
thanks @jamesisaac

PR #1029, opened 2020-01-30
Add note about Doctrine DataPersister not called when using messenger

vincentchalamon:


alanpoulain:
I would use "data persister" instead of "DataPersister".

dunglas:
Thanks @vincentchalamon

PR #1030, opened 2020-01-30
Add missing information about the PATCH operation

markwatney2016:


markwatney2016:
I just found the following on https://api-platform.com/docs/core/operations/:

> PATCH is also supported when using the JSON API format, as required by the specification

I am not sure if PATCH was intentionally not mentioned on the page I created this pull request for.

alanpoulain:
I think it's OK to mention it but IMO it would be good to mention it's not enabled by default if you don't use JSON:API.

soyuka:
thanks @markwatney2016

PR #1031, opened 2020-02-03
updated complete configuration

freezy-sk:
updated complete configuration with right order of params and default values


soyuka:
nice thanks @freezy-sk

PR #1037, opened 2020-02-11
add missing return statement in JS code

rendler-denis:
without the return statement JS throws an error: 

> Expected an assignment or function call and instead saw an expression  no-unused-expressions

this commit fixes this issue

### DIFF
--- a/admin/authentication-support.md
+++ b/admin/authentication-support.md
@@ -31,7 +31,7 @@ const apiDocumentationParser = entrypoint => parseHydraDocumentation(entrypoint,
                         api: result.api,
                         customRoutes: [
                             <Route path="/" render={() => {
-                                window.localStorage.getItem("token") ? window.location.reload() : <Redirect to="/login" />
+                                return window.localStorage.getItem("token") ? window.location.reload() : <Redirect to="/login" />
                             }} />
                         ],
                     });


alanpoulain:
It would be better to remove the curly brackets instead.

rendler-denis:
hey, @alanpoulain I've tried removing the bracelets like so:
`<Route path="/" render={() => window.localStorage.getItem("token") ? window.location.reload() : <Redirect to="/login"/>;`

but then the token is no longer used. I am new to React, mostly BE development and this was the only working solution I could find.

admin/authentication-support.md
```diff
@@ -31,7 +31,7 @@ const apiDocumentationParser = entrypoint => parseHydraDocumentation(entrypoint,
                         api: result.api,
                         customRoutes: [
                             <Route path="/" render={() => {
-                                window.localStorage.getItem("token") ? window.location.reload() : <Redirect to="/login" />
+                                return window.localStorage.getItem("token") ? window.location.reload() : <Redirect to="/login" />
                             }} />
```
soyuka: This doesn't look good, render should return a component not a `window.location.reload`, anyway without `return` it won't work, merging.

soyuka:
thanks @rendler-denis

rendler-denis:
glad I could help, @soyuka

PR #1040, opened 2020-02-19
Note about serialization groups related to DTO

EmilCataranciuc:
Currently it is not quite clear that serialization group annotations should be used on classes used for data transformation. At least not explicitly. See the note for inclusion proposal.


soyuka:
thanks @EmilCataranciuc

PR #1045, opened 2020-02-23
Install react-native-router-flux sub dependencies

basisak:
React-native-router-flux has sub dependencies that are not currently being installed. As of npm@3, neither yarn nor npm install sub dependencies by default.

If a new user is not familiar with the "has unmet peer dependency" warning and fails to install the listed sub dependencies, then expo will fail with errors. By adding the sub dependencies onto the existing yarn command then the new user will have one less thing to troubleshoot.


dunglas:
Thanks @basisak!

PR #1046, opened 2020-02-23
Add array example for openapi_context

reggermont:
I was facing this need and can't find the answer on the documentation. At first thought, I was trying to do something like `example: "["/assets/img/photo1.jpg", "/assets/img/photo2.jpg"]"` but it was returning errors or displaying the full string, as `example` is expecting a `PlainText` value

Had to try different methods to finally find my answer.

I think it should be in the documentation

Also fixed some old `swagger_context` declarations, as the last line says 
```md
To pass a context to the OpenAPI **v3** generator, use the `openapi_context` attribute (notice the prefix: `openapi_` instead of `swagger_`).
```


reggermont:
Hmm ... Should pull the project and edit myself, instead of using GitHub website

Close it to reopen a proper MR

PR #1047, opened 2020-02-23
Add array example for openapi_context

reggermont:
I was facing this need and can't find the answer on the documentation. At first thought, I was trying to do something like `example: "["/assets/img/photo1.jpg", "/assets/img/photo2.jpg"]"` but it was returning errors or displaying the full string, as `example` is expecting a `PlainText` value

Had to try different methods to finally find my answer.

I think it should be in the documentation

Also fixed some old `swagger_context` declarations, as the last line says 
```md
To pass a context to the OpenAPI **v3** generator, use the `openapi_context` attribute (notice the prefix: `openapi_` instead of `swagger_`).
```

### DIFF
--- a/core/swagger.md
+++ b/core/swagger.md
@@ -149,12 +149,31 @@ class Product // The class name will be used to name exposed resources
     public $name;
 
     /**
+     * @param array $name List of product pictures' relative URLs.
+     *
+     * @ORM\Column(type="simple_array", nullable=true)
+     *
+     * @ApiProperty(
+     *     attributes={
+     *         "openapi_context"={
+     *             "type"="array",
+     *             "items"={
+     *                 "type"="string",
+     *                 "example"="/assets/img/photo.jpg"
+     *             }
+     *         }
+     *     }
+     * )
+     */
+    public $pictures;
+
+    /**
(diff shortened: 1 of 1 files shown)


core/swagger.md
```diff
@@ -173,13 +192,20 @@ resources:
...
+          attributes:
+            openapi_context:
+              type: array
+              items: 
+                type: string
+                example: /assets/img/photo.jpg
         timestamp:
           attributes:
-            swagger_context:
+            openapi_context:
```
GregoireHebert, line 208: can you also change the [line 216](https://github.com/api-platform/docs/pull/1047/files#diff-5341978ee8bd732fef5d82209f643e57R216) to set
`"openapi": "3.0.2",`
instead of 
`"swagger": "2.0",`

And add a sentence [line 107](https://github.com/api-platform/docs/pull/1047/files#diff-5341978ee8bd732fef5d82209f643e57R107) to explain that using v2 it refers to `swagger_context` and v3 to `openapi_context`

vincentchalamon:
Please target 2.6 branch and rebase from it

vinceAmstoutz:

PR #1049, opened 2020-02-24
Update xdebug installation guide

gries:
XDEBUG_VERSION=2.7.2 is not compatible with php 7.4 and therefore has to be a higher version for api-platform 2.5.4


teohhanhui:
Thanks @gries! :tada:

PR #1052, opened 2020-02-24
Update data-persisters.md

siegfried1e:
Avoid recursion using the context.


soyuka:
mhhh this should definitely not happend in the first place

PR #1053, opened 2020-02-25
Use yarn dlx instead of npx

dunglas:

### DIFF
--- a/client-generator/nextjs.md
+++ b/client-generator/nextjs.md
@@ -10,9 +10,7 @@ The Next.js Client Generator generates components for Server Side Rendered appli
 
 Create a [Next.js application with express server](https://github.com/zeit/next.js/tree/canary/examples/custom-server-express). The easiest way is to execute:  
 
-    $ npx create-next-app your-app-name
-    # or
-    $ yarn create next-app your-app-name
+    $ yarn dlx create-next-app your-app-name
 
 ### Installing the Generator Dependencies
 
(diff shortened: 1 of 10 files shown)


dunglas:
Do not merge until the official release of Yarn 2.

client-generator/nextjs.md
```diff
@@ -10,8 +10,6 @@ The Next.js Client Generator generates components for Server Side Rendered appli
 
 Create a [Next.js application with express server](https://github.com/zeit/next.js/tree/canary/examples/custom-server-express). The easiest way is to execute:  
 
-    $ npx create-next-app your-app-name
-    # or
     $ yarn create next-app your-app-name
```
teohhanhui: There is no longer a `yarn create` command in yarn v2: https://next.yarnpkg.com/cli/create

As it was [a shortcut](https://classic.yarnpkg.com/en/docs/cli/create/), we should now do:

    $ yarn dlx create-* ...

teohhanhui: There is no longer a `yarn create` command in yarn v2.

As it was [a shortcut](https://classic.yarnpkg.com/en/docs/cli/create/), we should now do:

    $ yarn dlx create-* ...

dunglas: But Yarn 2 cannot be installed globally (you need Yarn 1 to install it) so it doesn't work.

PR #1063, opened 2020-03-01
Fix unsecured resource access using JWT integration

### LINKED ISSUE #1058: JWT integration docs is not complete
There are 2 types of integration with `LexikJWTAuthenticationBundle`

* from the root of `path` (`/`): https://api-platform.com/docs/core/jwt/#configuring-the-symfony-securitybundle
* from the prefix (`/api`): https://api-platform.com/docs/core/jwt/#adding-authentication-to-an-api-which-uses-a-path-prefix

IMO, both of them are incomplete. 

Example: when I use the first type of integration (by default), where all API endpoints start from the `/` root path, I can successfully fetch the `GET /greetings` without any token after JWT integration is done.

There is even a proof [in the tests](https://github.com/api-platform/demo/blob/5ecd2177087afbcce8f88ddfedbc7b95e738d961/api/tests/BooksTest.php#L17) of `api-platform/demo`.

What I would suggest adding is several sentences on how to configure security to **require** a JWT toker for `GET /greetings`

I tried this

```yaml
access_control:
        - { path: ^/authentication_token, roles: IS_AUTHENTICATED_ANONYMOUSLY }
        - { path: ^/,       roles: IS_AUTHENTICATED_FULLY }
```

and it works, but after this, Swagger Documentation is no longer works since it is on `/` path.

*Do you know how to properly configure it*?

The same applies for the second type of integration - when all the endpoints start from `/api` prefix.

I suggest to add something like this to the docs:

```yaml
access_control:
        - { path: ^/authentication_token, roles: IS_AUTHENTICATED_ANONYMOUSLY }
        - { path: ^/api,       roles: IS_AUTHENTICATED_FULLY }
```

Actually, this is what `LexikJWTAuthenticationBundle` [proposes in its docs](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Resources/doc/index.md#configuration).

chalasr:
Fixes #1058

### DIFF
--- a/core/jwt.md
+++ b/core/jwt.md
@@ -146,7 +146,7 @@ security:
                 failure_handler: lexik_jwt_authentication.handler.authentication_failure
 
     access_control:
-        - { path: ^/docs, roles: IS_AUTHENTICATED_ANONYMOUSLY } # Allows accessing the Swagger UI
+        - { path: ^/api/docs, roles: IS_AUTHENTICATED_ANONYMOUSLY } # Allows accessing the Swagger UI
         - { path: ^/authentication_token, roles: IS_AUTHENTICATED_ANONYMOUSLY }
         - { path: ^/, roles: IS_AUTHENTICATED_FULLY }
 ```
--- a/core/jwt.md
+++ b/core/jwt.md
@@ -83,6 +83,11 @@ security:
             guard:
                 authenticators:
                     - lexik_jwt_authentication.jwt_token_authenticator
+
+    access_control:
+        - { path: ^/docs, roles: IS_AUTHENTICATED_ANONYMOUSLY } # Allows accessing the Swagger UI
+        - { path: ^/authentication_token, roles: IS_AUTHENTICATED_ANONYMOUSLY }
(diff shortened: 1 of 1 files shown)


core/jwt.md
```diff
@@ -139,6 +144,11 @@ security:
                 password_path: password
                 success_handler: lexik_jwt_authentication.handler.authentication_success
                 failure_handler: lexik_jwt_authentication.handler.authentication_failure
+
+    access_control:
+        - { path: ^/docs, roles: IS_AUTHENTICATED_ANONYMOUSLY } # Allows accessing the Swagger UI
```
maks-rafalko: it should be `/api/docs`, just checked locally where all API endpoints are prefixed with `/api`

maks-rafalko:
```suggestion
        - { path: ^/api/docs, roles: IS_AUTHENTICATED_ANONYMOUSLY } # Allows accessing the Swagger UI
```

chalasr: Good catch, thanks!

soyuka:
thanks @chalasr !

PR #1065, opened 2020-03-03
Use `security` attribute instead of deprecated `access_control`

jonag:
Just a little patch to replace the deprecated attribute `access_control` with `security`. Don't know if the patch should target this branch or master. Let me know if I need to change.


alanpoulain:
Thanks @jonag.

PR #1070, opened 2020-03-20
Named sendWelcomeEmail function consistently

arch1pel:
Docs show calling of function sendWelcomeMail. But function itself is named sendWelcome**Em**ail


soyuka:
thanks @arch1pel

PR #1071, opened 2020-03-23
fix the subresource operation name when overriding path

bendavies:
This operation name was wrong.
The prefix must be removed when overriding a path

### DIFF
--- a/core/subresources.md
+++ b/core/subresources.md
@@ -207,7 +207,7 @@ You can control the path of subresources with the `path` option of the `subresou
  * ...
  * @ApiResource(
  *      subresourceOperations={
- *          "api_questions_answer_get_subresource"={
+ *          "answer_get_subresource"={
  *              "method"="GET",
  *              "path"="/questions/{id}/all-answers"
  *          },


core/subresources.md
```diff
@@ -207,7 +207,7 @@ You can control the path of subresources with the `path` option of the `subresou
  * ...
  * @ApiResource(
  *      subresourceOperations={
- *          "api_questions_answer_get_subresource"={
+ *          "answer_get_subresource"={
```
soyuka: this looks buggy, it should be the route name you get under `bin/console debug:router`, we definitely need to improve the subresources DX.

AdamKasp: information from lines https://github.com/api-platform/docs/pull/1071/files#diff-3a44ca00264369c7b87966b5596aeb33R194-R196
should also be updated, for now, it is wrong information.

dunglas:
@soyuka what about this PR? Should we close it?

soyuka:
It's not resolved I guess we can close it and related on https://github.com/api-platform/core/issues/3458 instead while this isn't fixed...

bendavies:
as far as i can remember the docs are incorrect so this was good to merge?

soyuka:
I do have the `api_questions_answer_get_subresource` name on my `bin/console` debug:router name

PR #1072, opened 2020-03-23
JWT documentation: Add separate firewall login exemple

atomatis:
Somebodies (like me) have troubles with this configuration (https://github.com/lexik/LexikJWTAuthenticationBundle/issues/489)
because login pattern must be equal to json/login check_path value.
I want add an exemple for help to resolv this very invisible issue.

### DIFF
--- a/core/jwt.md
+++ b/core/jwt.md
@@ -151,6 +151,51 @@ security:
         - { path: ^/, roles: IS_AUTHENTICATED_FULLY }
 ```
 
+### Use separate login firewall
+
+If you want use a dedied firewall for our login, you can do something like instead of:
+
+```yaml
+# api/config/packages/security.yaml
+security:
+    encoders:
+        App\Entity\User:
+            algorithm: argon2i
+
+    # https://symfony.com/doc/current/security.html#where-do-users-come-from-user-providers
+    providers:
+        # used to reload user from session & other features (e.g. switch_user)
+        app_user_provider:
+            entity:
+                class: App\Entity\User
+                property: email
+
+    firewalls:
+        dev:
+            pattern: ^/_(profiler|wdt)
+            security: false
+        login:
+            pattern: ^/authentication_token #must be same of json_login: check_path
+            stateless: true
+            anonymous: true
+            json_login:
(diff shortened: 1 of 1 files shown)


core/jwt.md
```diff
@@ -151,6 +151,51 @@ security:
...
+            pattern: ^/_(profiler|wdt)
+            security: false
+        login:
+            pattern: ^/authentication_token #must be same of json_login: check_path
+            stateless: true
+            anonymous: true
+            json_login:
+                check_path: /authentication_token
+                success_handler: lexik_jwt_authentication.handler.authentication_success
+                failure_handler: lexik_jwt_authentication.handler.authentication_failure
```
soyuka: We could only show this part as the rest is similar?

core/jwt.md
```diff
@@ -151,6 +151,51 @@ security:
         - { path: ^/, roles: IS_AUTHENTICATED_FULLY }
 ```
 
+### Use separate login firewall
+
+If you want use a dedied firewall for our login, you can do something like instead of:
```
maks-rafalko:
```suggestion
If you want to use a dedcated firewall for login, you can do something like this instead:
```

maks-rafalko: probably it's just me, but could you please explain what issue does this solution solves?

atomatis: Yes, of course (sorry, this is my try to first contrib on another project, i'm probably a little clumsy)

I try to use JWT token on my api-platform project and i follow this documentation but i want use another firewall for my authentication path.
So i'm write a security.yaml based on the 2th example but my JWT token don't work when i use it on ^/API path because the user stay null.
This is exactly the same case on that  **[issue](https://github.com/lexik/LexikJWTAuthenticationBundle/issues/489)**

For solve this, the "pattern" field must be set with the same value of "form_login:check_path:" value.
This is hard to debug and i think a newcomer on api-platform (like me) who want try this can be easily lock on this case. I want just improve the documentation for avoid that.

GregoireHebert:
I am not sure this PR, brings that much to the documentation.
I think this could be closed.

PR #1073, opened 2020-03-24
Update debugging.md

cnizzardini:
Configurations necessary for Mac + docker desktop setups per https://github.com/api-platform/api-platform/issues/796

### DIFF
--- a/distribution/debugging.md
+++ b/distribution/debugging.md
@@ -54,6 +54,20 @@ services:
       PHP_IDE_CONFIG: serverName=api-platform
 ```
 
+Note for Mac environments use the following:
+
+```yml
+      XDEBUG_CONFIG: >-
+        remote_enable=1
+        remote_host=docker.for.mac.localhost
+        remote_connect_back=0
+        remote_port=9000
+        idekey=PHPSTORM
+        remote_autostart=1
+        remote_mode=req
+        remote_handler=dbgp
+```
+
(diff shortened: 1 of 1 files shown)


distribution/debugging.md
```diff
@@ -54,6 +54,20 @@ services:
       PHP_IDE_CONFIG: serverName=api-platform
 ```
 
+Note for Mac environments use the following:
+
+```yml
+      XDEBUG_CONFIG: >-
+        remote_enable=1
+        remote_host=docker.for.mac.localhost
```
GregoireHebert, line 62: the real key changer necessary here is this line alone no?

cnizzardini: I tested this a few different times locally on my macOs Mojava version 10.14.6 running docker desktop 2.1.0.5. I needed the following configurations for breakpoints to work with PHP Storm:

```yml
        remote_enable=1
        remote_host=docker.for.mac.localhost
        remote_connect_back=0
        remote_port=9000
        idekey=PHPSTORM
```

cnizzardini: So it seems the pertinent changes are 

```yml
        remote_host=host.docker.internal
        remote_connect_back=0
```

I am not sure why remote_connect_back must be turned off for it to work on my local, but that is the behavior I am getting.

dunglas:
Thanks @cnizzardini!

PR #1080, opened 2020-03-28
Add mandatory id and annotations

dhernandez:
Adding an id to Book entity class and with the annotations included the example works from scratch including OpenAPI docs.

### DIFF
--- a/core/dto.md
+++ b/core/dto.md
@@ -64,6 +64,8 @@ Similarly, the `output` attribute is used during [the serialization process](ser
 
 The `input` and `output` attributes are taken into account by all the documentation generators (GraphQL and OpenAPI, Hydra).
 
+Note that `Book` entity needs an id property. The simplest way is adding a public property called `$id`, as in the example. However, as in any other entity, you can use a private property, with getter and setter functions, and/or named it as you wish, provided you annotate it with `@ApiProperty (identifier = true)`. For instance, you could have a property called `$code`. So the `InputDataTransformer` actually transforms the isbn into a code. And then in the `OutputDataTransformer`, from this code into the name.
+
 To create a `Book`, we `POST` a data structure corresponding to the `BookInput` class and get back in the response a data structure corresponding to the `BookOutput` class:
 
 ![Diagram post input output](images/diagrams/api-platform-post-i-o.png)
--- a/core/dto.md
+++ b/core/dto.md
@@ -6,7 +6,7 @@ However, it's sometimes useful to use a specific class to represent the input or
 
 ## Specifying an Input or an Output Data Representation
 
-For a given resource class, you may want to have a different representation of this class as input (write) or output (read). 
+For a given resource class, you may want to have a different representation of this class as input (write) or output (read).
 To do so, a resource can take an input and/or an output class:
 
 ```php
--- a/core/dto.md
+++ b/core/dto.md
@@ -64,7 +64,7 @@ Similarly, the `output` attribute is used during [the serialization process](ser
 
 The `input` and `output` attributes are taken into account by all the documentation generators (GraphQL and OpenAPI, Hydra).
 
-Note that `Book` entity needs an id property. The simplest way is adding a public property called `$id`, as in the example. However, as in any other entity, you can use a private property, with getter and setter functions, and/or named it as you wish, provided you annotate it with `@ApiProperty (identifier = true)`. For instance, you could have a property called `$code`. So the `InputDataTransformer` actually transforms the isbn into a code. And then in the `OutputDataTransformer`, from this code into the name.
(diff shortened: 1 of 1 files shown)


core/dto.md
```diff
@@ -6,7 +6,7 @@ However, it's sometimes useful to use a specific class to represent the input or
 
 ## Specifying an Input or an Output Data Representation
 
-For a given resource class, you may want to have a different representation of this class as input (write) or output (read).
+For a given resource class, you may want to have a different representation of this class as input (write) or output (read).
```
GregoireHebert, line 9: can you remove the trailing space?

core/dto.md
```diff
@@ -27,6 +27,7 @@ use App\Dto\BookOutput;
  */
 final class Book
 {
+    public $id;
```
GregoireHebert, line 30: The example is needs also a `@ApiProperty(identifier=true)`
And to be honest, I should also have a property called `$code`
So the `InputDataTransformer` actually transform the isbn into a code.
And then in the `OutputDataTransformer`, from this code into the name.

core/dto.md
```diff
@@ -78,6 +79,9 @@ We have the following `BookInput`:
 namespace App\Dto;
 
 final class BookInput {
+  /**
+    * @var string
+  */
```
GregoireHebert, line 84: This is not mandatory, it works without.
It just won't show the property in the example or the textarea being pre-filled in swagger. But it's a nice addition.

core/dto.md
dhernandez: The example only needs the annotation ` @ApiProperty(identifier=true)` if you use another name to the property, like `$code`. But the example works without the annotation with the `$id` property.
Although, IMHO, Book entity shouldn't need a mandatory id property, because you have a BookInput and a BookOutput DTOs and BookInput have an id (called `isbn`).
So, to make a simple full example, that works with this issue, I think that adding an id called `$id` to Book is the clearest way.
Maybe, we can add also a note explaining the option of using `$code` property, annotated as identifier, and transforming it to isbn.

GregoireHebert: yup it's a good idea, at least to clarify the process.

core/dto.md
dhernandez: Fixed in https://github.com/api-platform/docs/pull/1080/commits/d094c193672c6e9113aaa8ca94f9f5f1e93b1e46

core/dto.md
dhernandez: You can check https://github.com/api-platform/docs/pull/1080/commits/fabb572ce889b4e347c743023c5569c76c5e800d :slightly_smiling_face:

dhernandez:
Hi! What's happened with the PR? This is my first contribution so I don't know if you need anything else to merge. Please, tell me.

core/dto.md
```diff
@@ -142,6 +148,9 @@ To manage the output, it's exactly the same process. For example, we have the fo
 namespace App\Dto;
 
 final class BookOutput {
+  /**
+    * @var string
+  */
```
alanpoulain, line 153:
```suggestion
   */
```

alanpoulain, line 153:
```suggestion
   * @var string
   */
```

core/dto.md
```diff
@@ -78,6 +81,9 @@ We have the following `BookInput`:
 namespace App\Dto;
 
 final class BookInput {
+  /**
+    * @var string
+  */
```
alanpoulain, line 86:
```suggestion
   * @var string
   */
```

core/dto.md
```diff
@@ -63,6 +64,8 @@ Similarly, the `output` attribute is used during [the serialization process](ser
 
 The `input` and `output` attributes are taken into account by all the documentation generators (GraphQL and OpenAPI, Hydra).
 
+Note that `Book` entity needs an id property. The simplest way is adding a public property called `$id`, as in the example. However, as in any other entity, you can use a private property, with getter and setter functions, and/or named it as you wish, provided you annotate it with `@ApiProperty (identifier = true)`. For instance, you could have a property called `$code`. So the `InputDataTransformer` actually transforms the isbn into a code. And then in the `OutputDataTransformer`, from this code into the name.
```
alanpoulain, line 67:
```suggestion
Note that `Book` entity needs an id property. The simplest way is adding a public property called `$id`, as in the example. However, as in any other entity, you can use a private property, with getter and setter functions, and/or named it as you wish, provided you annotate it with `@ApiProperty(identifier=true)`. For instance, you could have a property called `$code`. So the `InputDataTransformer` actually transforms the isbn into a code. And then in the `OutputDataTransformer`, from this code into the name.
```

core/dto.md
```diff
@@ -142,6 +148,9 @@ To manage the output, it's exactly the same process. For example, we have the fo
 namespace App\Dto;
 
 final class BookOutput {
+  /**
+   * @var string
+   */
   public $name;
```
alanpoulain, line 154:
```suggestion
    /**
     * @var string
     */
    public $name;
```

core/dto.md
```diff
@@ -78,6 +81,9 @@ We have the following `BookInput`:
 namespace App\Dto;
 
 final class BookInput {
+  /**
+   * @var string
+   */
   public $isbn;
```
alanpoulain, line 87:
```suggestion
    /**
     * @var string
     */
    public $isbn;
```

alanpoulain:
Thank you @dhernandez.

PR #1082, opened 2020-03-30
Fix testing examples

norkunas:
Two things fixed:
 - `findIriBy` method is non static, so it shouldn't be accessed statically;
 - make `testLogin` as public method otherwise phpunit cannot access it


alanpoulain:
Thanks @norkunas.

PR #1083, opened 2020-03-31
Change deprecated bin/console server:start command to symfony server:…

nieuwenhuisen:
Change deprecated `bin/console server:start` command to `symfony server:start` in getting started section. And I added a link to de symfony local server documentation.

### DIFF
--- a/core/getting-started.md
+++ b/core/getting-started.md
@@ -228,7 +228,7 @@ If you want to serialize only a subset of your data, please refer to the [Serial
 **You're done!**
 
 You now have a fully featured API exposing your entities.
-Run the Symfony app (`bin/console server:run`) and browse the API entrypoint at `http://localhost:8000/api`.
+Run the Symfony app with the [Symfony Local Web Server](https://symfony.com/doc/current/setup/symfony_server.html) (`symfony server:start`) and browse the API entrypoint at `https://localhost:8000/api`.
 
 Interact with the API using a REST client (we recommend [Postman](https://www.getpostman.com/)) or an Hydra-aware application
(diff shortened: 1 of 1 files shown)


core/getting-started.md
```diff
@@ -228,7 +228,7 @@ If you want to serialize only a subset of your data, please refer to the [Serial
 **You're done!**
 
 You now have a fully featured API exposing your entities.
-Run the Symfony app (`bin/console server:run`) and browse the API entrypoint at `http://localhost:8000/api`.
+Run the Symfony app with the [Symfony Local Web Server](https://symfony.com/doc/current/setup/symfony_server.html) (`symfony server:start`) and browse the API entrypoint at `https://localhost:8000/api`.
```
GregoireHebert, line 231: I agree on the https part, but not on the command part.
If you only download the project itself, you do not have the symfony command available. But in both situation, the `bin/console` will work.

nieuwenhuisen: So far as I know the https part is only for the Symfony Local Web Server and not for the `bin/console server:run` command. Also the `bin/console server:run` command is deprecated and will be probably removed in the future. That's why I suggest the new Symfony binary. This is not installed by default, but is very easy to install with the given link.

GregoireHebert: I was not aware of it's deprecation :) 
Thanks for that 👍

alanpoulain, line 231:
```suggestion
Run the Symfony app with the [Symfony Local Web Server](https://symfony.com/doc/current/setup/symfony_server.html) (`symfony server:start`) and browse the API entrypoint at `http://localhost:8000/api`.
```

alanpoulain:
Thank you @nieuwenhuisen!

PR #1085, opened 2020-03-31
Fix: DataPersister decorator should call decorated supports()

amenophis:


alanpoulain:
Thank you @amenophis.

jamesisaac:
Ah I see, because decoration removes the original service?

Maybe decoration wasn't the best technique for this example in that case?  Because it would probably be cleaner to have 1 of these classes per entity/responsibility, rather than cramming lots of business logic into a single DataPersister?  (The same way it would have been done with EventListeners, which this technique supersedes).

What about injecting `DataPersisterInterface` into the constructor instead, rather than using decoration?  Although for that to work, it seems you have to bind `ApiPlatform\Core\DataPersister\ContextAwareDataPersisterInterface: '@api_platform.doctrine.orm.data_persister'` to avoid an infinite loop.

PR #1088, opened 2020-04-04
Add precision about exists filter and collection association

GregoireHebert:


alanpoulain:
Please target 2.5. We only maintain two branches: 2.5 and master.

PR #1089, opened 2020-04-04
Add precision about exists filter and collection association

GregoireHebert:
replace #1088


alanpoulain:
Thank you @GregoireHebert!

PR #1090, opened 2020-04-04
Add sequence validation example

GregoireHebert:

### DIFF
--- a/core/validation.md
+++ b/core/validation.md
@@ -360,7 +360,7 @@ class MySequencedGroup
 {
     public function __invoke()
     {
-        return new GroupSequence(['first', 'second']); // now, no matter who's first in the class declaration, it will be tested in this order.
+        return new GroupSequence(['first', 'second']); // now, no matter which is first in the class declaration, it will be tested in this order.
     }
 }
 ```
@@ -506,4 +506,3 @@ Constraints                                                                     
 [`Currency`](https://symfony.com/doc/current/reference/constraints/Currency.html)     | `http://schema.org/priceCurrency` |
 [`Isbn`](https://symfony.com/doc/current/reference/constraints/Isbn.html)             | `http://schema.org/isbn`          |
 [`Issn`](https://symfony.com/doc/current/reference/constraints/Issn.html)             | `http://schema.org/issn`          |
-
--- a/core/validation.md
+++ b/core/validation.md
@@ -412,6 +412,8 @@ class Greeting
      * @var string A nice person
      *
      * @ORM\Column
+     * 
+     * I want this "second" validation to be executed after the "first" one even though I wrote them in this order.
      * @One(groups={"second"})
      * @Two(groups={"first"})
      */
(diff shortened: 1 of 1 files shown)


core/validation.md
```diff
@@ -342,6 +342,88 @@ class Book
 }
 ```
 
+## Sequential Validation Groups
+
+If you need to determine the order in which your validation groups must be tested against, you can use a [group sequence](http://symfony.com/doc/current/validation/sequence_provider.html).
```
alanpoulain, line 347:
```suggestion
If you need to specify the order in which your validation groups must be tested against, you can use a [group sequence](http://symfony.com/doc/current/validation/sequence_provider.html).
```

core/validation.md
```diff
@@ -342,6 +342,88 @@ class Book
...
+
+namespace App\Validator;
+
+use Symfony\Component\Validator\Constraints\GroupSequence;
+
+class MySequencedGroup
+{
+    public function __invoke()
+    {
+        return new GroupSequence(['first', 'second']); // now, no matter who's first in the class declaration, it will be tested in this order.
```
alanpoulain, line 363:
```suggestion
        return new GroupSequence(['first', 'second']); // now, no matter which is first in the class declaration, it will be tested in this order.
```

core/validation.md
```diff
@@ -342,6 +342,88 @@ class Book
...
+class MySequencedGroup
+{
+    public function __invoke()
+    {
+        return new GroupSequence(['first', 'second']); // now, no matter which is first in the class declaration, it will be tested in this order.
+    }
+}
+```
+
+Just creating the class is not enough because Symfony does not see this service as being used. Therefor to prevent the service to be inlined, you need to enforce it to be public.
```
alanpoulain, line 368:
```suggestion
Just creating the class is not enough because Symfony does not see this service as being used. Therefore to prevent the service to be removed, you need to enforce it to be public.
```

alanpoulain: I don't think everyone understands "inlined".

core/validation.md
```diff
@@ -342,6 +342,88 @@ class Book
...
+     * @ORM\Column(type="integer")
+     */
+    private $id;
+
+    /**
+     * @var string A nice person
+     *
+     * @ORM\Column
+     * @One(groups={"second"})
+     * @Two(groups={"first"})
```
alanpoulain, line 416: Not really logical :stuck_out_tongue:

GregoireHebert: 😭

alanpoulain: Maybe you can add a comment?

GregoireHebert: done

alanpoulain:
Thanks @GregoireHebert :slightly_smiling_face:

norkunas:
Shouldnt this use the ValidationGroupsGeneratorInterface?

GregoireHebert:
@norkunas not in my example :)

PR #1093, opened 2020-04-10
Update dto.md

sfi-fgarnier:
Handle violations if errors occurred and let API Platform returns a list of violations normalized in Hydra format.

### DIFF
--- a/core/dto.md
+++ b/core/dto.md
@@ -16,6 +16,7 @@ To do so, a resource can take an input and/or an output class:
 namespace App\Entity;
 
 use ApiPlatform\Core\Annotation\ApiResource;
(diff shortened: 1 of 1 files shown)


sfi-fgarnier:
hi @alanpoulain ,

I'm using `Symfony\Component\Validator\Validator\ValidatorInterface` not `ApiPlatform\Core\Bridge\Symfony\Validator`.

Good point.
Thanks.

alanpoulain:
Maybe we should add a sentence to tell explicitly it's the API Platform validator then?

sfi-fgarnier:
Just lack of attention from me...
But it could be a good idea, yeah.

PR #1096, opened 2020-04-20
typo: link to #overriding-the-openapi-specification

Grafikart:
This link doesn't work in the current documentation.


dunglas:
Good catch, thanks!

PR #1098, opened 2020-04-24
Update serialization.md

AntoineRoue:
I'm not a PHP pro but for me there is no way that "$a ?? null" does not the same thing than "$a".


alanpoulain:
Yes it is. If the key doesn't exist in the array, PHP throws an error. The null coalescing operator avoids this behavior.

AntoineRoue:
Sorry for late answer, but if the "?? null" is only here to ignore the error, then the operator @ is more suited for this role, no ? However it's not very important and I don't know how to modify my commit.

PR #1103, opened 2020-05-04
Update security.md

michael-schaefer-eu:
Add an example for Voters description.


dunglas:
Nice addition, thanks @michael-schaefer-eu!

PR #1106, opened 2020-05-08
Improve the Serialization process guide page

UjinGE:
- Changed incorrect URL to Doctrine primary key manual.
- Added link to Doctrine composite identifiers guide page.
- Added link to Changing Identifier in a Doctrine Entity guide page.

### DIFF
--- a/core/serialization.md
+++ b/core/serialization.md
@@ -679,8 +679,8 @@ final class ApiNormalizer implements NormalizerInterface, DenormalizerInterface,
 
 ## Entity Identifier Case
 
-API Platform is able to guess the entity identifier using Doctrine metadata ([ORM](http://doctrine-orm.readthedocs.org/en/latest/reference/basic-mapping.html#identifiers-primary-keys), [MongoDB ODM](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/basic-mapping.html#identifiers)).
-For ORM, it also supports composite identifiers.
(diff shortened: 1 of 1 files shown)


core/serialization.md
```diff
@@ -736,6 +736,8 @@ must do the following:
 3. if you use Doctrine ORM, be sure to **not** mark this property with [the `@GeneratedValue` annotation](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/basic-mapping.html#identifier-generation-strategies)
   or use the `NONE` value
 
+For example [Changing Identifier in a Doctrine Entity](https://api-platform.com/docs/core/identifiers/#changing-identifier-in-a-doctrine-entity)
```
alanpoulain, line 739: Could you add a more explicit sentence? Also I would see it before, when Doctrine identifiers are still talking about.

UjinGE: Sorry for my bad English. I think that this link relating next information

> In some cases, you will want to set the identifier of a resource from the client (e.g. a client-side generated UUID, or a slug).

I detailed the text as best I could and raised it higher. If something is wrong, then please indicate what exactly and where to write.

core/serialization.md
```diff
@@ -727,7 +727,7 @@ App\Entity\Book:
             identifier: true
 ```
 
-In some cases, you will want to set the identifier of a resource from the client (e.g. a client-side generated UUID, or a slug).
+In some cases, you will want to set the identifier of a resource from the client (e.g. a client-side generated UUID, or a slug). For example, see how to [change identifier in a Doctrine Entity](https://api-platform.com/docs/core/identifiers/#changing-identifier-in-a-doctrine-entity)
```
alanpoulain, line 730: I think it doesn't really apply in this case. In this case, the Doctrine entity has no generated ID.

UjinGE: ok, I removed this link

alanpoulain:
Thank you @UjinGE!

PR #1120, opened 2020-05-30
Update default-order.md

amiralkhori:


alanpoulain:
Not really sure why it's needed?

Darkweak:
Already [there](https://api-platform.com/docs/core/default-order/#overriding-default-order)

PR #1128, opened 2020-06-19
Error of change property ?

guenneguezt:
As explain in :
For instance, you could have a property called `$code`. So the `InputDataTransformer` actually transforms the isbn into a code. And then in the `OutputDataTransformer`, from this code into the name.
It seems to be an error ?


soyuka:
The example is correct, we're using a data transfer object with an isbn property to setup a Book.

PR #1130, opened 2020-06-23
Change Interface of decorated DataPersister

danieljank:
If the decorated DataPersister is from type `DataPersisterInterface` there is no `$context` argument.
Alternatively we could stay with `DataPersisterInterface` and remove the `$context` argument in `supports()`, `persist()` and `remove()`

### DIFF
--- a/core/data-persisters.md
+++ b/core/data-persisters.md
@@ -90,7 +90,7 @@ final class UserDataPersister implements ContextAwareDataPersisterInterface
     private $decorated;
     private $mailer;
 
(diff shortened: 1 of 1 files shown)


soyuka:
Should we document instead that if the context is neeeded one could use the ContextAwareDataPersisterInterface

danieljank:
I don't think that we should blow up the documentation with that kind of information for this quite simple example.

My intention was to just make it consistent.

But that might just be a matter of taste. Could adjust that...

dunglas:
Thanks @danieljank!

PR #1131, opened 2020-06-24
Documentation Update: Disabling Elasticsearch for an operation

ndy40:
If you have added Elasticsearch to your project and later remove it or prefer to fetch the data from the database rather than from Elasticsearch, you can achieve this by adding the parameter `elasticsearch: false` to your operation. This forces the api endpoint to query the database instead of Elasticsearch for your data. 

Issue isn't visible in the documentation and cost us 2 days of investigating to figure it out.

### DIFF
--- a/core/operations.md
+++ b/core/operations.md
@@ -597,7 +597,8 @@ final class SwaggerDecorator implements NormalizerInterface
 That's it: your route is gone!
 
 ## Disabling ElasticSearch on Operations
-After enabling [Elasticsearch Support](elasticsearch.md) in your project, there may come a time you would prefer not to read your data from your Elasticsearch server and would rather fetch your data directly from your database. To achieve this, you can pass in a parameter to your api operation with the value `elasticsearch: true`. A simple example is below.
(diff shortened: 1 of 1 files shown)


core/operations.md
```diff
@@ -595,3 +595,45 @@ final class SwaggerDecorator implements NormalizerInterface
 ```
 
 That's it: your route is gone!
+
+## Disabling ElasticSearch on Operations
+After enabling [Elasticsearch Support](elasticsearch.md) in your project, there may come a time you would prefer not to read your data from your Elasticsearch server and would rather fetch your data directly from your database. To achieve this, you can pass in a parameter to your api operation with the value `elasticsearch: true`. A simple example is below.
```
soyuka, line 600:
```suggestion
After enabling [Elasticsearch Support](elasticsearch.md) in your project, there may come a time you would prefer not to read your data from your Elasticsearch server and would rather fetch your data directly from your database. To achieve this, you can pass in a parameter to your api operation with the value `elasticsearch: true`. 
For example:
```

ndy40:
@soyuka thanks for the suggestion. Merged the suggestion.

core/operations.md
```diff
@@ -595,3 +595,45 @@ final class SwaggerDecorator implements NormalizerInterface
 ```
 
 That's it: your route is gone!
+
+## Disabling ElasticSearch on Operations
```
dunglas, line 599: Could you move this docs in `elasticsearch.md` instead please? It will be easier to discover.

Also, can you add a blank line after the title?

PR #1133, opened 2020-06-25
Fix incorrect serialize_payload_fields default

benr77:
The current docs state that to serialize ALL payload fields, use `serialize_payload_fields: []`

Actually, this means NONE of the payload fields are serialized.

The default value, to serialize ALL payload fields, should be `serialize_payload_fields: ~`


soyuka:
nice thanks!

PR #1134, opened 2020-07-03
Fix instructions to install graphql-php library

falinsky:
There are breaking changes introduced in the latest major version of webonyx/graphql-php.


alanpoulain:
Thanks but the next patch version will be compatible with it (https://github.com/api-platform/core/pull/3621).

PR #1138, opened 2020-07-08
Fix vuetify generator docs

mrbookt:
Change "Update the scripts part of the new `composer.json`" to "Update the scripts part of the new `package.json`"


dunglas:
Good catch, thanks!

PR #1143, opened 2020-07-23
fix serialization getSum() bad indentation

FrancoisDoussin:
Just fix bad indentation in `core` -> `serialization` -> `Calculated Field` -> `getSum()` method !


alanpoulain:
Thanks!

PR #1146, opened 2020-08-04
Update serialization.md

rugolinifr:
Moves the "Calculated field" part outside the "Embedding relations" part.
Adds "Force IRI with relation of the same type" parts which describes the side effect of using serialization groups on a property of the same type of the current entity.
Adds the workaround using ApiProperty annotation described on the [issue 1001](https://github.com/api-platform/core/issues/1001#issuecomment-394162998).

### DIFF
--- a/core/serialization.md
+++ b/core/serialization.md
@@ -206,14 +206,15 @@ include the `name` property because of the specific configuration for this opera
 
 Refer to the [operations](operations.md) documentation to learn more.
 
-### Embedding Relations
+## Embedding Relations
 
 <p align="center" class="symfonycasts"><a href="https://symfonycasts.com/screencast/api-platform/relations?cid=apip"><img src="../distribution/images/symfonycasts-player.png" alt="Relations screencast"><br>Watch the Relations screencast</a></p>
 
 By default, the serializer provided with API Platform represents relations between objects using [dereferenceable IRIs](https://en.wikipedia.org/wiki/Internationalized_Resource_Identifier).
-They allow you to retrieve details for related objects by issuing extra HTTP requests.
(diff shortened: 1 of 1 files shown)


core/serialization.md
```diff
@@ -206,14 +206,15 @@ include the `name` property because of the specific configuration for this opera
...
+## Embedding Relations
 
 <p align="center" class="symfonycasts"><a href="https://symfonycasts.com/screencast/api-platform/relations?cid=apip"><img src="../distribution/images/symfonycasts-player.png" alt="Relations screencast"><br>Watch the Relations screencast</a></p>
 
 By default, the serializer provided with API Platform represents relations between objects using [dereferenceable IRIs](https://en.wikipedia.org/wiki/Internationalized_Resource_Identifier).
-They allow you to retrieve details for related objects by issuing extra HTTP requests.
+They allow you to retrieve details for related objects by issuing extra HTTP requests. However, for performance reasons, it is sometimes preferable to avoid forcing the client to issue extra HTTP requests.
 
-In the following JSON document, the relation from a book to an author is represented by an URI:
+### Serialization
```
soyuka, line 216: Normalization?

soyuka:
I like the change :+1:

core/serialization.md
rugolinifr: Indeed, my bad. Should i change it myself (and how)? I'm not familiar with github GUI.

dunglas: Yes you can do the change directly from the GitHub GUI. Just click on edit and to the change, GitHub will create an extra commit (and we will squash all commits when merging).

rugolinifr: thanks, it is done

soyuka:
thnaks @rugolinifr

PR #1148, opened 2020-08-06
Update performance.md

jb-reynaud:
## Why
`GetResponseEvent` is deprecated since 4.3, use RequestEvent instead

https://github.com/symfony/http-kernel/blob/4.4/Event/GetResponseEvent.php#L17


soyuka:
thanks @jb-reynaud

PR #1151, opened 2020-08-06
PR template: Describe what is last old-stable version

alexislefebvre:
By providing a link, users don't need to search for the page, they can find it directly.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,6 +1,6 @@
 <!--
 
-If your pull request fixes a BUG, use the last old-stable branch that contains the bug.
+If your pull request fixes a BUG, use the last stable branch that contains the bug.
 
 If your pull request documents a NEW FEATURE, use the `master` branch.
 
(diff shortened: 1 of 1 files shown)


alexislefebvre:
Does it need to target the 2.4 version so every PR will have the updated message?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,7 +1,9 @@
 <!--
 
-If your pull request fixes a BUG, use the last stable branch that contains the bug.
+If your pull request fixes a BUG, use the last old-stable branch (currently 2.5) that contains the bug.
```
soyuka, line 3: I'm not in favor of adding this here as we'd have to remember to change this file when the old-stable changes.

alexislefebvre: Yes, it can't be misleading if it's outdated. That's why I added an note on the `releases.md` page: https://github.com/api-platform/docs/pull/1151/commits/9640ad11268c526f4d5785fef7eb590bc81a935a#diff-6bc81b7c3f1ab223dbcacc61496efec7R7

If you think it won't work and people will miss it, we can remove the version here.

soyuka: please remove the version

soyuka:
thnaks @alexislefebvre

alexislefebvre:
You're welcome.

I was confused with `stable` and `old-stable`. But that's simple: both are stable versions. Thanks for the last edit.

PR #1153, opened 2020-08-19
Fix markdown codeblock formatting

Arnaud-J:
This simply adds php formatting to a codeblock that was missing it


soyuka:
thansk @Arnaud-J

PR #1155, opened 2020-08-23
Fix resources.yaml example

markwatney2016:


soyuka:
thanks @markwatney2016

PR #1156, opened 2020-08-25
Remove note about ObjectIds as they should work.

soyuka:
Remove note about MongoDB ObjectIds as they should work.


dunglas:
Thanks!

PR #1157, opened 2020-08-27
Fix webonyx/graphql version

Tithugues:
For a few weeks, a new stable version of webonyx/graphql with BC was released. API Platform should release on previous version until being updated.


alanpoulain (changes requested):
A release should be done by @dunglas today.

alanpoulain:
Released :slightly_smiling_face:

PR #1169, opened 2020-09-25
Change text/csv to application/csv

GrosTon1:
text/csv isn't supported. This PR changes that in content negotiation doc to application/csv, as suggested in the error.

https://nextcloud.clauzier.dev/s/LFCrDSg5BZ5K7mq


antograssiot:
That's likely due to a mistake in your config as csv is not enabled by default, but text/csv is correct and defined in the RFC 4180 linked just above so I don't think that this should be updated.

dunglas:
Closing for the reasons explained by @antograssiot. Thanks for contributing anyway!

PR #1170, opened 2020-10-01
Corrected PHP version number

toby-griffiths:
Bumped PHP version to 7.4, as per the container in v2.5.7 I've just downloaded.


dunglas:
Good catch, thanks.

toby-griffiths:
Happy to help… and an easy started for 10 for Hacktoberfest.

PR #1171, opened 2020-10-02
Missing ApiProperty uses in identifiers doc

Caligone:
Hi there
First time here
This PR fixes two missing ApiProperty uses in the identifiers documentation.
I got caught by this during testing the examples few hours ago.
Hope it helps


dunglas:
Good catch! Thanks.

PR #1174, opened 2020-10-03
Added POPO acronym where first explained

toby-griffiths:
This should help people seeing POPO further down the docs to work out what it means, to reduce their cognitive load.


dunglas:
Thanks @toby-griffiths

PR #1175, opened 2020-10-05
Fix: GetResponseEvent deprecation

Zowac:
GetResponseEvent is deprecated since Symfony 4.3, use RequestEvent instead


dunglas:
Thanks!

PR #1177, opened 2020-10-06
Wrong ContextAwareDataPersisterInterface use

Caligone:
Can be seen here: https://api-platform.com/docs/core/data-persisters/#decorating-the-built-in-data-persisters


dunglas:
Thanks!!

PR #1179, opened 2020-10-08
Fix : GetResponseForControllerResultEvent depreciation

Zowac:
- GetResponseForControllerResultEvent is deprecated since Symfony 4.3, use ViewEvent instead


soyuka:
thanks @Zowac

PR #1183, opened 2020-10-14
Update serialization.md

brunonic:
Missing argument for this use case.


dunglas:
This argument will be autowired if you keep the default configuration.

brunonic:
Ok my bad

PR #1184, opened 2020-10-15
[Admin] Manage correctly the token not present in local storage

alanpoulain:


benfavre:
If anything the fetchHeaders  should be a callback and not a constant

```php
const fetchHeaders = () => localStorage.getItem(config.tokenStorageKey) ?
    { Authorization: `Bearer ${localStorage.getItem(config.tokenStorageKey)}`} :
    {};
````

Also even after making this change I still see dispatched calls to API's without my token. This "fix" does not solve the underlying issue.

PR #1188, opened 2020-10-20
Update Swagger documentation to use V3

MartinMarx:
Docs are currently showing examples using both Swagger V2 and V3 (swagger_context vs openapi_context). Changing examples to V3 by default (which is the version set by default) would make the documentation more consistant and avoid version mistakes for new users.


dunglas:
Good idea, thanks @MartinMarx

PR #1190, opened 2020-10-21
fetchHeaders is var not function?

pamuche:
If I get it right, ```fetchHeaders``` isn't a function but the foredefined variable..


alanpoulain:
Yes sorry the right fix is this: https://github.com/api-platform/docs/pull/1191.

PR #1193, opened 2020-10-22
distribution: added notes about the GitHub template

COil:
Context: I've used the GitHub template to start testing APIP. When I arrived at the "admin" section, I realized that this component was removed because of this commit that was done recently: https://github.com/api-platform/api-platform/commit/d64d2fd27e953363e9eeb086db4a1e16a347bee8  

So it's better to warn the user : "use the GitHub template at your own risk", using the download of the last stable tag is better.


dunglas:
We shouldn't do that actually! Because it's a template repository, the `master` branch of `api-platform/api-platform` should be in sync with the with the stable version of `core` and the docs. Breaking changes should be in a `next` branch.

I'm in favor of closing this PR to not confuse users (even if the current situation is indeed already confusing, we'll fix it).

PR #1196, opened 2020-10-26
data provider: added missing context

COil:
Synchronized with the `BlogPostItemDataProvider` example to show that the context is also available in the `getCollection()` function.

### DIFF
--- a/core/data-providers.md
+++ b/core/data-providers.md
@@ -38,19 +38,19 @@ If no data is available, you should return an empty array.
 
 namespace App\DataProvider;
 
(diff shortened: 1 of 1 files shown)


core/data-providers.md
```diff
@@ -50,7 +50,7 @@ final class BlogPostCollectionDataProvider implements CollectionDataProviderInte
         return BlogPost::class === $resourceClass;
     }
 
-    public function getCollection(string $resourceClass, string $operationName = null): \Generator
```
dunglas, line 53: `iterable` ?

dunglas (approved):

PR #1197, opened 2020-10-27
Data Provider Exception vs Restricted interface

Caligone:
The usage of `ApiPlatform\Core\Exception\ResourceClassNotSupportedException` 
along the implementation of the `ApiPlatform\Core\DataProvider\RestrictedDataProviderInterface` in the Data Provider seems redundant.


dunglas:
Thanks @Caligone!

PR #1198, opened 2020-10-27
Missing ArrayCollection use in relations validation

Caligone:
In the relations validation, there is a ArrayCollection usage without its use.

I'm not sure wether or not this kind of fix are useful, but I regularly have to look for the path of a class in API Platform documentation example so I create quick PRs.

Feel free to decline and/or stop me

### DIFF
--- a/core/validation.md
+++ b/core/validation.md
@@ -467,6 +467,9 @@ For example:
 ```
 
 ```php
+
+use Doctrine\Common\Collections\ArrayCollection;
+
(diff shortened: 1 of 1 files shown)


dunglas:
Hi @Caligone, thanks for this patch and sorry for our lack of reactivity on this. It has been fixed in another PR in parallel so I'm closing, but it's definitely the kind of changes that are helpful.

PR #1199, opened 2020-10-27
Added an advice in the push-relations.md page

Romain:
Symfony HTTP Client is required to make this work.

### DIFF
--- a/core/push-relations.md
+++ b/core/push-relations.md
@@ -35,3 +35,5 @@ According to the [Preload W3C Candidate Recommendation](https://www.w3.org/TR/pr
 
 Using this feature maximises HTTP cache hits for your API resources.
 For best performance, this feature should be used in conjunction with [the built-in HTTP cache invalidation system (based on Varnish)](performance.md#enabling-the-built-in-http-cache-invalidation-system).
+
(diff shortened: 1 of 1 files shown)


dunglas:
This isn't the case. Any client supporting HTTP/2 Server Push (and especially browsers) will support this feature.

Romain:
Hum, interesting...
When I set it up, I got the following error in my logs: `request.CRITICAL: Uncaught PHP Exception Symfony\Component\ErrorHandler\Error\ClassNotFoundError: "Attempted to load trait "HttpClientTrait" from namespace "Symfony\Component\HttpClient"`.

Searching for a solution, I found [this comment](https://github.com/api-platform/api-platform/issues/1437#issuecomment-592522041) which solved the issue, hence my PR.

dunglas:
This error is probably not related to this feature. Are you using the test client? The test clients needs HttpClient.

Romain:
I'm not using it, but perhaps is it installed by default?
The weird thing is that this error appeared exactly when I set up the push relations.
How can I check if the test client is used? What's the name of the package?

dunglas:
Closing because I'm sure this component isn't needed to use this feature. It's probably a side effect related to a code segment triggered by the push. If the problem persist, please fill an issue on `api-platform/core`.

PR #1201, opened 2020-10-30
Update the way to add login with jwt token return in swagger. Previous did not work with 2.5

vibby:
Working documentation, for adding custom routes in swagger API.

### DIFF
--- a/core/jwt.md
+++ b/core/jwt.md
@@ -223,7 +223,21 @@ final class SwaggerDecorator implements NormalizerInterface
     {
         $docs = $this->decorated->normalize($object, $format, $context);
 
-        $docs['components']['schemas']['Token'] = [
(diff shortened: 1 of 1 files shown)


alanpoulain:
What was not working before?

vibby:
With the proposed code :
image
(note the emptiness in parameters area)

With my changes :
image
And the Model associated
image

alanpoulain:
I think you are using a recent version of Swagger. Are you on master maybe?
This is what I have:
image

vibby:
Hum, I think I did not set the version of OpenAPI anywhere. I just use api-platform/core on version 2.5.7 out of the box …

soyuka:
Mhh indeed this needs a fix but we should decorate the OpenAPI Factory as stated https://github.com/api-platform/docs/blob/master/core/openapi.md

PR #1203, opened 2020-11-02
Index name should be lower snake case

darthf1:
Don't know if this is evident, but it had me boggling for a time. Index name should be lower snake case, not just lower case.


dunglas:
Thanks @darthf1

PR #1204, opened 2020-11-02
Properties should be camel case

darthf1:
In the docs its really explicit that ElasticSearch should be populated with snake_case properties, but that gave me issues with normalizing nested objects. When I populated ElasticSearch with camelCase properties, serialization was Ok.


soyuka:
I stumbled upon this as well. Thanks for the edit.

meyerbaptiste:
This seems strange to me. Do you have an example? It looks more like a bug to me.
IIRC, the best practice is snake case for ES and if so, we should follow it.

PR #1207, opened 2020-11-10
demo: links to the related demo code

COil:
This update is related to the following PR on APIP demo application: https://github.com/api-platform/demo/pull/162

The idea is to add cross-links between docs and the demo so it's easier for people to navigate and find useful stuff.


alanpoulain (approved):
Great!

alanpoulain:
Thank you @COil.

PR #1210, opened 2020-11-12
Operations page - update YAML configuration

abdounikarim:
Hello, i just updated the yaml configuration. 😉


dunglas:
Good catch!

PR #1214, opened 2020-11-19
Operations page - add yaml and xml code block

abdounikarim:
Hello,

I just completed the code block with `yaml` and `xml` examples.


dunglas:
Thanks for these additions @abdounikarim, it's very useful.

PR #1215, opened 2020-11-19
Operations page - add yaml and xml code block

abdounikarim:
Hello,

I just add `yaml` and `xml` examples.


dunglas:
Awesome, thanks.

PR #1216, opened 2020-11-19
Operations page - add missing route name

abdounikarim:
Hello,

I just add missing route name.


alexislefebvre:
Isn't the YAML key already the name of the route? It looks like it repeat it.

dunglas:
@alexislefebvre is right (https://symfony.com/doc/current/routing.html#creating-routes-in-yaml-xml-or-php-files). Is right, I'm closing. Thanks for contributing!

PR #1240, opened 2021-01-06
Update JWT Documentation with the OpenApi way

garciasdos:
Update with the new OpenApi documentation way. It works with Api Platform 2.6.0 beta 1


dunglas:
Thank you very much @garciasdos!

PR #1247, opened 2021-01-18
Feat: update documentation for Exception

BoutValentin:
Update the Error and Exception documentation to be more precise about flatten Exception and introduce simplified decorators


BoutValentin:
Closing and replace by #1248 to base in master

PR #1252, opened 2021-01-21
Use "int" type instead of "integer" deprecated type

samnela:
"integer" type was deprecated in doctrine/mongodb-odm 2.1 and will be removed in 3.0. Use "int" instead.


alanpoulain (approved):

alanpoulain:
Thank you!

PR #1270, opened 2021-02-03
Change api platform documentation url

walva:
When you create a new api-platform project using `composer create-project symfony/skeleton bookshop-api` then `composer require api`, the doc is available through **https://localhost/api/** and not **https://localhost/docs/**.


alanpoulain:
Yes, that's wanted. We prefer to write the documentation in the distribution point of view.
There is already a note in https://github.com/api-platform/docs/blob/2.6/distribution/index.md saying that the endpoint will be different:
> Note: when installing API Platform this way, the API will be exposed as the /api/ path. You need to open http://localhost:8000/api/ to see the API documentation. If you are deploying API Platform directly on an Apache or NGINX webserver and getting a 404 error on opening this link, you will need to enable the rewriting rules for your specific webserver software.

walva:
Thanks I didn't notice this paragraph. Thanks for the review

Issue #1272, opened 2021-02-04
[2.6] JWT documentation broken


NicoHaase:
https://api-platform.com/docs/core/jwt/#adding-endpoint-to-swaggerui-to-retrieve-a-jwt-token contains code that only works on PHP 8. The `Model/PathItem` part contains named arguments, changing this to code compatible with PHP 7 would be great :) 

I have to figure out how to make it work before I can suggest an alternative

alanpoulain:
Hello,
This is wanted. Maintaining a documentation with a support of old PHP versions is too much work, we prefer to consider you are using the latest stable version of PHP.

NicoHaase:
Don't get me wrong, I can only imagine the nightmare of supporting versions that differ so much, but couldn't you make `api-platform/core` v2.6 also dependent on PHP 8 only if you don't want to provide documentation about how to get it work on PHP 7.2, which is the currently lowest version supported?

dunglas:
The API Platform documentation assumes that you use the latest version of the API Platform distribution, which is shipped with PHP 8. For clarity and to promote best practices, we always use in the docs the versions of our dependencies (PHP, Symfony...) shipped with the distribution.

At code level, API Platform support ms older versions of its dependencies, but it's mostly to ease the upgrade process. This allows upgrading package by package (start by upgrading PHP, then Symfony, then API Platform). However, we recommend to always upgrade your dependencies before starting to work on new features.

We will **not** document how to use the latest version of API Platform with old versions of its dependencies. This would require too much work, be a maintenance nightmare and would have almost no benefits.

TLDR: using the latest version of API Platform with non-latest dependencies is supported (and automatically tested) but we always assume you're using the latest versions in the documentation. If you want to mix old dependencies with new versions of API Platform, you're on your own (but it's usually easy to adapt the documentation for older versions).

Supporting the latest versions is already very tedious. We don't have the resource to document all possible dependency mixes and I don't think that it's a good idea anyway.

PR #1284, opened 2021-02-14
Update "Changing Operations [...]" section

yivi:
[This section](https://api-platform.com/docs/core/openapi/#changing-operations-in-the-openapi-documentation) uses a configuration format that works for version 2 of OpenApi, which clashes with the default for the distribution (v3). The documentation example simply does not work out of the box.

Users updating are also being [hit by this](https://github.com/api-platform/core/issues/4035).

Included an example with attributes, instead of annotations, since [we should always use the latest version of dependencies on documentation](https://github.com/api-platform/docs/issues/1272#issuecomment-773178233).


yivi:
The new `swagger-ui-2.png` is 100kb, because it's a bit taller to show the "example value" section. It's a bit heavy. A WebP version would be only 20kb, but since all other images are PNG I didn't want to add an image in a different version.

core/openapi.md
```diff
@@ -311,58 +311,93 @@ You also have full control over both built-in and custom operations documentatio
...
+                example:
+                  name: Mr. Rabbit
+                  description: Pink rabbit
+
 ```
 
 ```xml
 <?xml version="1.0" encoding="UTF-8" ?>
-<resources xmlns="https://api-platform.com/schema/metadata"
+<resources
```
alanpoulain, line 384: Why removing the `xmlns`?

core/openapi.md
```diff
@@ -374,17 +409,14 @@ resources:
...
-                                <attribute name="description">Pink rabbit</attribute>
-                            </attribute>
                         </attribute>
                     </attribute>
                 </attribute>
             </collectionOperation>
         </collectionOperations>
     </resource>
 </resources>
+
```
alanpoulain, line 419:
```suggestion
</resources>
```

core/openapi.md
```diff
@@ -311,58 +311,93 @@ You also have full control over both built-in and custom operations documentatio
 
 [codeSelector]
 
+```php
```
alanpoulain, line 314:
```suggestion
```php
<?php
// api/src/Entity/Rabbit.php

```

alanpoulain:
Hello, thank you for this PR :slightly_smiling_face:

core/openapi.md
yivi: Likely I got it wrong in copy and paste. In my own projects, it's the only thing I have: `<resources xmlns="https://api-platform.com/schema/metadata">`. But trying to match the docs fuller style it seems I messed it up. I'll correct it in a second.

core/openapi.md
```diff
@@ -374,17 +411,14 @@ resources:
...
-                                <attribute name="description">Pink rabbit</attribute>
-                            </attribute>
                         </attribute>
                     </attribute>
                 </attribute>
             </collectionOperation>
         </collectionOperations>
     </resource>
 </resources>
+
```
alanpoulain, line 421:
```suggestion
    </resource>
</resources>
```

alanpoulain (approved):

alanpoulain:
Thank you @yivi!

PR #1286, opened 2021-02-15
Need readableLink: true for embed objects.

### LINKED ISSUE #1236: No mention of readableLink is required for embedded
Embedded of objects must have `readableLink: true` in order to render the full object instead of iri, however it is never mentioned in the doc. May I know is this something we need to fix or am I just missing something?

https://api-platform.com/docs/core/serialization/#embedding-relations

kayue:
Embedded of objects must have readableLink: true in order to render the full object instead of iri, however it is never mentioned in the doc.

Fix #1236.

### DIFF
--- a/core/serialization.md
+++ b/core/serialization.md
@@ -231,8 +231,9 @@ In the following JSON document, the relation from a book to an author is by defa
 ```
 
 It is possible to embed related objects (in their entirety, or only some of their properties) directly in the parent
-response through the use of serialization groups. By using the following serialization groups annotations (`@Groups`),
-a JSON representation of the author is embedded in the book response:
+response through the use of serialization groups. By using the serialization groups annotations (`@Groups`) and the 
+`readableLink` attribute of the `ApiPlatform\Core\Annotation\ApiProperty` annotation, a JSON representation of the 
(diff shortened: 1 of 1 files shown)


alanpoulain:
It shoudn't be necessary though (at least for the normal case). Maybe it's a bug. Is it in a Sylius context?

kayue:
I am not sure in what condition it becomes optional. It looks like it is required in normalizer.

In HAL Normalizer:

https://github.com/api-platform/core/blob/0d784fd1da03b1fe4def1222e928f73630110a3c/src/Hal/Serializer/ItemNormalizer.php#L157-L161

In `AbstractItemNormalizer`

https://github.com/api-platform/core/blob/0d784fd1da03b1fe4def1222e928f73630110a3c/src/Serializer/AbstractItemNormalizer.php#L654-L679

alanpoulain:
The `readableLink` attribute in the property metadata is set here: https://github.com/api-platform/core/blob/0d784fd1da03b1fe4def1222e928f73630110a3c/src/Metadata/Property/Factory/SerializerPropertyMetadataFactory.php#L132.
Could you check why it isn't `true` in your case?

PR #1287, opened 2021-02-16
Add global resources defaults entry

alanpoulain:


core/configuration.md
```diff
@@ -90,40 +91,47 @@ api_platform:
         order_parameter_name: 'order'
 
         pagination:
```
dunglas, line 92: We should deprecate this whole section, isn't it?

core/configuration.md
```diff
@@ -218,40 +226,44 @@ api_platform:
             url:       
         
     http_cache:
```
dunglas, line 220: Maybe this one too?

core/configuration.md
```diff
@@ -218,40 +226,44 @@ api_platform:
...
-
     mercure:
-        # Enabled by default with installed symfony/mercure-bundle
+        # Enabled by default with installed symfony/mercure-bundle.
         enabled: false
 
         # The URL sent in the Link HTTP header. If not set, will default to MercureBundle's default hub URL.
         hub_url: null
 
     messenger:
```
dunglas, line 253: And thus one...

core/configuration.md
alanpoulain: There is no resource-wide equivalent for some global parameters (like `partial_parameter_name`).

core/configuration.md
alanpoulain: Same answer.

core/configuration.md
alanpoulain: Not sure if we want to deprecate this one.
Using it will remove all the messenger services whereas it's not the case for the `defaults` one. But we can change it.
As you wish.

core/configuration.md
dunglas: But shouldn't we uniformize and add them?

alanpoulain: Does it make sense? I'm not sure having a resource-wide `partial_parameter_name` is a good idea for example. Having a parameter name depending of the resource is very error prone.
Anyway the choice has been made in this PR: https://github.com/api-platform/core/pull/3151.
If we want to change it it will be for the `2.7` version isn't it?

dunglas: Yes

PR #1293, opened 2021-02-20
Complete example for Messenger with persistence

andreystavitsky:


soyuka:
thanks @andreystavitsky

PR #1323, opened 2021-03-25
Update debugging.md

brettins:
Xdebug 3 no longer uses the remote_ convention, they have been removed in Xdebug 3. 
Need XDEBUG_MODE: debug for step debugging

Added sample launch.json for .vscode

These changes are all for Ubuntu 20.04, so maybe the different configs could be separated out?

### DIFF
--- a/distribution/debugging.md
+++ b/distribution/debugging.md
@@ -70,7 +70,7 @@ Note for Mac environments use the following:
         remote_handler=dbgp
 ```
 
-In VSCode, alongside the default php config in `launch.json`, you'll need path mappings for a docker image.
+In VSCode, alongside the default PHP configuration in `launch.json`, you'll need path mappings for the Docker image.
 ```json
 {
     "version": "0.2.0",
--- a/distribution/debugging.md
+++ b/distribution/debugging.md
@@ -46,8 +46,10 @@ services:
       XDEBUG_MODE: debug
       XDEBUG_CONFIG: >-
         client_host=host.docker.internal
-        client_port=9003 # default port for Xdebug 3
-        idekey=PHPSTORM # or VSCODE if you are debugging with VSCode
+        # default port for Xdebug 3
+        client_port=9003 
+        # or VSCODE if you are debugging with VSCode
+        idekey=PHPSTORM 
       # This should correspond to the server declared in PHPStorm `Preferences | Languages & Frameworks | PHP | Servers`
       # Then PHPStorm will use the corresponding path mappings
       PHP_IDE_CONFIG: serverName=api-platform
@@ -100,9 +102,3 @@ PHP …
     with Xdebug v3.0.2, Copyright (c) 2002-2021, by Derick Rethans
     …
 ```
-
-
-
-
-
-
--- a/distribution/debugging.md
+++ b/distribution/debugging.md
@@ -87,7 +87,8 @@ In VSCode, alongside the default PHP configuration in `launch.json`, you'll need
     ]
 }
 ```
-Note for Linux environments, the `client_host` setting of `host.docker.internal` will not work, you will need the actual local IP address of your computer.
+
+Note: For Linux environments, the `client_host` setting of `host.docker.internal` will not work, you will need the actual local IP address of your computer.
 
 ## Troubleshooting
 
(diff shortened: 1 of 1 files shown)


distribution/debugging.md
```diff
@@ -81,3 +100,9 @@ PHP …
     with Xdebug v3.0.2, Copyright (c) 2002-2021, by Derick Rethans
     …
 ```
+
```
dunglas, line 103: These extra blank lines must be removed.

dunglas (changes requested):

distribution/debugging.md
```diff
@@ -61,13 +63,33 @@ Note for Mac environments use the following:
...
         remote_connect_back=0
-        remote_port=9000
+        remote_port=9003
         idekey=PHPSTORM
         remote_autostart=1
         remote_mode=req
         remote_handler=dbgp
 ```
 
+In VSCode, alongside the default php config in `launch.json`, you'll need path mappings for a docker image.
```
alanpoulain, line 73: I don't think it's a good idea to add IDE specific related configuration.

brettins: I think for 99% of the docs that is definitely the case, but in the case of debugging, particularly step debugging, it is inherently tied to an IDE. Before my modifications there were already specifics things to php storm in this doc, and there are mentions of php storm and vs code in the recommended IDEs for API Platform. Docker is also recommended, and this path mapping is a quirk of docker and API platform together. 

If the intent is to be IDE agnostic, then there likely shouldn't be recommendations to use php storm and vs code elsewhere.

For me, knowing this config info would have saved me a few hours, and this doesn't clutter up the docs much.

dunglas: I'm in favor of having docs for VSCode and PHPStorm as it's the two most popular tools in the PHP world.

alanpoulain: Ok then :smile:

distribution/debugging.md
```diff
@@ -43,12 +43,14 @@ services:
...
+      # default port for Xdebug 3 is 9003
+      # idekey=VSCODE if you are debugging with VSCode
       XDEBUG_CONFIG: >-
-        remote_enable=1
-        remote_host=host.docker.internal
-        remote_connect_back=1
-        remote_port=9000
-        idekey=PHPSTORM
+        client_host=host.docker.internal
+        client_port=9003
```
alanpoulain, line 52: I don't think this line is necessary. This is the default port for XDebug 3.

distribution/debugging.md
alanpoulain, line 73:
```suggestion
In VSCode, alongside the default PHP configuration in `launch.json`, you'll need path mappings for the Docker image.
```

distribution/debugging.md
```diff
@@ -61,13 +63,33 @@ Note for Mac environments use the following:
...
         remote_handler=dbgp
 ```
 
+In VSCode, alongside the default php config in `launch.json`, you'll need path mappings for a docker image.
+```json
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Listen for Xdebug",
```
alanpoulain, line 79:
```suggestion
            "name": "Listen for XDebug",
```

distribution/debugging.md
```diff
@@ -61,13 +63,33 @@ Note for Mac environments use the following:
...
+            "port": 9003,
+            "log": true,
+            "pathMappings": {
+                "/srv/api": "${workspaceFolder}/api"
+            },
+        },
+    ]
+}
+```
+Note for Linux environments, the `client_host` setting of `host.docker.internal` will not work, you will need the actual local IP address of your computer.
```
alanpoulain, line 91:
```suggestion

Note: For Linux environments, the `client_host` setting of `host.docker.internal` will not work, you will need the actual local IP address of your computer.
```

distribution/debugging.md
brettins: I'm happy to remove it, but the line I'm replacing `remote_port=9000` was also an unnecessary line, as the default port for Xdebug 2 was 9000 - I'm not sure if the intend was to just be informative?

alanpoulain: I don't know but I think it's safe to remove it.

alanpoulain (approved):

alanpoulain:
Thank you @brettins!

PR #1326, opened 2021-03-26
Fix #2686: deprecate allow_plain_identifiers option

vincentchalamon:


core/serialization.md
```diff
@@ -408,6 +408,59 @@ class Person
...
+### Plain identifiers
+
+Instead of sending an iri to set a relation, you may want to send a plain identifier. To do so, you must create your own denormalizer:
+
+```yaml
+# api/config/services.yaml
+services:
+    'App\Serializer\PlainIdentifierNormalizer':
+        # By default .inner is passed as argument
+        decorates: 'api_platform.jsonld.normalizer.item'
```
soyuka, line 420: Are you sure that's enough? To me it's more like:

```
decorates: `api_platform.jsonld.normalizer.item`
arguments: ['App\Serializer\PlainIdentifierNormalizer.inner']
```

core/serialization.md
```diff
@@ -408,6 +408,59 @@ class Person
...
+
+    public function supportsDenormalization($data, $type, $format = null)
+    {
+        return $this->decorated->supportsDenormalization($data, $type, $format);
+    }
+
+    public function denormalize($data, $class, $format = null, array $context = [])
+    {
+        if (null !== $data['relation']) {
+            $relation = $this->iriConverter->getItemIriFromResourceClass(Relation::class, [$data['relation']]);
```
soyuka, line 453: $relation = $this->iriConverter->getItemIriFromResourceClass(Relation::class, ['id' => $data['relation']]);

core/serialization.md
vincentchalamon: The Symfony autowiring will detect the first argument as the same interface as the decorated one, and will pass the .inner as argument by default.

https://symfony.com/doc/current/service_container/service_decoration.html
> If you’re using the default services.yaml configuration, the decorated service is automatically injected when the constructor of the decorating service has one argument type-hinted with the decorated service class.

core/serialization.md
```diff
@@ -408,6 +408,59 @@ class Person
 
 ```
 
+### Plain identifiers
```
alanpoulain, line 411:
```suggestion
### Plain Identifiers
```

core/serialization.md
```diff
@@ -408,6 +408,59 @@ class Person
...
+
+Instead of sending an iri to set a relation, you may want to send a plain identifier. To do so, you must create your own denormalizer:
```
alanpoulain, line 413:
```suggestion
Instead of sending an IRI to set a relation, you may want to send a plain identifier. To do so, you must create your own denormalizer:
```

core/serialization.md
```diff
@@ -408,6 +408,59 @@ class Person
...
+
+```yaml
+# api/config/services.yaml
+services:
+    'App\Serializer\PlainIdentifierNormalizer':
```
alanpoulain, line 418:
```suggestion
    App\Serializer\PlainIdentifierNormalizer:
```

vincentchalamon: Must be fixed in all other parts of the doc too

core/serialization.md
```diff
@@ -408,6 +408,59 @@ class Person
...
+namespace App\Serializer;
+
+use ApiPlatform\Core\Api\IriConverterInterface;
+use ApiPlatform\Core\Exception\ItemNotFoundException;
+use App\Entity\Relation;
+use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
+
+final class PlainIdentifierNormalizer implements DenormalizerInterface
+{
+    private $decorated;
```
alanpoulain, line 436:
```suggestion
    private DenormalizerInterface $decorated;
```

core/serialization.md
```diff
@@ -408,6 +408,59 @@ class Person
...
+    private $iriConverter;
```
alanpoulain, line 437:
```suggestion
    private IriConverterInterface $iriConverter;
```

core/serialization.md
```diff
@@ -408,6 +408,59 @@ class Person
...
+
+    public function __construct(DenormalizerInterface $decorated, IriConverterInterface $iriConverter)
+    {
+        $this->decorated = $decorated;
+        $this->iriConverter = $iriConverter;
+    }
+
+    public function supportsDenormalization($data, $type, $format = null)
```
alanpoulain, line 445:
```suggestion
    public function supportsDenormalization($data, $type, $format = null): bool
```

core/serialization.md
vincentchalamon: @alanpoulain I used the same syntax as every service example in the doc

alanpoulain: It's as you prefer! If it was in `2.6` I would have suggest you to change all the examples but since it's on `main`, we shouldn't change the other examples.

core/bootstrap.md
```diff
@@ -444,7 +443,7 @@ $hydraConstraintViolationNormalizer = new HydraConstraintViolationListNormalizer
 
 $problemErrorNormalizer = new ErrorNormalizer($debug, $defaultContext);
 
-$itemNormalizer = new ItemNormalizer($propertyNameCollectionFactory, $propertyMetadataFactory, $iriConverter, $resourceClassResolver, $propertyAccessor, $nameConverter, $classMetadataFactory, $dataProvider, $allowPlainIdentifiers, $logger, $dataTransformers, $resourceMetadataFactory, /** resourceAccessChecker **/ null);
+$itemNormalizer = new ItemNormalizer($propertyNameCollectionFactory, $propertyMetadataFactory, $iriConverter, $resourceClassResolver, $propertyAccessor, $nameConverter, $classMetadataFactory, $dataProvider, $logger, $dataTransformers, $resourceMetadataFactory, /** resourceAccessChecker **/ null);
```
guilliamxavier, line 446: That's removing, not just deprecating, right?

vincentchalamon: You're right, I should keep it for 2.7 and add a deprecation and a TODO to remove it in 3.0

PR #1331, opened 2021-03-31
Updated Mac env setting names for Xdebug v3

rettal:
I have updated the setting names to match Xdebug v3 setting names for Mac environments.

### DIFF
--- a/distribution/debugging.md
+++ b/distribution/debugging.md
@@ -59,13 +59,8 @@ Note for Mac environments use the following:
 
 ```yml
       XDEBUG_CONFIG: >-
-        client_port=9003
         idekey=PHPSTORM
-        remote_handler=dbgp
-        discover_client_host=0
         client_host=docker.for.mac.localhost
-        mode=debug
-        start_with_request=trigger
 ```
 
 In VSCode, alongside the default PHP configuration in `launch.json`, you'll need path mappings for the Docker image.
--- a/distribution/debugging.md
+++ b/distribution/debugging.md
@@ -59,14 +59,13 @@ Note for Mac environments use the following:
 
 ```yml
       XDEBUG_CONFIG: >-
-        remote_enable=1
-        remote_host=docker.for.mac.localhost
-        remote_connect_back=0
-        remote_port=9003
+        client_port=9003
         idekey=PHPSTORM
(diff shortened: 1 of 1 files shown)


rettal:
Ref: https://xdebug.org/docs/upgrade_guide

distribution/debugging.md
```diff
@@ -59,14 +59,13 @@ Note for Mac environments use the following:
...
-        remote_connect_back=0
-        remote_port=9003
+        client_port=9003
         idekey=PHPSTORM
-        remote_autostart=1
-        remote_mode=req
         remote_handler=dbgp
+        discover_client_host=0
+        client_host=docker.for.mac.localhost
+        mode=debug
```
alanpoulain, line 67: The `mode` is not necessary since there is already `XDEBUG_MODE`.

distribution/debugging.md
```diff
@@ -59,14 +59,13 @@ Note for Mac environments use the following:
 
 ```yml
       XDEBUG_CONFIG: >-
-        remote_enable=1
-        remote_host=docker.for.mac.localhost
-        remote_connect_back=0
-        remote_port=9003
+        client_port=9003
```
alanpoulain, line 62: Since it's the default port, it's not necessary, you can remove it.

distribution/debugging.md
```diff
@@ -59,14 +59,13 @@ Note for Mac environments use the following:
...
         idekey=PHPSTORM
-        remote_autostart=1
-        remote_mode=req
         remote_handler=dbgp
+        discover_client_host=0
+        client_host=docker.for.mac.localhost
+        mode=debug
+        start_with_request=trigger
```
alanpoulain, line 68: This is the default in `debug` mode, it should be removed as well.

rettal:
Hi @alanpoulain, Thank you for reviewing my change. I took the decision to just replace - like for like - the setting names that were already listed on the docs page - even though some are default but I agree with your comments. I will make the changes.

rettal:
Hi @alanpoulain, I have made the changes (and tested locally on my machine) and works as expected, and docs look much simpler now. Thank you.

alanpoulain:
Thank you @rettal.

PR #1334, opened 2021-04-02
Add information for app_swagger_ui on routes

ericovasconcelos:
This just add to documentation the information that he needs to add the route to Swagger UI on project configuration if need to use a custom url for that.


alanpoulain:
See https://github.com/api-platform/docs/pull/1604.

Issue #1336, opened 2021-04-03
the expected error on the missing book title does not appear


CedCannes:
Hi, 
I followed the how to and I have an unexpected result
in  docs/distribution/index.md 
`L629 Oops, we forgot to add the title. Submit the request anyway, you should get a 500 error with the following message:`

This error does not appear because the title field is initialized to '' and not to null.

```
/** The title of this book. */
    public string $title = '';
```

In addition the documentation proposes to add a validator to prevent the error from happening again but doesn't add one to the title field mentioned above but expects the following exception to appear:
```
{
      "propertyPath": "title",
      "message": "This value should not be blank."
    }
```

PR #1337, opened 2021-04-04
For the how to scenario to run as expected.

CedCannes:
PR for https://github.com/api-platform/docs/issues/1336

### DIFF
--- a/distribution/index.md
+++ b/distribution/index.md
@@ -260,7 +260,7 @@ class Book
     public ?string $isbn = null;
 
     /** The title of this book. */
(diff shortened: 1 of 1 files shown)


dunglas:
Thanks for fixing the docs, however I think we should try another approach. Now that it's possible (it wasn't when I have first written this entry) having non-nullable properties is a better practice, especially because it "naturally" prevents this class of errors.
I would prefer to change the wording of the docs (removing the part about the 500 error) instead of making the filed nullable just to be able to trigger an error.

PR #1345, opened 2021-04-10
Update file-upload.md

brettins:
System returns a "normalizer is null" error without including this interface implementation.


alanpoulain (approved):
Indeed, thanks.

PR #1349, opened 2021-04-14
Add notes on usage of `ApiProperty` `security_post_denormalize`.

DavidBennettUK:
Documents https://github.com/api-platform/core/pull/4184

PR #1351, opened 2021-04-23
Update security ApiProperty variables

xavren:


alanpoulain:
Thanks but I think it's better to favor this one: https://github.com/api-platform/docs/pull/1349.

xavren:
Hey @alanpoulain, 

`request` is actually unavailable from variables in this context
`object` is not the current resource during denormalization but the resourceClass

The PR #1349 actually is adding some informations about previous object that is nice but different from this PR, merging this contributions would be nice

alanpoulain:
I didn't see the part about the `request`, sorry.
For the `object`, it will be `null` instead of the resource class when https://github.com/api-platform/core/pull/4184 is merged. But it will be on `main`, so I guess we can merge your PR in the meantime.
I have been too quick, sorry!

alanpoulain:
Is `user` available?

xavren:
Yes, user is there too

PR #1357, opened 2021-05-07
Fixes to Content Negotiation / Serialization docs

jamesisaac:
- Docs state that only JSON-LD is enabled by default, which isn't the case: https://github.com/api-platform/core/blob/de5988dacc871a7c1565aa8dea8c30102984cff9/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php#L216...L218
  - While JSON may only enabled for the sake of Swagger, it still means that someone can use content negotiation anywhere on the API to request data in an unexpected format... perhaps bypassing a custom normalizer.  Important for security considerations developers are aware that more formats exist.
  - (HTML is enabled by default too, but from what I see this doesn't provide a way to actually interact with the main API - so probably doesn't need to be mentioned?)
- Added the missing `myformat` example which is talked about but not shown.
- In the Serialization section, the approach recommended for adding a custom serializer to multiple content types won't work due to duplicate YAML key.

### DIFF
--- a/core/serialization.md
+++ b/core/serialization.md
@@ -694,7 +694,9 @@ services:
 Note: this normalizer will work only for JSON-LD format, if you want to process JSON data too, you have to decorate another service:
 
 ```yaml
-    App\Serializer\ApiNormalizer:
+    # Need a different name to avoid duplicate YAML key
+    'app.serializer.normalizer.item.json':
+        class: 'App\Serializer\ApiNormalizer'
         decorates: 'api_platform.serializer.normalizer.item'
 ```
 
--- a/core/content-negotiation.md
+++ b/core/content-negotiation.md
@@ -2,7 +2,7 @@
 
 The API system has built-in [content negotiation](https://en.wikipedia.org/wiki/Content_negotiation) capabilities.
 
-By default, only the [JSON-LD](https://json-ld.org) format is enabled. However API Platform Core supports many more formats and can be extended.
+By default, only the [JSON-LD](https://json-ld.org) and JSON formats are enabled. However API Platform Core supports many more formats and can be extended.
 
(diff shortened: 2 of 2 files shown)


core/serialization.md
```diff
@@ -694,7 +694,9 @@ services:
 Note: this normalizer will work only for JSON-LD format, if you want to process JSON data too, you have to decorate another service:
 
 ```yaml
-    App\Serializer\ApiNormalizer:
+    # Need a different name to avoid duplicate YAML key
+    'App\Serializer\ApiNormalizer.json':
```
alanpoulain, line 698: Shouldn't we change it to: `app.serializer.normalizer.item.json` (and the previous one too)?

jamesisaac: I'm not sure what best practice is in this scenario.  Most Symfony docs these days just use the full class name, e.g.: https://symfony.com/doc/current/service_container/service_decoration.html , which makes things more concise.

But that pattern breaks if you want the same class to decorate 2 different services, as is the case here.

Maybe best to just update the name for the 2nd one to your suggestion, so that anyone who is just supporting a single content-type sees the more typical naming convention?

jamesisaac: Or maybe it's very common for users to want to support multiple content types, so worth using the `app.serializer` naming convention for both the services?  Let me know what you'd prefer.

jamesisaac: Good point about autowiring -- if these all used classical naming, then the `ApiNormalizer` class would not be available for autowired injection in other classes, unless an alias was also added, right? https://symfony.com/doc/current/service_container/autowiring.html#using-aliases-to-enable-autowiring

Can imagine that ending up catching some people out?  I've updated PR for now to leave first instance with FQCN, and next with classical naming.

alanpoulain (approved):

alanpoulain:
Thank you @jamesisaac.

PR #1358, opened 2021-05-10
fix: Clarify __invoke distinction

botris:
When creating custom controllers the invoke method parameter name differs depending on the HTTP method.

### DIFF
--- a/core/controllers.md
+++ b/core/controllers.md
@@ -65,7 +65,7 @@ This action will be automatically registered as a service (the service name is t
 API Platform automatically retrieves the appropriate PHP entity using the data provider then deserializes user data in it,
 and for `POST` and `PUT` requests updates the entity with data provided by the user.
 
-**Warning: when using `POST` or `PUT` the `__invoke()` method parameter [MUST be called `$data`](https://symfony.com/doc/current/components/http_kernel.html#getting-the-controller-arguments)**, otherwise, it will not be filled correctly!
+**Warning: when using `POST` or `PUT`, the `__invoke()` method parameter [MUST be called `$data`](https://symfony.com/doc/current/components/http_kernel.html#getting-the-controller-arguments)**, otherwise, it will not be filled correctly!
 When using `GET` the method parameter the `__invoke()` method parameter should be called the same as the entity identifier. So for path `/user/{uuid}/bookmarks` you must use `__invoke($uuid)`.
 
 Services (`$bookPublishingHandler` here) are automatically injected thanks to the autowiring feature. You can type-hint any service
--- a/core/controllers.md
+++ b/core/controllers.md
@@ -65,7 +65,8 @@ This action will be automatically registered as a service (the service name is t
 API Platform automatically retrieves the appropriate PHP entity using the data provider then deserializes user data in it,
 and for `POST` and `PUT` requests updates the entity with data provided by the user.
 
-**Warning: the `__invoke()` method parameter [MUST be called `$data`](https://symfony.com/doc/current/components/http_kernel.html#getting-the-controller-arguments)**, otherwise, it will not be filled correctly!
+**Warning: when using `POST` or `PUT` the `__invoke()` method parameter [MUST be called `$data`](https://symfony.com/doc/current/components/http_kernel.html#getting-the-controller-arguments)**, otherwise, it will not be filled correctly!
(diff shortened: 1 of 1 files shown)


core/controllers.md
```diff
@@ -65,7 +65,8 @@ This action will be automatically registered as a service (the service name is t
 API Platform automatically retrieves the appropriate PHP entity using the data provider then deserializes user data in it,
 and for `POST` and `PUT` requests updates the entity with data provided by the user.
 
-**Warning: the `__invoke()` method parameter [MUST be called `$data`](https://symfony.com/doc/current/components/http_kernel.html#getting-the-controller-arguments)**, otherwise, it will not be filled correctly!
+**Warning: when using `POST` or `PUT` the `__invoke()` method parameter [MUST be called `$data`](https://symfony.com/doc/current/components/http_kernel.html#getting-the-controller-arguments)**, otherwise, it will not be filled correctly!
```
alanpoulain, line 68:
```suggestion
**Warning: when using `POST` or `PUT`, the `__invoke()` method parameter [MUST be called `$data`](https://symfony.com/doc/current/components/http_kernel.html#getting-the-controller-arguments)**, otherwise, it will not be filled correctly!
```

core/controllers.md
```diff
@@ -65,7 +65,8 @@ This action will be automatically registered as a service (the service name is t
...
+When using `GET` the method parameter the `__invoke()` method parameter should be called the same as the entity identifier. So for path `/user/{uuid}/bookmarks` you must use `__invoke($uuid)`.
```
alanpoulain, line 69:
```suggestion
When using `GET`, the `__invoke()` method parameter should be called the same as the entity identifier. So for the path `/user/{uuid}/bookmarks`, you must use `__invoke($uuid)`.
```

alanpoulain:
Are you sure it's the case? Where is the attribute set?

core/controllers.md
botris: Language update

botris:
When I have this entity:
```
/**
 * @ApiResource(
 *     itemOperations={
 *         "example_start": {
 *              "method": "GET",
 *              "path": "/example/start/{uuid}.{_format}",
 *              "controller": StartExampleController::class,
 *              "read": false,
 *          }
 *     },
 *     collectionOperations={}
 * )
 */
class Example
{
    /**
     * @ApiProperty(identifier=true)
     */
    private UuidInterface $uuid;
```
And I do this in my controller
```
public function __invoke($data): Response
{
//
}
```
I get: `"Could not resolve argument $data of \"app\\controller\\startexamplecontroller()\", maybe you forgot to register the controller as a service or missed tagging it with the \"controller.service_arguments\"?"` 

But it works if I do
```
public function __invoke($uuid): Response
{
//
}
```
Which leads me to believe that the argument name must match the identifier name

alanpoulain:
OK I see why now, it's because the route parameters are given directly to the controller method.
I've added your change there https://github.com/api-platform/docs/pull/1390, thank you!

core/controllers.md
```diff
@@ -65,7 +65,8 @@ This action will be automatically registered as a service (the service name is t
 API Platform automatically retrieves the appropriate PHP entity using the data provider then deserializes user data in it,
 and for `POST` and `PUT` requests updates the entity with data provided by the user.
 
-**Warning: the `__invoke()` method parameter [MUST be called `$data`](https://symfony.com/doc/current/components/http_kernel.html#getting-the-controller-arguments)**, otherwise, it will not be filled correctly!
+**Warning: when using `POST` or `PUT`, the `__invoke()` method parameter [MUST be called `$data`](https://symfony.com/doc/current/components/http_kernel.html#getting-the-controller-arguments)**, otherwise, it will not be filled correctly!
+When using `GET`, the `__invoke()` method parameter should be called the same as the entity identifier. So for the path `/user/{uuid}/bookmarks`, you must use `__invoke($uuid)`.
```
guilliamxavier, line 69: I think that's wrong: the cause was not (only) you using `GET` but `"read": false`

alanpoulain: You can also retrieve the identifier even if `read` is `true`. But you are right, it should be modified to say that is also works for `GET`.

alanpoulain: See: https://github.com/api-platform/docs/pull/1392

PR #1363, opened 2021-05-19
docs: Add notice about disableReboot in tests with multiple requests

carlobeltrame:
See https://github.com/api-platform/api-platform/issues/1668
This notice can be removed again once that issue is resolved.

### DIFF
--- a/distribution/testing.md
+++ b/distribution/testing.md
@@ -228,7 +228,7 @@ from [AliceBundle](https://github.com/hautelook/AliceBundle/blob/master/README.m
 test, purge the database, load fixtures, begin a transaction, and, at the end of each test, roll back the
 transaction previously begun. Because of this, you can run your tests without worrying about fixtures.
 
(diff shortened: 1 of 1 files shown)


distribution/testing.md
```diff
@@ -228,6 +228,8 @@ from [AliceBundle](https://github.com/hautelook/AliceBundle/blob/master/README.m
 test, purge the database, load fixtures, begin a transaction, and, at the end of each test, roll back the
 transaction previously begun. Because of this, you can run your tests without worrying about fixtures.
 
+There is one caveat though: In some tests, it is necessary to perform multiple requests in one test, for example when creating a user via the API and checking that a subsequent login using the same password works. However, the client will by default reboot the kernel, which will reset the database. You can prevent this by adding `$client->disableReboot();` to such tests.
```
alanpoulain, line 231:
```suggestion
There is one caveat though: in some tests, it is necessary to perform multiple requests in one test, for example when creating a user via the API and checking that a subsequent login using the same password works. However, the client will by default reboot the kernel, which will reset the database. You can prevent this by adding `$client->disableReboot();` to such tests.
```

alanpoulain:
Thanks!

PR #1366, opened 2021-05-25
[docs] add missing subresource yaml documentation

johnkrovitch:
Add missing subresource yaml documentation

### DIFF
--- a/core/subresources.md
+++ b/core/subresources.md
@@ -159,7 +159,7 @@ class Answer
 # api/config/api_platform/resources.yaml
 App\Entity\Answer:
     subresourceOperations:
-        api_questions_answer_get_subresource:
+        api_answer_get:
             method: 'GET'
             normalization_context: {groups: ['foobar']}
 ```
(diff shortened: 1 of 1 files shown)


core/subresources.md
```diff
@@ -237,11 +252,25 @@ The `subresourceOperations` attribute also allows you to add an access control o
  }
 ```
 
+```yaml
+# config/api/resources/Answer.yaml
+App\Entity\Answer:
+    properties:
+        question:
+            subresource:
+                security: 'is_granted("ROLE_AUTHENTICATED")'
```
alanpoulain, line 261: It's not exactly the same configuration as the attribute. @soyuka is it normal?

johnkrovitch: In my tests, I could not make it works using the same configuration. Maybe should I fill a bug instead ?

johnkrovitch: Ok I can confirm, there is a bug with the security parameter and subresources. The parameter is not taken into account. But if I hardcode te following code into the `ApiPlatform\Core\Security\EventListener\DenyAccessListener` :

```php
#\ApiPlatform\Core\Security\EventListener\DenyAccessListener  l.88
$isGranted = 'is_granted("ROLE_ADMIN")';
````
the security parameter works. I can try to find where the problem is, but subresources will be removed in 3.0, isn't it ?

vincentchalamon: Hello @johnkrovitch,
Sorry for the delay of answer. Yes the subresources are deprecated in 2.7 and removed in 3.0, but the 2.7 will be maintained for months. If you feel motivated and identified this as a bug, I suggest you open a PR to fix it on 2.7, please :-)

johnkrovitch: Hello @vincentchalamon !
 No problem. I try to find again my use case where the issue appears. And submit a bugfix PR when I have some times (surely next week) :)

Thanks for your response

vinceAmstoutz:

PR #1376, opened 2021-06-14
fix(docs): add missing service configuration

lobodol:
Hi there,

I followed [this documentation](https://api-platform.com/docs/core/operations/#expose-a-model-without-any-routes) and noticed that decorator was never called. Comparing with [this chapter](https://api-platform.com/docs/core/openapi/#overriding-the-openapi-specification), I noticed the service configuration part was missing.

Here it is.

Regards,

### DIFF
--- a/core/operations.md
+++ b/core/operations.md
@@ -623,6 +623,14 @@ It's almost done, we have just one final issue: our fake item operation is visib
 To remove it, we will need to [decorate the Swagger documentation](openapi.md#overriding-the-openapi-specification).
 Then, remove the route from the decorator:
 
+```yaml
+# api/config/services.yaml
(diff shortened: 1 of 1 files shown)


lobodol:
Actually, I asked myself the same question :)

At first glance, I just followed the documentation and faced the error. Then, I saw the link about overriding the OpenAPI specification.

Should we assume all readers are as dummy as me 😅?

alanpoulain:
IMO it's a little strange to have both the sentence and the full configuration to do so.

lobodol:
It does make sens. I let you close this PR if you think it's redundant.

alanpoulain:
Thanks anyway.

PR #1382, opened 2021-06-17
docs: add YAML notation for single property deprecation

jzabrodin:
Add yaml notation  for single property deprecation


alanpoulain:
Thanks!

PR #1384, opened 2021-06-20
docs: change annotation to php 8 attributes, change links to ssl, som…

Chris53897:
…e php 8 changes

### DIFF
--- a/admin/handling-relations.md
+++ b/admin/handling-relations.md
@@ -62,11 +62,11 @@ For instance, if your API returns:
   "hydra:member": [
     {
       "@id": "/books/07b90597-542e-480b-a6bf-5db223c761aa",
-      "@type": "http://schema.org/Book",
+      "@type": "https://schema.org/Book",
       "title": "War and Peace",
       "author": {
         "@id": "/authors/d7a133c1-689f-4083-8cfc-afa6d867f37d",
-        "@type": "http://schema.org/Author",
+        "@type": "https://schema.org/Author",
         "firstName": "Leo",
         "lastName": "Tolstoi"
       }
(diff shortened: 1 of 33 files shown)


alanpoulain:
Hello and thank you for your PR!
To reduce the burden on the maintainers, could you:
- target `2.6` instead of `main` (it will avoid conflicts when we merge `2.6` into `main`)?
- split your PR on multiple ones (for instance the HTTPS links and the attributes, but you can also split the attributes per page)?

admin/handling-relations.md
```diff
@@ -62,11 +62,11 @@ For instance, if your API returns:
   "hydra:member": [
     {
       "@id": "/books/07b90597-542e-480b-a6bf-5db223c761aa",
-      "@type": "http://schema.org/Book",
+      "@type": "https://schema.org/Book",
```
dunglas, line 65: This should be reverted as we currently don't support entirely the `https` scheme for Schema.org (both Core and the JavaScript tools must be updated to support both schemes).

Another option is to patch these toolbox cours now that both schemes are supported by the vocabulary.

dunglas (changes requested):

Chris53897:
Hi. I will try to help as much as i can. Sorry, i thougth the SSL-Change was a no-brainer. 
My main focus at the moment is https://github.com/api-platform/api-platform/issues/1947
After that i try to find time to make the changes @alanpoulain suggested.

admin/handling-relations.md
Chris53897: I guess the requested split for https scheme in another request is a good idea. In general my opinon is that is always good to switch to https if available.

dunglas: Sure, but this will cause bug in this specific case because it's a namespace, and the new namespace is currently not supported. See https://github.com/api-platform/schema-generator/pull/329#issuecomment-830853254

dunglas: However, we want to add support for the `https://schema.org` namespace, so we can keep this PR open until it's supported, then merge it!

vinceAmstoutz:
@Chris53897 Thanks so much for your PR and the time you invested in contributing to this project! We truly appreciate your effort and involvement.

PR #1387, opened 2021-06-21
docs: Remove empty string defaults from entities

GaryClarke:
I'm not getting the SQL exception in 'Validating Data' section; empty strings are being stored instead of null being attempted.

The simplest fix is to remove the default empty strings from the entities.

This is using the Symfony and Composer install.

### DIFF
--- a/distribution/index.md
+++ b/distribution/index.md
@@ -647,6 +647,11 @@ Modify the following files as described in these patches:
       */
 +    #[Assert\Isbn]
      public ?string $isbn = null;
+     
+      * @ORM\Column
+      */
++    #[Assert\NotBlank]
+     public string $title = '';
  
       * @ORM\Column(type="text")
       */
--- a/distribution/index.md
+++ b/distribution/index.md
@@ -260,13 +260,13 @@ class Book
     public ?string $isbn = null;
 
     /** The title of this book. */
-    public string $title = '';
+    public string $title;
 
     /** The description of this book. */
-    public string $description = '';
+    public string $description;
 
     /** The author of this book. */
-    public string $author = '';
+    public string $author;
 
     /** The publication date of this book. */
     public ?\DateTimeInterface $publicationDate = null;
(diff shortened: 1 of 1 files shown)


alanpoulain:
I think it would be better to change only the `title` field by:
```php
public ?string $title = null;
```

GaryClarke:
@alanpoulain works for me...I'll update with this

dunglas:
Allowing null values here is bad DB and entity design IMHO. I would prefer to just change the wording to tell that allowing the user to not fill the title is not what we except, and introduce validation with the `#[NotBlank]` assertion. WDYT?

alanpoulain:
Not sure what you mean Kevin. The resource is not linked yet to DB. The Doctrine attribute can be not nullable but this one needs to be nullable in order to deserialize it correctly and have a `NotBlank` validation on it working as expected.

dunglas:
Exactly!

alanpoulain:
But currently (and sadly?) there is no other way to correctly have validation without using for instance DTOs.

dunglas:
I'm not sure to understand @alanpoulain

alanpoulain:
My bad maybe it works since `null` is not given directly in the input body.
You can leave it like this then but you need to modify the following paragraphs (when linking to Doctrine) @GaryClarke.

GaryClarke:
@alanpoulain good point...done now 👍

dunglas:
I don't think that we must remove the default values (this could trigger "initialized value" exceptions. Couldn't we remove the part about the SQL error and say that sending no values or an empty string doesn't trigger an error while it should?

GaryClarke:
OK...I'll change to that.

dunglas:
Thank you very much @GaryClarke, and sorry for the delay.

GaryClarke:
You're welcome @dunglas 🤝 ..even though I'd totally forgot about this 😂

PR #1389, opened 2021-06-23
docs: use attributes in controllers page

alanpoulain:


dunglas (approved):
+1 but maybe should we just follow Symfony's best practices (even if I don't like this one) and extends `AbstractController` as well as changing `__invoke` to another name.
It will probably make this documentation easier to understand for users used to "the Symfony way".

WDYT?

core/controllers.md
```diff
@@ -154,24 +156,20 @@ use ApiPlatform\Core\Annotation\ApiResource;
...
- *     }
- * })
- */
+#[ApiResource(itemOperations: [
+    'get',
+    'post_publication' => [
+        'method' => 'POST',
+        'path' => '/books/{id}/publication',
+        'controller' => CreateBookPublication::class,
+        'normalization_context' => ['groups' => ['publication']],
```
dunglas, line 165:
```suggestion
        'normalization_context' => ['groups' => 'publication'],
```

(syntactic sugar)

alanpoulain: Personally I prefer to always use arrays but OK.

alanpoulain:
> +1 but maybe should we just follow Symfony's best practices (even if I don't like this one) and extends `AbstractController` as well as changing `__invoke` to another name.
> It will probably make this documentation easier to understand for users used to "the Symfony way".
> 
> WDYT?

Why not for the `AbstractController`. I'm not so sure about changing `__invoke` though, we would need to reference the method name in the `controller` attribute and we would lose the easy way to only use the FQCN.

PR #1390, opened 2021-06-23
feat: thanks to the argument resolver, no need to use $data anymore

alanpoulain:
Related: https://github.com/api-platform/core/pull/3263


julienfalque:
Thank you for taking care of this!

PR #1394, opened 2021-06-29
docs: fix code sample to disable GET item operation

cristi-contiu:
The code sample for yaml and XML referenced a different controller then the code sample for PHP Attributes.

### DIFF
--- a/core/operations.md
+++ b/core/operations.md
@@ -260,7 +260,7 @@ App\Entity\Book:
         get: ~
     itemOperations:
         get:
-            controller: App\Controller\NotFoundAction
+            controller: ApiPlatform\Core\Action\NotFoundAction
             read: false
             output: false
 ```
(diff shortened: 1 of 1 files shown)


alanpoulain:
Hello, you should target `2.6` instead.

cristi-contiu:
@alanpoulain  Thank you for your reply. As a side note, this issue affects the [2.5 docs](https://api-platform.com/docs/v2.5/core/operations/) as well as 2.6.

alanpoulain:
> @alanpoulain Thank you for your reply. As a side note, this issue affects the [2.5 docs](https://api-platform.com/docs/v2.5/core/operations/) as well as 2.6.

Yes but we don't maintain old versions.

alanpoulain (approved):

alanpoulain:
Thank you!

PR #1395, opened 2021-07-04
Include @Ignore annotation in serialization docs

rowanparker:


core/serialization.md
```diff
@@ -153,6 +156,19 @@ In addition to the `groups` key, you can configure any Symfony Serializer option
 Any serialization and deserialization group that you specify will also be leveraged by the built-in actions and the Hydra
 documentation generator.
 
+The `isbn` property in the above example will not appear in any collection or item requests. It will however appear in the `/api/contexts/Book` request.
```
dunglas, line 159: This looks like a bug to me. If the property is marked as internal using serialization groups, it shouldn't appear in the context. Would you mind opening an issue on `core/` about this (or even a PR if you feel motivated :D) and to remove this sentence, as it's probably a not idea to document a bug.

dunglas (changes requested):
Otherwise, nice addition!

PR #1396, opened 2021-07-06
Update documentation for php 7

alihasana:
Currently documention for creating JwtDecorator in php 8, so i got few errors in php7


alanpoulain:
Hello, thanks for the PR but we don't want to maintain multiple versions of PHP in the documentation. Only the last stable one.

PR #1397, opened 2021-07-08
Update example in "Calling multiple DataPersisters"

jmontoyaa:
The example doesn't work when there are multiple data persisters.

Explanation and solution posted here https://github.com/api-platform/api-platform/issues/1606

### DIFF
--- a/core/data-persisters.md
+++ b/core/data-persisters.md
@@ -230,7 +230,10 @@ final class BlogPostDataPersister implements ContextAwareDataPersisterInterface,
 services:
     # ...
     App\DataPersister\BlogPostDataPersister: ~
-        # Uncomment only if autoconfiguration is disabled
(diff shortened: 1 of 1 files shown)


core/data-persisters.md
```diff
@@ -230,7 +230,10 @@ final class BlogPostDataPersister implements ContextAwareDataPersisterInterface,
 services:
     # ...
     App\DataPersister\BlogPostDataPersister: ~
-        # Uncomment only if autoconfiguration is disabled
-        #arguments: ['@App\DataPersister\BlogPostDataPersister.inner']
-        #tags: [ 'api_platform.data_persister' ]
+        arguments: ['@api_platform.doctrine.orm.data_persister']
```
alanpoulain, line 233: In the example, the entity manager is used, not the API Platform data persister.

jmontoyaa: you're right, thanks, closing

PR #1402, opened 2021-07-23
Explain how the default IRI normalization works

silvan-vivars:
The serialiser can (and will) turn referenced ApiResources into IRIs - if it knows the referenced item is an ApiResource. It determines this based on the Doctrine annotation, declared type or DocBlock type annotation. If an ApiResource is not a Doctrine entity (which is the recommended approach for non-prototyping apps), it is very easy to not "get" this default behaviour if no type is declared.

### DIFF
--- a/core/serialization.md
+++ b/core/serialization.md
@@ -234,6 +234,31 @@ Refer to the [operations](operations.md) documentation to learn more.
 By default, the serializer provided with API Platform represents relations between objects using [dereferenceable IRIs](https://en.wikipedia.org/wiki/Internationalized_Resource_Identifier).
 They allow you to retrieve details for related objects by issuing extra HTTP requests. However, for performance reasons, it is sometimes preferable to avoid forcing the client to issue extra HTTP requests.
 
+**Note:** The serializer automatically creates the IRIs if it can determine that the target type is also an ApiResource. It does this based on the Doctrine configuration, the declared PHP type or the type annotation declared in the DocBlock. This means that if your ApiResource is also a Doctrine entity this will work automatically. However, **if your ApiResource is not a Doctrine entity, special care must be taken:** make sure to declare the type or add a type annotation in a DocBlock. For example:
+
+```PHP
+<?php
+#[ApiResource]
+class Bar {...}
+
+#[ApiResource]
+class Foo {
+  public $bars; // <- This will NOT work; $bars will return as an array of embedded Bar objects instead of an array of IRIs.
+}
+
+
+#[ApiResource]
+class Foo {
+  // This will work; $bars will return as an array of IRIs.
+  /**
+   * @var Bar[]
+   */
+  public $bars;
+}
+```
+
+
+
 **Note:** We strongly recommend using [Vulcain](https://vulcain.rocks) instead of this feature. Vulcain allows creating faster (better hit rate) and better designed APIs than relying on compound documents, and is supported out of the box in the API Platform distribution.
 
 ### Normalization


core/serialization.md
```diff
@@ -234,6 +234,31 @@ Refer to the [operations](operations.md) documentation to learn more.
 By default, the serializer provided with API Platform represents relations between objects using [dereferenceable IRIs](https://en.wikipedia.org/wiki/Internationalized_Resource_Identifier).
 They allow you to retrieve details for related objects by issuing extra HTTP requests. However, for performance reasons, it is sometimes preferable to avoid forcing the client to issue extra HTTP requests.
 
+**Note:** The serializer automatically creates the IRIs if it can determine that the target type is also an ApiResource. It does this based on the Doctrine configuration, the declared PHP type or the type annotation declared in the DocBlock. This means that if your ApiResource is also a Doctrine entity this will work automatically. However, **if your ApiResource is not a Doctrine entity, special care must be taken:** make sure to declare the type or add a type annotation in a DocBlock. For example:
+
+```PHP
```
alanpoulain, line 239:
```suggestion
```php
```

core/serialization.md
```diff
@@ -234,6 +234,31 @@ Refer to the [operations](operations.md) documentation to learn more.
...
+<?php
+#[ApiResource]
+class Bar {...}
+
+#[ApiResource]
+class Foo {
+  public $bars; // <- This will NOT work; $bars will return as an array of embedded Bar objects instead of an array of IRIs.
+}
```
alanpoulain, line 247: I don't think it is necessary to add the "bad" example.

core/serialization.md
```diff
@@ -234,6 +234,31 @@ Refer to the [operations](operations.md) documentation to learn more.
...
+
+
+#[ApiResource]
+class Foo {
+  // This will work; $bars will return as an array of IRIs.
+  /**
+   * @var Bar[]
+   */
+  public $bars;
```
alanpoulain, line 256:
```suggestion
  public array $bars;
```

core/serialization.md
```diff
@@ -234,6 +234,31 @@ Refer to the [operations](operations.md) documentation to learn more.
...
+}
+```
+
+
+
```
alanpoulain, line 261: Lines to remove.

core/serialization.md
```diff
@@ -234,6 +234,31 @@ Refer to the [operations](operations.md) documentation to learn more.
 By default, the serializer provided with API Platform represents relations between objects using [dereferenceable IRIs](https://en.wikipedia.org/wiki/Internationalized_Resource_Identifier).
 They allow you to retrieve details for related objects by issuing extra HTTP requests. However, for performance reasons, it is sometimes preferable to avoid forcing the client to issue extra HTTP requests.
 
+**Note:** The serializer automatically creates the IRIs if it can determine that the target type is also an ApiResource. It does this based on the Doctrine configuration, the declared PHP type or the type annotation declared in the DocBlock. This means that if your ApiResource is also a Doctrine entity this will work automatically. However, **if your ApiResource is not a Doctrine entity, special care must be taken:** make sure to declare the type or add a type annotation in a DocBlock. For example:
```
alanpoulain, line 237:
```suggestion
**Note:** The serializer automatically creates the IRIs if it can determine that the target type is also an ApiResource. It does this based on the Doctrine configuration, the declared PHP type or the type annotation declared in the PHPDoc. This means that if your ApiResource is also a Doctrine entity this will work automatically. However, if your ApiResource is not a Doctrine entity, special care must be taken: make sure to declare the type or add a type annotation in the PHPDoc. For example:
```

core/serialization.md
silvan-vivars: I spent the most of my friday debugging why this wasn't working. I figure it's best to be as clear and unambiguous as possible to illustrate what works and what doesn't. Especially considering people tend to read the code snippets first (I do, at least)

core/serialization.md
silvan-vivars: My thought was to show the absolute minimum necessary to display how this works so I purposefully did not declare the type as one could then question if it is the `@var` that makes the behaviour work, or the `array` declaration.

core/serialization.md
alanpoulain: I understand, but I don't think there are examples like this in the other parts of the documentation, and it's not the case in the Symfony documentation either.
I would prefer to be consistent in how the examples are presented.

core/serialization.md
alanpoulain: Don't you think it's better like this? It shows that the PHPDoc has priority over the PHP type.

PR #1370, opened 2021-05-26
docs: add missing serialization yaml documentation

johnkrovitch:
Add missing serialization yaml documentation

PR #1403, opened 2021-08-07
Fix hidden PHP code snippet and add clarifying sentence

carlobeltrame:
In #1370, YAML documentation was added to the serializer page. In one instance, there were two PHP snippets being merged into a single code selector. Since the code selector cannot handle two PHP tabs (if you click the second one, the first one will be activated), this hid some information from that page.
This confused us a lot when setting up our embedding infrastructure, so I also added a sentence to describe the mechanism.


alanpoulain (approved):

alanpoulain:
Thank you.

PR #1404, opened 2021-08-13
Vuetify Generator: Fix section i18n

DimNS:
In the section about i18n, instead of creating the file src/plugins/vuetify.js you need to create a file src/i18n.js


DimNS:
The pull-request was created incorrectly, I will recreate it

PR #1405, opened 2021-08-13
Vuetify Generator: Fix deprecated

DimNS:
v-content is deprecated it should be replaced with v-main


DimNS:
The pull-request was created incorrectly, I will recreate it

PR #1416, opened 2021-08-27
Update filters.md

jmediatechnology:
It is better to use `Type::BUILTIN_TYPE_STRING` in the documentation. 
The reasons for it are: 
- It's a Symfony thing
- By using Symfony Property Info Type the manual conveys the right thinking direction. The custom filter wont work if you use `'boolean'` as the type, the valid value for boolean is `'bool'` which is the value of the constant `Type::BUILTIN_TYPE_BOOL`, 
- By using `Type::BUILTIN_TYPE_STRING` in the documentation, developers will be pushed to the right thinking direction when building a custom filter


dunglas (approved):

dunglas:
LGTM

PR #1417, opened 2021-08-31
Adds  argument to service JWTDecorator

pamuche:
Like in  https://api-platform.com/docs/core/openapi/ its now (Symfony 5.3.7) necessary to add this argument. Otherwise it  will produce circular reference for `App\OpenApi\JwtDecorator`.


kochen:
@pamuche See #1418 & https://github.com/kochen/docs/pull/1 :D

dunglas:
Thanks for contributing to the docs!

PR #1420, opened 2021-08-31
Update serialization.md : Remove unused import in BookAttributeNormalizer

xElysioN:
`use Symfony\Component\Serializer\Normalizer\NormalizerInterface;` is not used anymore in `BookAttributeNormalizer`

https://api-platform.com/docs/core/serialization/#changing-the-serialization-context-on-a-per-item-basis


dunglas:
Good catch @xElysioN

PR #1422, opened 2021-09-03
docs: 2.7 update

PierreThibaudeau:
· php8 attributes instead of annotations.
· Removes item & collection operations
· ApiPlatform\Core\Annotation\ to ApiPlatform\Metadata\
· forgotten file names
· adds "declare_strict" at the beginning of php  examples

### DIFF
--- a/core/controllers.md
+++ b/core/controllers.md
@@ -31,7 +31,6 @@ First, let's create your custom operation:
 ```php
 <?php
 // api/src/Controller/CreateBookPublication.php
-
 namespace App\Controller;
 
 use App\Entity\Book;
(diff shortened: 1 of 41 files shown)


core/controllers.md
```diff
@@ -155,20 +160,24 @@ You may want different serialization groups for your custom operations. Just con
...
-        'method' => 'POST',
-        'path' => '/books/{id}/publication',
-        'controller' => CreateBookPublication::class,
-        'normalization_context' => ['groups' => 'publication'],
-    ],
-])]
+#[ApiResource]
+#[Get]
+#[Post(
+    name: 'post_publication',
```
soyuka, line 176:
```suggestion
    name: 'publication', 
```

core/controllers.md
```diff
@@ -231,19 +240,23 @@ operation attribute:
...
-        'method' => 'POST',
-        'path' => '/books/{id}/publication',
-        'controller' => CreateBookPublication::class,
-        'read' => false,
-    ],
-])]
+#[ApiResource]
+#[Get]
+#[Post(
+    name: 'post_publication',
```
soyuka, line 255:
```suggestion
    name: 'publication', 
```

dunglas:
I would remove the strict type declaration as it's not related with API Platform (the library works well with or without enabling this feature).

Maybe could we mention somewhere that it's supported, or add them only in the getting started guide to show this as a best practice, but "polluting" every snippet with this declaration isn't worth it IMHO.

PierreThibaudeau:
> I would remove the strict type declaration as it's not related with API Platform (the library works well with or without enabling this feature).
> 
> Maybe could we mention somewhere that it's supported, or add them only in the getting started guide to show this as a best practice, but "polluting" every snippet with this declaration isn't worth it IMHO.

I thought it would be nice to standardise these snippets but i could do the opposite, i.e. remove them all.

PR #1423, opened 2021-09-06
fix: Broken links on Data Providers page

Kr42:
On this page https://api-platform.com/docs/core/data-providers/


dunglas:
Thanks @Kr42

PR #1428, opened 2021-09-09
Change annotations to attributes

Catgolin:
Changed annotations into php 8 attributes in the admin and core pages.


dunglas:
Thank you very much!!

PR #1447, opened 2021-10-20
fix: remove 'anonymous' security in JWT config

### LINKED ISSUE #1445: JWT Authentication SecurityBundle configuration references 'anonymous' security
The [JWT Authentication docs](https://github.com/api-platform/docs/blob/087014d8f7a537240a2b7b8c09b613315104904c/core/jwt.md#configuring-the-symfony-securitybundle) make use of the `anonymous` option for the main firewall config.

This was [removed in the 5.1-dev branch](https://github.com/symfony/symfony/commit/09f90792e7b7d5f2383e59c0f9693d5d7de8ec4f#diff-150aa7e77e0d0973854990a3aaaae5de010f211042c34a9cfd8b529582f0cd4dR50) and causes an error when used:

```
/srv/api # php bin/console

In AnonymousFactory.php line 52:                                                                                                                   
The authenticator manager no longer has "anonymous" security. Please remove this option under the "main" firewall.
```

There is an extensive reasoning in the linked commit. I _think_ the bottom line is mentioning the option in the api-platform documentation configuration example can simply be removed for the same effect.

wiese:
The anonymous option for the firewall config was removed in symfony 5.1
as explained here: https://github.com/symfony/symfony/pull/36574

Using the option caused an error but it can be skipped. "[A] user that
is not authenticated is similar to an "unknown" user: They both have no
rights at all."

Resolves #1445

### DIFF
--- a/core/jwt.md
+++ b/core/jwt.md
@@ -84,7 +84,6 @@ security:
             security: false
         main:
             stateless: true
-            anonymous: true
             provider: app_user_provider
             json_login:
                 check_path: /authentication_token


alanpoulain:
Thank you.

PR #1448, opened 2021-10-20
feat: use JWT bundle's command to create keys

### LINKED ISSUE #1446: [Suggestion] JWT Authentication key pair generation via (lexik) bundle command
The [JWT Authentication docs](https://api-platform.com/docs/core/jwt/#installing-lexikjwtauthenticationbundle) show how to install and set up the lexik JWT AuthenticationBundle, incl. generating the key pair.

Apparently, the bundle now (added 8 months ago) comes with a [GenerateKeyPairCommand](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/3f30e12920d38fb00bd2720677cbfc2464911caa/Command/GenerateKeyPairCommand.php) to – as the name implies – generate the key pair. Maybe documenting the use of this console command instead of the list of CLI commands can make the api-platform docs less cumbersome and less prone to fall out of sync with the bundle.

wiese:
Creation of directory and files is now covered by
lexik/jwt-authentication-bundle's command.

This somewhat causes the directory/file location somewhat pop up out of
nowhere in the subsequent permission change (`setfacl`). That could be
avoided by extracting JWT_SECRET_KEY and JWT_PUBLIC_KEY from api's `.env`
file (section maintained by the bundle) but it adds bloat without apparent
benefit.

The previously documented behaviour was closer to using the `--overwrite`
option on the command, but I doubt it is in the user's best interest in
a starter guide. Using `--skip-if-exists` would make some sense, but
bailing feels like the sane option – the user obviously touched this
area before and should be made aware.

Resolves #1446

### DIFF
--- a/core/jwt.md
+++ b/core/jwt.md
@@ -24,10 +24,7 @@ Then we need to generate the public and private keys used for signing JWT tokens
 docker-compose exec php sh -c '
     set -e
     apk add openssl
-    mkdir -p config/jwt
-    jwt_passphrase=${JWT_PASSPHRASE:-$(grep ''^JWT_PASSPHRASE='' .env | cut -f 2 -d ''='')}
-    echo "$jwt_passphrase" | openssl genpkey -out config/jwt/private.pem -pass stdin -aes256 -algorithm rsa -pkeyopt rsa_keygen_bits:4096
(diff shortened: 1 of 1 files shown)


dunglas:
Thanks!

PR #1455, opened 2021-10-28
docs: Enhance Json schema documentation

ES-Six:
This PR is intended to improve the JSON Schema documentation to specify the Overriding JSON Schema specification use case (usefull when making unit tests).

This PR concern an already existing functionnality that was undocumented.


ES-Six:
All requested changes have been taken into account.

ES-Six:
Do not hesitate if you need additional changes or improvements 😉.

.gitignore
```diff
@@ -1 +1,2 @@
 /website
+.idea/
```
dunglas, line 2: To revert

dunglas (approved):

.gitignore
ES-Six: Ok, I will revert .gitignore .

Why do you want to not gitignore IDE files ?

ES-Six:
Requested changes are now applied.

.gitignore
alanpoulain: I think it's mainly because of this: https://dev.to/oleksiyrudenko/please-do-not-pollute-projects-gitignore-3h8e

ES-Six: Ok, so you are recommending to use the global gitignore.

ES-Six: Modification applied.

core/json-schema.md
```diff
@@ -21,6 +21,75 @@ docker-compose exec php \
...
+
+use ApiPlatform\Core\Annotation\ApiResource;
+use Doctrine\ORM\Mapping as ORM;
+use Symfony\Component\Serializer\Annotation\Groups;
+
+#[ORM\Entity]
+#[ApiResource(
+    collectionOperations: [
+        'get' => ['normalization_context' => ['groups' => 'greeting:collection:get']],
+    ],
```
alanpoulain, line 53: I don't think it's necessary for this example.

core/json-schema.md
```diff
@@ -21,6 +21,75 @@ docker-compose exec php \
...
+            "items" => ["type" => "integer"]
+        ]
+    ])]
+    public function getSomeNumbers(): array {
+        return [1, 2, 3, 4];
+    }
+}
+```
+
+Be careful: the `json_schema_context` property does not always have the same properties as the `openapi_context` property.
```
alanpoulain, line 87: Are you sure? OpenAPI and JSON Schema should be fully compatible now.

ES-Six: I tested it while doing unit tests and using the `assertMatchesResourceCollectionJsonSchema` method and I while testing on an entity with JSON column in it.

For this doctrine JSON column, I was using the openapi_context "type" => "json" to make it appear well in the documentation. But in a unit testing context, the JSON Schema automatically generated ignore the openapi_context.

Example of api property used in my project : 

```php
#[ApiProperty(attributes: [
        "openapi_context" => [
            "type" => "json",
            "example"=>"{...json content here...}"
        ]
    ])]
```

So I tried to set the JSON schema to "type" => "json".
Like this : 
Example of api property used in my project : 

```php
#[ApiProperty(attributes: [
        "openapi_context" => [
            "type" => "json",
            "example"=>"{...json content here...}"
        ],
        "json_schema_context" => [
            "type" => "json"
        ]
    ])]
```

But it doesn't work because the type JSON is not a part of the JSON Schema specification, so unit test fails as it take the default type for this context inside API Platform instead.

Available JSON Schema types are : http://json-schema.org/understanding-json-schema/reference/type.html

So to solve the problem in this specific case, I need to use the type to make the unit test to pass

```php
#[ApiProperty(attributes: [
        "openapi_context" => [
            "type" => "json",
            "example"=>"{...json content here...}"
        ],
        "json_schema_context" => [
            "type" => "array"
        ]
    ])]
```

So perhaps, perhaps it's not an error, as it also seems to works when I specify "type" => "array" for both.

With "type" => "json" in the openapi_context the schema in the swagger do appear like this (excpected behaviour) :
image

With "type" => "array" in the openapi_context the schema in the swagger do appear like this (not the excpected behaviour) :

image

core/json-schema.md
ES-Six: Ok, I can remove it if you want, I just reuse an existing exemple in the documentation so it looks like a complete API Platform ressource.

core/json-schema.md
alanpoulain: I don't think `json` type is a valid data type. Swagger UI simply displays it by default as a string.

ES-Six: I tested more and I observed :
1. Every doctrine entity with a JSON column (with type JSON) is considered as string type by default in the generated swagger doc by API Platform.

2. The type "json" appear to be generated correctly in the swagger doc json result :

image


3. openapi or swagger type "json" is not documented but is really understood by the swagger doc as we can observe on the screen capture that "json" appear as a type. I could not reproduce this behavior without specify the "json" as "openapi_context" type (I tried, array, object, string and nothing (to get the default type)):


So I conclude that something is handling well the json type (perhaps in swagger, perhaps in api platform I don't know).
So if "json" is not an officially valid type for OpenAPI and Swagger Contexts, I don't know why it works so well, everything works perfectly with the "json" type, it cannot be a coincidence is it an undocumented feature 😅 ?

Indeed, the "json" type seems not officially supported by the openapi data types specs, but we can observe he is recognized by the swagger doc as it appears explicitly as a "json" type and is handled well (see 1st screen capture), I use it since more than 1 month in 2 projects and no problem.

I'm investigating more...

ES-Six: Part 2 :

After more investigations, it seems when any unknown type is set, swagger is ignoring it, but it display the invalid type type and rely on the "items" property

ES-Six: > I don't think `json` type is a valid data type. Swagger UI simply displays it by default as a string.

As any invalid type seems to be supported without throwing an error or warning and is displayed in the swagger doc : how an invalid type should be handled ? Why it works ?

ES-Six: > Are you sure? OpenAPI and JSON Schema should be fully compatible now.

> I don't think json type is a valid data type. Swagger UI simply displays it by default as a string.

As I investigated more here is what I understood :

Yes, OpenAPI and JSON Schema are fully compatible. So I need to change the doc. Indeed I miss understood these 2 specifications.

The type property is only an informative text OR if it's a known value (example : string, integer, array, object, etc...) it will be used to handle specific things on the swagger doc.

Example :
```php
#[ApiProperty(attributes: [
	"openapi_context" => [
        "type" => "json"
    ]
])]
```

core/json-schema.md
ES-Six: This modification is now applied. The ApiRessource attribute is now the shortest form as possible.

core/json-schema.md
ES-Six: This erroneous part is now removed :)

ES-Six:
Requested changes are now applied.

core/json-schema.md
```diff
@@ -21,6 +21,66 @@ docker-compose exec php \
...
+use Symfony\Component\Serializer\Annotation\Groups;
+
+#[ORM\Entity]
+#[ApiResource]
+class Greeting
+{
+    #[ORM\Id]
+    #[ORM\GeneratedValue]
+    #[ORM\Column(type: "integer")]
+    #[Groups("greeting:collection:get")]
```
alanpoulain, line 56:
```suggestion
```

core/json-schema.md
```diff
@@ -21,6 +21,66 @@ docker-compose exec php \
...
+You will need to add the `json_schema_context` property in the [`ApiProperty`](openapi.md#using-the-openapi-and-swagger-contexts) attribute to do this, example:
+
+```php
+<?php
+
+namespace App\Entity;
+
+use ApiPlatform\Core\Annotation\ApiResource;
+use Doctrine\ORM\Mapping as ORM;
+use Symfony\Component\Serializer\Annotation\Groups;
```
alanpoulain, line 47:
```suggestion
```

alanpoulain (approved):

alanpoulain:
Thank you @ES-Six.

PR #1456, opened 2021-10-29
fix: [DOC] Use argument binding instead of decoration for Data Persister

danieleambrosino:
Using classic decoration for the custom Data Persister in services.yaml will replace the core persister.
Probably this isn't the most common case scenario; normally you would _add_ your custom Data Persister to the chain of persisters instead of replacing it.
Use attribute binding instead, as already explained in SymfonyCasts' video.


soyuka:
I'm doing this also :)

PR #1457, opened 2021-10-29
fix: [doc] MultipartDecoder to avoid Invalid IRI

Snowbaha:
this fix allow to add another relation not required. "" has to be null.


Snowbaha:
Nop but I tried and no errors : 
image

So do you think it's useless or it can prevent some issue?

alanpoulain:
I think sending an empty value is a valid use case and should not be converted to `null`.

Snowbaha:
Ok no problem, thank you for your advice!

PR #1458, opened 2021-10-30
docs: Upgrade Filters documentation to use php8 attributes instead of annotations

ES-Six:
This PR is intended to enhance the Filter documentation by converting Doctrine annotations to PHP8 attributes since doctrine now supports well the PHP8 attributes: 

Here is the list of modifications:
- Every doctrine annotations in the examples have been replaced by doctrine attributes (as in the latest doctrine version, attribute usage is now enabled by default).
- Entity properties typing is now enforced.
- The Doctrine ORM Filter example (UserFilter.php) shows how to create and use a custom attribute instead of how to create a custom annotation.


core/filters.md
```diff
@@ -1410,10 +1410,8 @@ class Order
 {
     // ...
 
-    /**
-     * @ORM\ManyToOne(targetEntity="User")
-     * @ORM\JoinColumn(name="user_id", referencedColumnName="id")
-     */
+    #[ORM\ManyToOne(targetEntity: User::class)]
```
dunglas, line 1413: You can add `User` as the type of the property  and remove the `targetEntity` prop.

core/filters.md
```diff
@@ -1422,23 +1420,25 @@ class Order
...
-/**
- * @Annotation
- * @Target("CLASS")
- */
+#[Attribute(Attribute::TARGET_CLASS)]
 final class UserAware
 {
     public $userFieldName;
+    
+    public function __construct(string $userFieldName)
```
dunglas, line 1438: Why do we need this construbtor?

core/filters.md
```diff
@@ -1468,33 +1466,35 @@ Now, create a Doctrine filter class:
...
+        // The Doctrine filter is called for any query on any entity
+        // Check if the current entity is "user aware" (marked with an attribute)
+        $attributes = $targetEntity->getReflectionClass()->getAttributes();
+        $userAware = null;
+        foreach($attributes as $attribute) {
+            if ($attribute->getName() === UserAware::class) {
+                $userAware = $attribute;
+                break;
+            }
         }
```
dunglas, line 1487:
```suggestion
        $userAware = $targetEntity->getReflectionClass()->getAttributes(UserAware::class)[0] ?? null;
```

core/filters.md
```diff
@@ -1468,33 +1466,35 @@ Now, create a Doctrine filter class:
...
+        $fieldName = $userAware?->getArguments()['userFieldName'] ?? null;
+        if ($fieldName === '' || is_null($fieldName)) {
             return '';
         }
 
-        $fieldName = $userAware->userFieldName;
         try {
             // Don't worry, getParameter automatically escapes parameters
             $userId = $this->getParameter('id');
-        } catch (\InvalidArgumentException $e) {
```
dunglas, line 1497: You should revert this one, we follow Symfony CS.

core/filters.md
```diff
@@ -1524,7 +1519,11 @@ doctrine:
                 class: App\Filter\UserFilter
 ```
 
-Add a listener for every request that initializes the Doctrine filter with the current user in your bundle services declaration file.
+Done: Doctrine will automatically filter all "UserAware" entities!
```
dunglas, line 1522:
```suggestion
Done: Doctrine will automatically filter all `UserAware`entities!
```

core/filters.md
```diff
@@ -1524,7 +1519,11 @@ doctrine:
...
+
+Using PHP8 attributes, there is no need to add a listener for every request that initializes the Doctrine filter with the current user in your bundle services declaration file.
```
dunglas, line 1524:
```suggestion
Using PHP attributes, there is no need to add a listener for every request that initializes the Doctrine filter with the current user in your bundle services declaration file.
```

dunglas: IMHO this sentence can be removed because the old code has been removed.

core/filters.md
```diff
@@ -1524,7 +1519,11 @@ doctrine:
...
+
+If you are using annotations instead of attributes, you can do as follows to pass a doctrine annotation `Reader` instance to your Doctrine Filter :
```
dunglas, line 1526: I would remove the example using annotations.

core/filters.md
```diff
@@ -1590,60 +1589,62 @@ final class UserFilterConfigurator
...
-    public $name;
+    public ?string $name;
 
-    /**
-     * @ORM\OneToMany(targetEntity="DummyCarColor", mappedBy="car")
-     */
+    #[ORM\OneToMany(mappedBy: "car", targetEntity: DummyCarColor::class)]
     #[ApiFilter(SearchFilter::class, properties: ['colors.prop' => 'ipartial'])]
-    public $colors;
+    public ?Collection $colors;
```
dunglas, line 1623:
```suggestion
    public Collection $colors;
```

(as it's always initialized)

core/filters.md
```diff
@@ -1590,60 +1589,62 @@ final class UserFilterConfigurator
...
-     */
+    #[ORM\Id]
+    #[ORM\GeneratedValue]
+    #[ORM\Column(type: 'integer')]
+    private ?int $id = null;
+
+    #[ORM\Column(type: 'string')]
     #[ApiFilter(SearchFilter::class, strategy: 'partial')]
-    public $name;
+    public ?string $name;
```
dunglas, line 1619:
```suggestion
    public ?string $name = null;
```

(or it will never be null)

dunglas (changes requested):
Good job!

core/filters.md
ES-Six: Ok, example using annotations is now removed.

core/filters.md
ES-Six: Ok, change is reverted.

core/filters.md
ES-Six: It's a mistake, I removed the constructor.

core/filters.md
ES-Six: If I understand well, you want it to be like this : `#[ORM\ManyToOne("User")]`.

core/filters.md
ES-Six: Sentence removed :)

core/filters.md
ES-Six: Example updated.

ES-Six:
All requested changes have been taken into account.

dunglas:
Thanks!!

PR #1464, opened 2021-11-08
source folder for build caddy change

JvB94:
caddy is a part of the dockerfile who is located in api folder. The current source folder does not exist


dunglas:
Good catch, thanks!

PR #1467, opened 2021-11-12
fix: remove anonymous option under firewall

samnela:
The authenticator manager no longer has "anonymous" security.


dunglas:
Good catch, thanks!

PR #1473, opened 2021-11-22
chore: update XML/YAML for 3.0 format

vincentchalamon:
Downgrade github/super-linter to 3.17.0 because of https://github.com/github/super-linter/issues/2255


soyuka:
Thanks @vincentchalamon huge work !

PR #1477, opened 2021-11-24
fix: remove legacy doctrine php 8 attribute note

solverat:
This is supported in 2.9: https://www.doctrine-project.org/2021/05/24/orm2.9.html


alanpoulain:
Thanks!

PR #1484, opened 2021-12-14
docs: Update authentication-support.md

4c0n:
There is no App.js file anymore, further more the demo uses the typescript file index.tsx as included in this change.

### DIFF
--- a/admin/authentication-support.md
+++ b/admin/authentication-support.md
@@ -15,7 +15,6 @@ import { hydraDataProvider as baseHydraDataProvider, fetchHydra as baseFetchHydr
 import parseHydraDocumentation from "@api-platform/api-doc-parser/lib/hydra/parseHydraDocumentation";
 import authProvider from "utils/authProvider";
 import { ENTRYPOINT } from "config/entrypoint";
-import Login from "components/admin/Login";
 
 const getHeaders = () => localStorage.getItem("token") ? {
   Authorization: `Bearer ${localStorage.getItem("token")}`,
@@ -63,7 +62,7 @@ const dataProvider = baseHydraDataProvider({
 const AdminLoader = () => {
   if (typeof window !== "undefined") {
     const { HydraAdmin } = require("@api-platform/admin");
-    return <HydraAdmin dataProvider={dataProvider} authProvider={authProvider} entrypoint={window.origin} loginPage={Login} />;
+    return <HydraAdmin dataProvider={dataProvider} authProvider={authProvider} entrypoint={window.origin} />;
   }
 
   return <></>;
--- a/admin/authentication-support.md
+++ b/admin/authentication-support.md
@@ -19,65 +19,65 @@ import { ENTRYPOINT } from "config/entrypoint";
 import Login from "components/admin/Login";
 
 const getHeaders = () => localStorage.getItem("token") ? {
-    Authorization: `Bearer ${localStorage.getItem("token")}`,
+  Authorization: `Bearer ${localStorage.getItem("token")}`,
 } : {};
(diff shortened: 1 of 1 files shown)


alanpoulain:
Thanks, could you keep the style like it was before (2 spaces)?

admin/authentication-support.md
```diff
@@ -6,66 +6,80 @@ for more information.
 
 In short, you have to tweak the data provider and the api documentation parser like this:
 
-```javascript
-// admin/src/App.js
+```typescript
+// pwa/pages/admin/index.tsx
 
+import Head from "next/head";
+import React from "react";
```
alanpoulain, line 13: Not needed.

admin/authentication-support.md
```diff
@@ -6,66 +6,80 @@ for more information.
...
 import { Redirect, Route } from "react-router-dom";
-import { HydraAdmin, hydraDataProvider as baseHydraDataProvider, fetchHydra as baseFetchHydra, useIntrospection } from "@api-platform/admin";
+import { hydraDataProvider as baseHydraDataProvider, fetchHydra as baseFetchHydra, useIntrospection } from "@api-platform/admin";
 import parseHydraDocumentation from "@api-platform/api-doc-parser/lib/hydra/parseHydraDocumentation";
-import authProvider from "./authProvider";
+import authProvider from "utils/authProvider";
+import { ENTRYPOINT } from "config/entrypoint";
```
alanpoulain, line 18: It doesn't exist. Please do not just copy/paste from the demo.

4c0n: yet it is right here: https://github.com/api-platform/api-platform/blob/04b0b30a545acb5bebbe2dd76bb4e3d6aac69a5a/pwa/config/entrypoint.ts#L1

4c0n: If you're actually talking about the `authProvider` it also didn't exist in the original example and the text still clearly states to look at the demo for an example or to read the `react-admin` docs.

4c0n:
> Thanks, could you keep the style like it was before (2 spaces)?

done

admin/authentication-support.md
alanpoulain: You're right, it was my mistake.

admin/authentication-support.md
```diff
@@ -6,15 +6,17 @@ for more information.
...
 
+import Head from "next/head";
 import { Redirect, Route } from "react-router-dom";
-import { HydraAdmin, hydraDataProvider as baseHydraDataProvider, fetchHydra as baseFetchHydra, useIntrospection } from "@api-platform/admin";
+import { hydraDataProvider as baseHydraDataProvider, fetchHydra as baseFetchHydra, useIntrospection } from "@api-platform/admin";
 import parseHydraDocumentation from "@api-platform/api-doc-parser/lib/hydra/parseHydraDocumentation";
-import authProvider from "./authProvider";
+import authProvider from "utils/authProvider";
+import { ENTRYPOINT } from "config/entrypoint";
+import Login from "components/admin/Login";
```
alanpoulain, line 18: This one really doesn't exist, isn't it?

4c0n: You're right this one doesn't exist. Would you like me to add it to this doc? I haven't gotten it to work in my project yet though, so maybe it's better that I add it after that in case modifications are needed.

alanpoulain: It's only useful for the demo, I don't think we should have it in the doc.

alanpoulain (approved):

alanpoulain:
Thanks!

PR #1488, opened 2021-12-20
fix: [doc] Use the Mailer component instead of SwiftMailer

mdoutreluingne:
I corrected for this page, the component  it for sending email because SwiftMailer is no longer maintained. So I replaced by the component Mailer of Symfony.


alanpoulain (approved):

dunglas (approved):

dunglas:
Thanks!

PR #1496, opened 2022-01-24
fix(doc): path to EventPriorities.php

lobodol:
Hello there, just fixing path to `EventPriorities` that seems to have moved.


alanpoulain:
Thanks.

PR #1497, opened 2022-01-26
fix(doc): Suggest ApiResource attribute "validationGroups" for generator

mbrodala:
See https://github.com/api-platform/core/pull/4640


vinceAmstoutz:
Already [Closed here](https://github.com/api-platform/core/pull/4640#issuecomment-1064969160)

PR #1503, opened 2022-02-10
Docs: JWT needs user identity field

BurningDog:
If the user identity field is not specified, then although a JWT token is successfully issued, any requests with it result in a `401 Invalid credentials`. Given that elsewhere on the JWT docs the user identity field is specified as `email`, we need to change the JWT bundle config from the default of `username` to `email`.

This PR also adds some helpful info for how to decode a JWT token for debugging.

### DIFF
--- a/core/jwt.md
+++ b/core/jwt.md
@@ -45,6 +45,22 @@ environment, where you don't want to accidentally invalidate all your clients' t
 For more information, refer to [the bundle's documentation](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Resources/doc/index.md)
 or read a [general introduction to JWT here](https://jwt.io/introduction/).
 
+## Configuring LexikJWTAuthenticationBundle
+
+Once the bundle has been installed, a new file is created: `config/packages/lexik_jwt_authentication.yaml`. Adjust it as follows:
+
+```yaml
+lexik_jwt_authentication:
+    secret_key: '%env(resolve:JWT_SECRET_KEY)%'
+    public_key: '%env(resolve:JWT_PUBLIC_KEY)%'
+    pass_phrase: '%env(JWT_PASSPHRASE)%'
+    user_identity_field: email # key under which the user identity will be stored in the token payload
+```
+
+This configures the JWT to use the user's `email` as the identity field.
+
+The [full default configuration](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/2.x/Resources/doc/1-configuration-reference.md#full-default-configuration) may be of interest.
+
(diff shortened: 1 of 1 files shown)


alanpoulain:
Hello,
This is not what is done in the demo though: https://github.com/api-platform/demo/blob/main/api/config/packages/lexik_jwt_authentication.yaml.
Are you sure it is necessary?

BurningDog:
@alanpoulain oh, that's interesting. The demo works because of https://github.com/api-platform/demo/blob/main/api/src/Entity/User.php#L99-L105

However, I couldn't find any reference to setting up a User object like the demo one inside the API Platform core documentation. https://api-platform.com/docs/core/fosuser-bundle/#fosuserbundle-integration says to use https://symfony.com/doc/current/security/user_provider.html#entity-user-provider or https://symfony.com/doc/current/security/user_providers.html#creating-a-custom-user-provider

I followed the former, which uses `email` as the identifier. Did I miss where to set up the User object as it's been done in the demo?

alanpoulain:
If you used the `UserInterface` before Symfony 6, you needed to have a `getUsername` method. Since SF 6, it's `getUserIdentifier`, but it will be probably changed in LexikJWTAuthenticationBundle soon (see https://github.com/lexik/LexikJWTAuthenticationBundle/issues/881).

BurningDog:
The issue is that the [LexikJWTAuthenticationBundle sets the config value of `user_identity_field` to `username` by default](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/2.x/DependencyInjection/Configuration.php#L76-L78). The `User` entity then requires this function:

```php
    public function getUsername(): string
    {
        return (string) $this->email;
    }
```

This function is not mentioned at https://api-platform.com/docs/core/jwt/ or at https://symfony.com/doc/current/security.html#a-create-your-user-class which means that when trying to get a JWT authentication token by posting `email` and `password` to `/authentication_token` this exception is thrown:

```
Symfony\Component\PropertyAccess\Exception\NoSuchPropertyException

Can't get a way to read the property "username" in class "App\Entity\User".
```

So we either need to document `user_identity_field: email` in `config/packages/lexik_jwt_authentication.yaml` or that the `getUsername()` function needs to be added to the `User` entity.

BurningDog:
So do we wait for https://github.com/lexik/LexikJWTAuthenticationBundle/issues/881 to land, or update the API Platform docs in the meantime?

alanpoulain:
I think we can wait a little.

BurningDog:
Ok! I'll close this, then.

PR #1509, opened 2022-02-24
docs: add note on using PUT method to upload files

alcalyn:
I lost few hours on trying to figure why I had an empty filebag, using first or second file upload examples.

So I added a note.

### DIFF
--- a/core/file-upload.md
+++ b/core/file-upload.md
@@ -42,6 +42,9 @@ resource (in our case: `Book`).
 This example will use a custom controller to receive the file.
 The second example will use a custom `multipart/form-data` decoder to deserialize the resource instead.
 
+**Note**: Uploading files won't work in `PUT` or `PATCH` requests, you must use `POST` method to upload files.
+See [this issue on Symfony](https://github.com/symfony/symfony/issues/36146) talking of this behaviour.
+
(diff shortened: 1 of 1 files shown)


core/file-upload.md
```diff
@@ -42,6 +42,9 @@ resource (in our case: `Book`).
 This example will use a custom controller to receive the file.
 The second example will use a custom `multipart/form-data` decoder to deserialize the resource instead.
 
+**Note**: Uploading files won't work in `PUT` or `PATCH` requests, you must use `POST` method to upload files.
+See [this issue on Symfony](https://github.com/symfony/symfony/issues/36146) talking of this behaviour.
```
alanpoulain, line 46:
```suggestion
See [the related issue on Symfony](https://github.com/symfony/symfony/issues/9226) and [the related bug in PHP](https://bugs.php.net/bug.php?id=55815) talking about this behavior.
```

alcalyn: Oh, nice completion for the php bug, thanks !
At first I choose the issue 36146 because less old, and anyway it have a link to the oldest one.

alanpoulain: Since it was marked as duplicated, I prefer to use the main one.

PR #1511, opened 2022-03-01
Fixed entity name for subresourceOperations Using Serialization Groups

TekPike:

### DIFF
--- a/core/subresources.md
+++ b/core/subresources.md
@@ -133,7 +133,7 @@ You may want custom groups on subresources, you can set `normalization_context` 
 
 ```php
 <?php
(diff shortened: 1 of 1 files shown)


vincentchalamon:
Hi @TekPike, thanks for your contribution!

The declaratin of the subresource is done thanks to the `@ApiSubresource` annotation on the `Question::$answer` property. But its configuration (`normalization_context`, etc.) must be done on the target entity `Answer`.

We know it adds complexity, that's why we completely revamped the subresources: https://github.com/api-platform/core/blob/main/docs/adr/0000-subresources-definition.md

PR #1525, opened 2022-03-30
Add precision about behavior of get op. when identifier changes

welcoMattic:
Considering this code
```php
#[ApiProperty(identifier: false)]
public $id;

#[ApiProperty(identifier: true)]
public $code;
```

The get operation will now expect `code` value to retrieve the resource.

### DIFF
--- a/core/identifiers.md
+++ b/core/identifiers.md
@@ -174,6 +174,8 @@ final class Person
 }
 ```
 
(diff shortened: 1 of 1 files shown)


vinceAmstoutz (approved):
Thanks @welcoMattic!

PR #1532, opened 2022-04-15
Update caddy.md

mholt:
Not a big deal, just a closing backtick seems to be missing here. :man_shrugging:


distribution/caddy.md
```diff
@@ -31,5 +31,5 @@ For instance, if you want to route all requests to a path starting with `/admin`
         {header.Accept}.matches("\\btext/html\\b")
 -        && !{path}.matches("(?i)(?:^/docs|^/graphql|^/bundles/|^/_profiler|^/_wdt|\\.(?:json|html$|csv$|ya?ml$|xml$))")
 +        && !{path}.matches("(?i)(?:^/admin|^/docs|^/graphql|^/bundles/|^/_profiler|^/_wdt|\\.(?:json|html$|csv$|ya?ml$|xml$))")
```
alanpoulain, line 33: You also added `admin` here though.

mholt: I did not; it was already there. If you look at the refined diff, I only added the backtick at the end.

I think GitHub is erroneously rendering these lines of the patch block as part of _this_ patch, which didn't actually change.

dunglas:
Thanks Matt!

PR #1536, opened 2022-04-21
feat(doc): use snake case names on operation sub options

gnito-org:
It took me too much frustration and digging to accidentally discover this.


alanpoulain:
I think this is too much. Adding a note in `Configuring Operations` would be sufficient.

gnito-org:
I disagree. It's easier and more easily understandable to show by example than describe it in verbose text. The work is already done. It just needs merging. :smile:

alanpoulain:
You can show by example in `Configuring Operations` too.
> It just needs merging.

Sure, but it needs to be maintained too :wink: 

Anyway, this is my opinion, maybe @dunglas or someone else prefers to merge as is.

gnito-org:
From your last comment it sounds if the additional header was the primary concern. I removed the header and moved the text and example into the `Configuring Options` section.

vincentchalamon (changes requested):

PR #1537, opened 2022-04-22
feat(doc): how to incorporate static code tables in an API

gnito-org:

### DIFF
--- a/core/static-code-tables.md
+++ b/core/static-code-tables.md
@@ -53,7 +53,7 @@ class ItemType
         $this->description = $description;
     }
 
-    public static function createInstance(string $id): ?self
+    public static function createInstance(?string $id): ?self
     {
         $id = self::extractIdFromIri($id);
         if (!isset(self::$data[$id])) {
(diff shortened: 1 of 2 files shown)


dunglas:
Nice one! Couldn't you use the new `enum` type introduced in PHP 8.1 for this use case?

gnito-org:
I have not looked at the new `enum` feature in depth. From cursory reading it appears that an `enum` is static. How would one instantiate it so that API Platform can treat is as just another entity, with an IRI?

gnito-org:
@dunglas I added an `enum` section to the explanation.

core/static-code-tables.md
```diff
@@ -0,0 +1,579 @@
...
+<?php
+// api/src/Entity/ItemType.php
+
+namespace App\Entity;
+
+use ApiPlatform\Core\Annotation\ApiProperty;
+use ApiPlatform\Core\Annotation\ApiResource;
+use Symfony\Component\Serializer\Annotation\Groups;
+
+#[ApiResource(
```
vincentchalamon, line 213: Why creating an ApiResource on an enum?

PR #1553, opened 2022-05-26
Feat (Doc): Making existing data persisters resumeable using decoration

S1ructure:
Wouldnt it be the better way to combine the decoration approach to make an existing datapersister (here doctrines) resumeable?


vincentchalamon:
The DataPersister approach is supposed to be entity-specific. By decorating it, you don't know exactly which DataPersister you decorate (except if you manually specify the original one).

Also, the original DataPersister has a priority of -1000 to allow custom DataPersisters to be easily registered thanks to autowiring and autoconfiguration.

If you need to save the data + send it to Messenger, I think it's still a good approach to use a custom DataProvider, cause not all your entities need the same behavior.

PR #1555, opened 2022-06-03
Update testing.md: Change getContainer as function instead of variable

leeyuentuen:
Change getContainer as a function instead of variable


dunglas:
Thanks!

PR #1557, opened 2022-06-10
Removed - unused command

zackha:
docker-compose no longer uses the --no-cache command.


vincentchalamon:
`docker compose build` command still uses the `--no-cache` option: https://docs.docker.com/engine/reference/commandline/compose_build/

PR #1560, opened 2022-06-20
Fix custom filter example

KDederichs:
While trying to update to 2.7 I noticed that the example code in the custom filter section is wrong, so here's a small fix for that :)


vincentchalamon (approved):

vincentchalamon:
Thanks @KDederichs

PR #1562, opened 2022-06-21
Fix UI Template path

hanishsingla:


florent-andre:
+1
it's about swagger ui template path.

alanpoulain:
Solved in 2.6.

PR #1565, opened 2022-06-23
Fix JWT authentication test file with static container

thomas-fortin:
Fix an error thrown while launching the AuthenticationTest because of the use of `self::$container`. Assign it in a variable first fixes the issue.


dunglas (approved):

vincentchalamon:
Thanks @thomas-fortin

PR #1570, opened 2022-07-10
fix extends and filterPropery example for RegExpFilter

tacman:
Fixes https://github.com/api-platform/core/issues/4823


tacman:
Argh.  I wanted to do a PR against main, not against 2.6.

PR #1587, opened 2022-08-03
Update state-processor.md

Xusifob:
The service '@api_platform.doctrine.orm.state.persist_processor' no longer exists, but replacing it with "@ApiPlatform\Doctrine\Common\State\PersistProcessor" seems to be working in my local implementation

Maybe the service needs to be tagged and the documentation is correct

### DIFF
--- a/core/state-processors.md
+++ b/core/state-processors.md
@@ -132,7 +132,7 @@ services:
     # ...
     App\State\UserProcessor:
         bind:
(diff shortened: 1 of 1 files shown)


alanpoulain:
This new service name will be introduced in the next release.
See https://github.com/api-platform/core/issues/4815.

PR #1589, opened 2022-08-08
Update ex to use ApiPlatform\Metadata\ApiResource;

rwindelz:
Code example in documentation appears to use older `ApiPlatform\Core\Annotation\ApiResource`

Found Issue #4485 which noted a workaround to revert to `use ApiPlatform\Core\Annotation\ApiResource;`


alanpoulain:
You are in the 2.6 version, this is why you are viewing the old attributes. Please switch to the main (2.7) version.

PR #1591, opened 2022-08-09
Update minikube.md

italozaina:
This is the way that worked to me

### DIFF
--- a/deployment/minikube.md
+++ b/deployment/minikube.md
@@ -16,7 +16,11 @@ Finally, [install Helm](https://helm.sh/docs/intro/install/). We'll use it to de
 
 ## Building and Pushing Docker Images
 
-First, build the images:
+First, you need point your docker client to the VM's docker daemon(after minikube start) by running:
+
+    eval $(minikube docker-env)
+
+And now you can build the images:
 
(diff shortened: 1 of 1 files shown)


alanpoulain:
I think it would be better to refer to https://minikube.sigs.k8s.io/docs/handbook/pushing/ instead.

italozaina:
Already has a reference to the documentation("If you use Mac or Windows, [refer to the documentation](https://minikube.sigs.k8s.io/docs/handbook/registry/#docker-on-macos) to learn how to expose") in **the tutorial** that does not maintain a cohesive understanding of _step-by-step_, I believe that this way will **be more accurate**.

alanpoulain:
You are in the context of pushing a Docker image, it makes sense to refer to the corresponding documentation on the Minikube website instead of duplicating its content.
It's also easier to maintain it, since it's not our responsibilities to update the documentation if Minikube changes its way of doing it.

vinceAmstoutz:

Unfortunately, we must close this PR because the documentation currently has a redirection to the Minikube doc and your proposal, as you can see [here](https://api-platform.com/docs/deployment/minikube/#building-and-pushing-docker-images).

PR #1599, opened 2022-08-30
Added example in custom filter

zzAIMoo:
Added an example in the Custom Filter section of the documentation to show how to use multiple properties in a custom filter

### DIFF
--- a/core/filters.md
+++ b/core/filters.md
@@ -23,16 +23,16 @@ to a Resource in two ways:
 
 1. Through the resource declaration, as the `filters` attribute.
 
-   For example having a filter service declaration:
+    For example having a filter service declaration:
 
     ```yaml
     # api/config/services.yaml
     services:
         # ...
         offer.date_filter:
-            parent: 'api_platform.doctrine.orm.date_filter'
-            arguments: [ { dateProperty: ~ } ]
-            tags:  [ 'api_platform.filter' ]
+            parent: "api_platform.doctrine.orm.date_filter"
+            arguments: [{ dateProperty: ~ }]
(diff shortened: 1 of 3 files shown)


core/filters.md
```diff
@@ -1277,6 +1277,43 @@ class Offer
...
+    public string $name;
+    public int $cost;
+}
+```
+These properties can then be accessed in the custom filter like this:
+```php
+//App/Filter/CustomAndFilter.php
+protected function filterProperty(string $property, $value, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null) {  
+  $rootAlias = $queryBuilder->getRootAliases()[0];  
+  foreach ($this->getProperties() as $prop => $ignore) { //NOTE: we put ignore because we want the key name, not the value inside which will be null
```
alanpoulain, line 1305: The `getProperties` method seems to already retrieve only the keys.

alanpoulain:
Could you only keep your commit please?

core/filters.md
zzAIMoo: it returns an associative array which has the properties as keys and null as the value of the keys, so using => $ignore i give the null value to the variable $ignore and the key value to $prop, now that i think about it it's better to use array_keys($this->getProperties()) so that there's no need for a new variable

alanpoulain: You're right, I was in the ES filter...
Yes please use `array_keys`.

zzAIMoo:
(closed because of problems with vscode, reopening new one)

PR #1600, opened 2022-08-30
Added example in custom filter

zzAIMoo:
Added an example in the Custom Filter section of the documentation to show how to use multiple properties in a custom filter

### DIFF
--- a/core/filters.md
+++ b/core/filters.md
@@ -1269,6 +1269,43 @@ class Offer
 }
 ```
 
+When creating a custom filter you can specify multiple properties of a resource using the usual filter syntax:
+```php
+<?php
+// api/src/Entity/Offer.php
+
+namespace App\Entity;
+
+use ApiPlatform\Core\Annotation\ApiFilter;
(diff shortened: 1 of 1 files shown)


core/filters.md
```diff
@@ -1269,6 +1269,43 @@ class Offer
...
+{
+    // ...
+    public string $name;
+    public int $cost;
+}
+```
+These properties can then be accessed in the custom filter like this:
+```php
+//App/Filter/CustomAndFilter.php
+protected function filterProperty(string $property, $value, QueryBuilder $queryBuilder, QueryNameGeneratorInterface $queryNameGenerator, string $resourceClass, string $operationName = null) {
```
alanpoulain, line 1295: `$operationName` doesn't exist anymore in `main`, does it?

zzAIMoo: yeah didn't notice, changing it right now

alanpoulain (approved):

alanpoulain:
Thanks!

PR #1609, opened 2022-09-20
Fix CurrentUserExtension example

121593:
In the example, the methods `applyToCollection` and `applyToItem` implements incorrectly their definitions :

* [QueryCollectionExtensionInterface](https://github.com/api-platform/core/blob/3.0/src/Doctrine/Orm/Extension/QueryCollectionExtensionInterface.php)
* [QueryItemExtensionInterface](https://github.com/api-platform/core/blob/3.0/src/Doctrine/Orm/Extension/QueryItemExtensionInterface.php)

The problem may also exists on v2.7 documentation

### DIFF
--- a/core/extensions.md
+++ b/core/extensions.md
@@ -73,12 +73,12 @@ final class CurrentUserExtension implements QueryCollectionExtensionInterface, Q
         $this->security = $security;
     }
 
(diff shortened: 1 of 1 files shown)


alanpoulain (approved):

alanpoulain:
Thanks!

PR #1614, opened 2022-09-26
Update state-processors.md

kconde2:
I got error using @api_platform.doctrine.orm.state.persist_processor because it doesn't exist

### DIFF
--- a/core/state-processors.md
+++ b/core/state-processors.md
@@ -132,7 +132,7 @@ services:
     # ...
     App\State\UserProcessor:
         bind:
-            $decorated: '@api_platform.doctrine.orm.state.persist_processor'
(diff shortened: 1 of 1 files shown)


alanpoulain:
The service name exists in 3 but not in 2.7: https://github.com/api-platform/core/pull/4859

kconde2:
> The service name exists in 3 but not in 2.7: [api-platform/core#4859](https://github.com/api-platform/core/pull/4859)

Thanks for your reply.

I understand but this is the documentation for 2.7 right?

Is the idea that everything that works for v3 will also work for 2.7 in the future?

In any case, it made me waste time looking for the right service.

soyuka:
will be fixed in the next patch

PR #1620, opened 2022-10-02
Update state-providers.md

Mathieu33260:
Specify allowed operations for state providers.


core/state-providers.md
```diff
@@ -17,6 +17,8 @@ In the following examples we will create custom state providers for an entity cl
 Note, that if your entity is not Doctrine-related, you need to flag the identifier property by using
 `#[ApiProperty(identifier: true)` for things to work properly (see also [Entity Identifier Case](serialization.md#entity-identifier-case)).
 
+Also note that state providers only apply `Get` and `GetCollection` operations.
```
dunglas, line 20: They are called for PATCH and PUT too.

PR #1624, opened 2022-10-10
Document the new release process announced at API Platform Conference

Deuchnord:
On 15 Sept., 2022, a new release process has been announced starting from API Platform 3.0.
This PR provides some adjustments to the documentation to reflect those changes.

~~Remains to do: maybe a graph similar to the one we can see on [Symfony release page](https://symfony.com/releases#symfony-releases-calendar) to give a clear view of the process?~~

### DIFF
--- a/extra/releases.md
+++ b/extra/releases.md
@@ -1,6 +1,19 @@
 # The Release Process
 
 API Platform follows the [Semantic Versioning](https://semver.org) strategy.
+A new minor version is released every six months, and a new major version is released every two years, along with a last minor version on the previous major one with the same features and an upgrade path.
+
+For example:
+
+- version 3.0 has been released on 15 September, 2022;
+- version 3.1 will be released on March, 2023;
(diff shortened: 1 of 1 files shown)


alanpoulain:
I don't think we should add a graph: the Symfony one seems to be generated automatically. In our case we don't want to maintain an image file.

Deuchnord:
> I don't think we should add a graph: the Symfony one seems to be generated automatically. In our case we don't want to maintain an image file.

I was thinking about a static image with an example of roadmap (e.g. API P 3.0-3.4 + 4.0), that would not change with when new versions are released. But maybe the text is enough?

alanpoulain:
I think it would perturb the user if it is not updated with the latest versions each time a new version is out.

Deuchnord:
Sound fair, no graph then :)

alanpoulain:
Thanks @Deuchnord.

PR #1631, opened 2022-10-20
Minor grammatical fixes in subresources.md

warddave:
Minor grammatical fixes.

### DIFF
--- a/core/subresources.md
+++ b/core/subresources.md
@@ -8,7 +8,7 @@ Subresources work very well by implementing your own state [providers](./state-p
 or [processors](./state-processors.md). In API Platform we provide a working Doctrine layer for
 subresources providing you add the correct configuration for URI Variables.
 
-## URI Variables Configuration
+## URI Variable Configuration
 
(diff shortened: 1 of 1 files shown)


core/subresources.md
```diff
@@ -8,7 +8,7 @@ Subresources work very well by implementing your own state [providers](./state-p
 or [processors](./state-processors.md). In API Platform we provide a working Doctrine layer for
 subresources providing you add the correct configuration for URI Variables.
 
-## URI Variables Configuration
+## URI Variable Configuration
```
alanpoulain, line 11: Why this change?

warddave: `Variable configuration` is a compound noun and in the case of a compound noun to make it plural, the last word uses the plural form and the first word uses the singular form..

e.g it's `Dog parks` not `dogs park`

So it'd be either `Variable configuration` (singular) or `Variable configurations` (plural).

But, in this case `variable configuration` is  slightly more natural because there is one type of configuration for the variables and the title refers to that one type of configuration, not the individual variables. 

`variable configurations` wouldn't be entirely incorrect either, but it's slightly awkward.

 `variables configuration` is definitely incorrect.

I'm a native english speaker 😄  - I had to think through the why for a second - it just sounded incorrect.

PR #1633, opened 2022-10-28
Prevent "window is not defined" when using auth

gplanchat:
As the `dataProvider` from Hydra uses the `window` object, we need to provide a way to make things work during SSR where this object does not exist.


alanpoulain:
The `admin` should not be executed server-side, so I don't think this change is good.

gplanchat:
Ok, I forgot this part when migrating from version 2.6 https://github.com/api-platform/demo/blob/main/pwa/pages/admin/index.tsx#L3-L5

PR #1637, opened 2022-10-31
XML examples for subresources configuration

yivi:
Added some XML examples for subresource configuration, so users not configuring resource mapping with attributes can get an easier understanding of this.


alanpoulain (approved):

alanpoulain:
Thank you @yivi.

PR #1638, opened 2022-10-31
Fix documentation for using a service to generate validation groups

mark-gerarts:
The documentation says the following annotation should be used if you want to dynamically set validation groups using a service:

```php
#[ApiResource(validationContext: ['groups' => [AdminGroupsGenerator::class]])
```

However, this does not work. The API Platform validator only tries to fetch a service when a string is passed as `groups`:

```
if (
    $this->container &&
    \is_string($validationGroups) &&
    $this->container->has($validationGroups) &&
    ($service = $this->container->get($validationGroups)) &&
    \is_callable($service)
) {
```


alanpoulain:
Thank you @mark-gerarts.

PR #1642, opened 2022-11-02
Fix invalid PHP example

aschempp:
I just noticed this bug while copying the code to my project 🙃


alanpoulain:
Thanks :slightly_smiling_face:

PR #1647, opened 2022-11-09
fix(controllers): replace "path" by "uriTemplate" attribute

nxtpge:
The documentation related to the version 2.7, 3.0 and 3.1 describing the custom operation configuration example contains a version 2.6 mandatory attribute ("path" instead of "uriTemplate").


alanpoulain:
Thanks!

PR #1652, opened 2022-11-10
fix(operations): replace entity "Book" by "User" for .yaml and .xml e…

nextpageblog:
…xamples

The .yaml and .xml examples for defining which operation to use to generate the IRI contain previous versions configurations ("App\Entity\Book" instead of "App\Entity\User").


alanpoulain:
Thanks!

PR #1653, opened 2022-11-10
fix(controllers): replace "path" by "uriTemplate" attribute

nxtpge:
The documentation related to the version 3.0 and 3.1 describing the custom operation configuration example contains a version 2.6 mandatory attribute ("path" instead of "uriTemplate").


alanpoulain:
Already merged in 2.7?

alanpoulain:
Exactly, it will be done later.

PR #1655, opened 2022-11-14
feat(controllers): using the PlaceholderAction

nxtpge:
The documentation leads to create a custom controller/action in order to create a custom operation.

However, API Platform provides an undocumented alternative: using the `ApiPlatform\Action\PlaceholderAction` which prevents to create a custom controller in order to create a custom operation.

Indeed, in complex cases, it prevents ending with a codebase overwhelmed by custom controllers which only exist because of the existence of the mandatory `controller` attribute.

In the long term, it could be a good idea for the Developer Experience to make this attribute optional so that API Platform uses the `ApiPlatform\Action\PlaceholderAction` by default.

### DIFF
--- a/core/controllers.md
+++ b/core/controllers.md
@@ -157,7 +157,7 @@ Complex use cases may lead you to create multiple custom operations.
 
 In such a case, you will probably create the same amount of custom controllers while you may not need to perform custom logic inside.
 
-To avoid that, API Platform provides the `ApiPlatform\Action\PlaceholderAction` which behaves the same when using the [built-in operations](https://api-platform.com/docs/core/operations/#operations).
+To avoid that, API Platform provides the `ApiPlatform\Action\PlaceholderAction` which behaves the same when using the [built-in operations](operations.md#operations).
 
 You just need to set the `controller` attribute with this class. Here, the previous example updated:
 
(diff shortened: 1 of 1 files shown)


dunglas (approved):
I wonder if we couldn't automate this in the core library directly (if no controller is set, use `PlaceholderAction` by default).

core/controllers.md
```diff
@@ -151,6 +151,74 @@ App\Entity\Book:
 It is mandatory to set the `method`, `path` and `controller` attributes. They allow API Platform to configure the routing path and
 the associated controller respectively.
 
+## Using the PlaceholderAction
+
+Complex use cases may lead you to create multiple custom operations.
+
+In such a case, you will probably create the same amount of custom controllers while you may not need to perform custom logic inside.
+
+To avoid that, API Platform provides the `ApiPlatform\Action\PlaceholderAction` which behaves the same when using the [built-in operations](https://api-platform.com/docs/core/operations/#operations).
```
alanpoulain, line 160:
```suggestion
To avoid that, API Platform provides the `ApiPlatform\Action\PlaceholderAction` which behaves the same when using the [built-in operations](operations.md#operations).
```

alanpoulain:
Would you like to make the PR in the core @nxtpge?

alanpoulain:
Alright then! You can open the PR and ask questions on it if needed.

PR #1657, opened 2022-11-15
getting started: better Docker commands

dunglas:

### DIFF
--- a/distribution/index.md
+++ b/distribution/index.md
@@ -73,13 +73,13 @@ Similarly, on Windows, only [Docker for Windows](https://docs.docker.com/docker-
 Open a terminal, and navigate to the directory containing your project skeleton. Run the following command to start all
 services using [Docker Compose](https://docs.docker.com/compose/):
 
-Download and build the latest versions of the images:
+Download the latest versions of the images:
 
 ```console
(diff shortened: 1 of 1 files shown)


distribution/index.md
```diff
@@ -73,13 +73,13 @@ Similarly, on Windows, only [Docker for Windows](https://docs.docker.com/docker-
...
-Download and build the latest versions of the images:
+Download the latest versions of the images:
 
 ```console
-docker compose build --pull --no-cache
+docker compose pull --include-deps
 ```
 
-Start Docker Compose in detached mode:
+Then build images and Start Docker Compose in detached mode:
```
alanpoulain, line 82: Why removing the `build` command?

dunglas: It's automatically done when you run `docker compose up` (if necessary).

alanpoulain: Then maybe we should remove this part of the sentence?

dunglas: Why? It does build :)

dunglas: The instructions are good and clear IMHO

alanpoulain: As you want.

PR #1658, opened 2022-11-16
Update index.md

Mano-Lis:
Specifying type:'datetime' for $publicationDate causes a denormalization failure. Removing it or replacing it with 'datetime_immutable' fixes the issue.


dunglas:
Thank you!

PR #1660, opened 2022-11-21
fix property name for displaying related resource name instead of iri

PawelSuwinski:
It seems that this does not work with new metadata system:
```
#[ApiProperty(types: ["https://schema.org/name"])]
private $name;
```
but this one works:
```
#[ApiProperty(iris: ["https://schema.org/name"])]
private $name;
```
https://api-platform.com/docs/admin/schema.org/#displaying-related-resources-name-instead-of-its-iri


alanpoulain:
Shouldn't `iris` not exist anymore @soyuka?

PR #1663, opened 2022-11-24
Document requirements for resources that are only acessible as a subresource

mergeMarc:
I recently ran into a limitation when referencing a resource that is only accessible as a subresource (api-platform/core#5108).
This PR adds a section about the requirements when using subresources in this way.

### Todos
- [ ] \(Optional) Add example code if desired.

### DIFF
--- a/core/subresources.md
+++ b/core/subresources.md
@@ -296,3 +296,9 @@ class Company {
     // ...
 }
 ```
+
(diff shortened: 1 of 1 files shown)


mergeMarc:
I can no longer maintain this PR.
I still think this addition would help people.

PR #1670, opened 2022-12-05
Typo: Removed quotes around XxxException:class constants

gplanchat:
For some reason quotes were applied around `ProductNotFoundException::class` and `ProductWasRemovedException::class` in the docs.


alanpoulain:
Thanks!

PR #1674, opened 2022-12-08
Update container directory for volume mount

MaSpeng:
The [documentation](https://hub.docker.com/_/mongo) of the MongoDB docker image points out that the DB is stored at `/data/db`, described in the **Caveats** section under **Where to Store Data**.
> The -v /my/own/datadir:/data/db part of the command mounts the `/my/own/datadir` directory from the underlying host system as `/data/db` inside the container, where MongoDB by default will write its data files.


dunglas (approved):

dunglas:
Thanks!

PR #1692, opened 2023-01-03
feat(subresource): Link security

KDederichs:
Docs PR for Link Security feature

### DIFF
--- a/core/subresources.md
+++ b/core/subresources.md
@@ -242,3 +242,26 @@ class Company {
     // ...
 }
 ```
+
+## Security
+
+In order to use Symfony's built-in security system on subrecources the security option of the `Link` attribute can be used.
+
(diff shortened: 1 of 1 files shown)


Enz000:
it can be interesting if you use an XML / Yaml format 😁

core/subresources.md
```diff
@@ -242,3 +242,28 @@ class Company {
     // ...
 }
 ```
+
+## Security
+
+In order to use Symfony's built-in security system on subrecources the security option of the `Link` attribute can be used.
```
nesl247, line 248:
```suggestion
In order to use Symfony's built-in security system on subresources the security option of the `Link` attribute can be used.
```

KDederichs:
Looking at this:
The `enable_link_security` config that was added in the end is missing, I'll PR it in in a bit.

PR #1698, opened 2023-01-11
Replace "informations" with "information"

andytait:
In this context, "information" should not be pluralised.

Some nouns do not have a plural form. "Information" is one of them. Information is a [mass noun](https://en.wikipedia.org/wiki/Mass_noun). It has no plural form.

Additionally, CI flagged that "CLI" should be replaced with "command-line". This is resolved too.


alanpoulain:
Could you fix this to make the CI happy please?
image

alanpoulain:
Thank you!

PR #1426, opened 2021-09-07
fix(2.7 upgrade): text inconsistencies

PierreThibaudeau:
_api_item_operation_name => _api_operation_name
snake_case to camelCase
removes 'collectionOperation' and 'itemOperation' references.
Replaces some texts to mathc the new behavior

PR #1702, opened 2023-01-23
Fix: Remove duplicated docs and fix constructors

timkrase:
The docs page for operations contains a part that's a duplicate from the controllers page and the the controllers page is not using the new constructor syntax for all code examples.
@alanpoulain fixed both of this with #1439 but #1426 (which was created earlier and contained changes in that part) seems to be merged afterwards and probably reintroduced the problem. 

This PR removes the duplicated part from the operations page again (which is still available at [the controllers doc page](https://api-platform.com/docs/core/controllers/#alternative-method) and changes the constructor on the controller docs page.

### DIFF
--- a/core/controllers.md
+++ b/core/controllers.md
@@ -44,12 +44,9 @@ use Symfony\Component\HttpKernel\Attribute\AsController;
 #[AsController]
 class CreateBookPublication extends AbstractController
 {
-    private $bookPublishingHandler;
-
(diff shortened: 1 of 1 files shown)


alanpoulain (approved):

alanpoulain:
It keeps reappearing! Thank you @timkrase.
Merged in 3.1 (available by default in the doc soon).

PR #1713, opened 2023-02-08
plainPassoword validator needs validationContext

PawelSuwinski:
```
class User
(...)
    #[Assert\NotBlank(groups: ['user:create'])]
    #[Groups(['user:create', 'user:update'])]
    private ?string $plainPassword = null;

```

Per group  validator needs proper validationContext set to make it work.


alanpoulain (approved):

alanpoulain:
Thanks @PawelSuwinski.

PR #1716, opened 2023-02-11
Fix Typo in API-Platform Command

Jimmerz28:
Now the command will properly work when copy/pasting.


Jimmerz28:
Interpreted the command incorrectly, closing.

PR #1712, opened 2023-02-07
setPlainPassword argument name typo

PawelSuwinski:
Fix argument typo.

PR #1717, opened 2023-02-13
include default validation group

PawelSuwinski:
I realized after time that default validation group must be given explicitly when validationContext is used, so this add-on (#1712) PR for completeness.


alanpoulain (approved):

alanpoulain:
Maybe using `Constraint::DEFAULT_GROUP` would have been better though.

PR #1725, opened 2023-02-22
Note about PUT/PATCH in introduction §

Nayte91:
I didn't see the note and spent a lot of time figuring out what happens, because I was trying to Upload to an Existing Resource, and the note is not in this chapter. So I think it deserves to belong in the introduction paragraph.

### DIFF
--- a/core/file-upload.md
+++ b/core/file-upload.md
@@ -8,6 +8,9 @@ recommended you [read the documentation of
 VichUploaderBundle](https://github.com/dustin10/VichUploaderBundle/blob/master/docs/index.md)
 before proceeding. It will help you get a grasp on how the bundle works, and why we use it.
 
+**Note**: Uploading files won't work in `PUT` or `PATCH` requests, you must use `POST` method to upload files.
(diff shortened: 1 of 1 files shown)


alanpoulain:
I'm not sure changing its place would solve the issue. IMO the issue is more "graphical": we don't have a way to highlight a note like the Symfony documentation.

Nayte91:
By my experience, yesterday evening : 
- Searched how to upload a file without exposing a new resource and found [this page](https://api-platform.com/docs/core/file-upload/),
- Went directly to the "[Uploading to an Existing Resource with its Fields](https://api-platform.com/docs/core/file-upload/#uploading-to-an-existing-resource-with-its-fields)" chapter without reading the "[Uploading to a Dedicated Resource](https://api-platform.com/docs/core/file-upload/#uploading-to-a-dedicated-resource)" one,
- Tried to achieve it and test with some PATCH requests on my existing Entity,
- Table flipped after 2 hours,
- Looked at it today,
- Found this very important note in the other chapter.

It's pretty straightforward, and as I don't think I made major mistake by skipping the optional part, I argue that this note _must_ be in the introductory remarks rather than in a skippable section.

But I also agree with the fact that docs may need an icon for important notes :)

alanpoulain:
Alright :slightly_smiling_face:

PR #1728, opened 2023-02-27
Adding a link to the React Admin Symfonycasts video

weaverryan:
Hi!

An updated API Platform tutorial is now on Symfonycasts for API Platform 3 - https://symfonycasts.com/screencast/api-platform

All the existing links in the docs that pointed to the old version of this tutorial work just fine: the new screencast uses the same URLs. We did add one new video showing how to bootstrap the React Admin, which might merit a link :).

When Ep2 is finished, I'll do the same audit - that has at least one new topic (state processors) that we can add a link to.

Thanks!


dunglas:
Thanks Ryan!

PR #1733, opened 2023-03-11
Update elasticsearch documentation

dannyvw:
Mapping and one index for different models is deprecated, so removed it from the documentation. Also updated the ApiResource attribute.

Related to https://github.com/api-platform/core/pull/5450


soyuka:
awesome thanks @dannyvw !

PR #1739, opened 2023-04-05
Sort Post arguments

JabriAbdelilah:


alanpoulain:
Why?

JabriAbdelilah:
> 

To respect the method signature and get rid of the IDE warning.
image

soyuka:
we don't support positional arguments on metadata.

PR #1740, opened 2023-04-12
Fixed swagger_definition_name to openapi_definition_name

virtualize:
When changing the name of a definition the newer `openapi_definition_name` key should be used, as `swagger_definition_name` does not seem to work when using api-platform/core 3.*


alanpoulain (approved):

alanpoulain:
Thank you.

PR #1742, opened 2023-04-16
Fix docker compose command

Huppys:
Change "docker compose exec" to "docker compose run" in Testing documenation


alanpoulain:
Why?

Huppys:
I did miss the point of running the docker service while development to be able to run ``docker compose exec ...``.

TL;DR:
Sorry. I was wrong.

PR #1746, opened 2023-04-24
Typo in Subresources page

franckranaivo:
"things" instead of "thinks" from  the sentence "and if we want thinks to work".


alanpoulain (approved):

alanpoulain:
Thanks.

PR #1747, opened 2023-04-28
Fix Hoppscotch broken URL

edamov:
Replace https://docs.hoppscotch.io/features (404 Not found) with https://docs.hoppscotch.io/documentation/features/rest-api-testing/


dunglas:
Thanks!

PR #1751, opened 2023-05-03
Add information on how to get a user in a provider

RobQuistNL:
As this is the most basic usecases for using a provider ever (showing entities that only relate to the current user), adding it to the docs would have saved me a lot of time.

There's way more docs to be updated - nothing seems to even remotely indicate some kind of ACL / User related data filtering.

Isn't it one of the most basic usecases ever? Where you want someone to only see their own data?

This usecase doesn't seem to be documented anywhere. It doesn't seem doable by using the annotations (albeit there's security and IRI overrides everywhere) it would never just shows this users' own stuff.

### DIFF
--- a/core/state-providers.md
+++ b/core/state-providers.md
@@ -125,6 +125,40 @@ use App\State\BlogPostProvider;
 class BlogPost {}
 ```
 
+## Getting the user in a provider
+
+Receiving a user in your provider to allow for filtering of entities can be achieved by adding the `Security` class in your constructor;
+
+```php
+<?php
+
+namespace App\State;
+
+use ApiPlatform\Metadata\CollectionOperationInterface;
+use ApiPlatform\Metadata\Operation;
+use ApiPlatform\State\ProviderInterface;
+use App\Repository\BlogPostRepository;
+use Symfony\Bundle\SecurityBundle\Security;
+
+class BlogPostProvider implements ProviderInterface
+{
+    public function __construct(
+        private readonly BlogPostRepository $blogPostRepository,
+        private readonly Security $security
+    ) {
+    }
+
+    public function provide(Operation $operation, array $uriVariables = [], array $context = []): object|array|null
+    {
+        if ($operation instanceof CollectionOperationInterface) {
+            return $this->blogPostRepository->findAllForUser($this->security->getToken()->getUser());
+        }
+
+        return $this->blogPostRepository->find($uriVariables['id']);
+    }
+}
+```
+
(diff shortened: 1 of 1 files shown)


core/state-providers.md
```diff
@@ -125,6 +125,40 @@ use App\State\BlogPostProvider;
 class BlogPost {}
 ```
 
+## Getting the user in a provider
```
dunglas, line 128:
```suggestion
## Getting The Currently Connected User in a Provider
```

core/state-providers.md
```diff
@@ -125,6 +125,40 @@ use App\State\BlogPostProvider;
...
+
+namespace App\State;
+
+use ApiPlatform\Metadata\CollectionOperationInterface;
+use ApiPlatform\Metadata\Operation;
+use ApiPlatform\State\ProviderInterface;
+use App\Repository\BlogPostRepository;
+use Symfony\Bundle\SecurityBundle\Security;
+
+class BlogPostProvider implements ProviderInterface
```
dunglas, line 143:
```suggestion
final class BlogPostProvider implements ProviderInterface
```

core/state-providers.md
```diff
@@ -125,6 +125,40 @@ use App\State\BlogPostProvider;
...
+    ) {
+    }
+
+    public function provide(Operation $operation, array $uriVariables = [], array $context = []): object|array|null
+    {
+        if ($operation instanceof CollectionOperationInterface) {
+            return $this->blogPostRepository->findAllForUser($this->security->getToken()->getUser());
+        }
+
+        return $this->blogPostRepository->find($uriVariables['id']);
```
dunglas, line 157: Maybe should this be removed as this doesn't look related?

Also, you should set the `Vary` HTTP header (on the ApiResource attribute) to prevent any cache pollution issue.

dunglas:
This is only useful if you don't use Doctrine (or at least the built-in Doctrine provider), otherwise you should use an extension instead: https://api-platform.com/docs/core/extensions/

Could you make it clear in this page? Thanks for improving our docs!

RobQuistNL:
:exploding_head: it's called extensions...... (I am using Doctrine)

Okay, I had no idea. I found "filters" and "subresources" and "state providers" but I did _not_ look at "extensions" for this type of filtering / ownership logic...

Thanks for the quick review, I'll run through this stuff again and edit the docs.

vinceAmstoutz:
@RobQuistNL Thanks so much for your PR and the time you invested in contributing to this project! We truly appreciate your effort and involvement.

PR #1752, opened 2023-05-03
Update security.md

RobQuistNL:
Turns out you can!

~Edit: also turns out the documentation code is broken!~
Edit edit: Turns out I was wrong about it being broken - I had `security: 'is_granted("BOOK_CREATE", object)'` instead of `securityPostDenormalize: 'is_granted("BOOK_CREATE", object)'` in my code. :facepalm:

### DIFF
--- a/core/security.md
+++ b/core/security.md
@@ -241,11 +241,11 @@ class BookVoter extends Voter
 
     protected function supports($attribute, $subject): bool
     {
-        if ($attribute == 'BOOK_LIST' && $subject === null) {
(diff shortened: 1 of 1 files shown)


dunglas:
Both examples aren't achieving the same thing. You're allowing the current user to list all entries, even if belonging to another user. Filters allow to filter the collection content to retrieve only entries owned by an user.

RobQuistNL:
With filters you mean extensions, right? I'll edit the docs to redirect people looking for this stuff to the extensions.

The LIST example is indeed purely just the action of listing. But I now understand that its fine to allow someone to always list, we can then use those extensions to "filter" out any records.

I'll edit the docs accordingly!

Issue #1694, opened 2023-01-03
Recommend Foundry instead of Alice


dunglas:
Update the testing guide to recommend Foundry, which a more modern and pleasant solution, and is now officially recommended by Symfony.

Related to #1693

RobQuistNL:
Actually, even including `alice` now breaks the testsuite;
https://github.com/api-platform/api-platform/issues/2437

PR #1699, opened 2023-01-20
Add documentation to use Zenstruck Foundry instead of Alice

loic425:

PR #1756, opened 2023-05-10
docs(testing): recommend foundry instead of alice

Romaixn:
Related to #1694.


dunglas:
We already merged this change but unfortunately in 3.0: https://github.com/api-platform/docs/pull/1699

PR #1763, opened 2023-06-01
fix: LexikJWTAuthenticationBundle isn't supported out of the box anymore

divine:
After upgrading [LexikJWTAuthenticationBundle](https://github.com/lexik/LexikJWTAuthenticationBundle) to latest version [v2.19.0](https://github.com/lexik/LexikJWTAuthenticationBundle/releases) it's not supported anymore.

Login endpoint just disappeared and I needed to manually add the configuration.

See @ https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1119


alanpoulain:
If you use https://github.com/api-platform/core/pull/5609, does it work?

PR #1768, opened 2023-06-08
Fix typo in security.md

jay-low:
**Original**: It means *than* for `PUT` or `PATCH` requests, ...
**Fixed**: It means *that* for `PUT` or `PATCH` requests, ...


soyuka:
thanks

PR #1793, opened 2023-08-10
fix(distribution): replace dead links

Romaixn:
Replaced all the dead links reported by [Lychee](https://github.com/lycheeverse/lychee) with the new existing links.

### DIFF
--- a/distribution/index.md
+++ b/distribution/index.md
@@ -50,7 +50,7 @@ API Platform uses these model classes to expose and document a web API having a 
 * and basically everything needed to build modern APIs.
 
 One more thing, before we start: as the API Platform distribution includes [the Symfony framework](https://symfony.com),
-it is compatible with most [Symfony bundles](https://flex.symfony.com)
+it is compatible with most [Symfony bundles](https://symfony.com/bundles)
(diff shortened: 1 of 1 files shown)


distribution/index.md
```diff
@@ -722,7 +722,7 @@ You can also try things a bit more complex:
 ```
 
 The GraphQL implementation supports [queries](https://graphql.org/learn/queries/), [mutations](https://graphql.org/learn/queries/#mutations),
-[100% of the Relay server specification](https://facebook.github.io/relay/docs/en/graphql-server-specification.html), pagination,
+[100% of the Relay server specification](https://relay.dev/docs/guides/graphql-server-specification/), pagination,
 [filters](../core/filters.md) and [access control rules](../core/security.md).
 You can use it with the popular [RelayJS](https://facebook.github.io/relay/) and [Apollo](https://www.apollographql.com/docs/react/)
```
dunglas, line 727: https://relay.dev here too?

Romaixn: https://facebook.github.io/relay seems to redirect correctly to https://relay.dev. 
Indeed, I will modify it just in case the redirection stops working one day

PR #1794, opened 2023-08-11
doc: replace the old relay url with the new one

Romaixn:
Replace https://facebook.github.io/relay by https://relay.dev


dunglas:
Thanks!

Issue #994, opened 2019-11-24
Missing documentation to register a new encoder


matt6697:
I'm trying to register a new encoder following the [https://api-platform.com/docs/core/content-negotiation/](url) **Supporting custom format** example.

However, I get a `"Serialization for the format myformat is not supported"`when accessing API Platform after having followed the documented example.

```
{
  "type": "https://tools.ietf.org/html/rfc2616#section-10",
  "title": "An error occurred",
  "detail": "Serialization for the format myformat is not supported",
  "trace": [
    {
      "namespace": "",
      "short_class": "",
      "class": "",
      "type": "",
      "function": "",
      "file": "/mnt/c/Users/matthieu/Documents/GitHub/SPMetadataParser/vendor/symfony/serializer/Serializer.php",
      "line": 121,
      "args": []
    },
    {
      "namespace": "Symfony\\Component\\Serializer",
      "short_class": "Serializer",
      "class": "Symfony\\Component\\Serializer\\Serializer",
      "type": "->",
      "function": "serialize",
      "file": "/mnt/c/Users/matthieu/Documents/GitHub/SPMetadataParser/vendor/api-platform/core/src/EventListener/SerializeListener.php",
      "line": 119,
      "args": [
        [
          "object",
          "ApiPlatform\\Core\\Api\\Entrypoint"
        ],
        [
          "string",
          "myformat"
        ],
        [
          "array",
          []
        ]
      ]
    },
[...]
```

**Step 1 : Created a custom encoder class following the [https://symfony.com/doc/current/serializer/custom_encoders.html](url) documentation**
For the test and to match the reference documentation, the encoder returns a yaml content.

It is not clear at all in the documentation how to link the **myformat** new format with the `MyFormatEncoder` class. I used the `const FORMAT = 'myformat';` to miror how the XML encoder class is created, but I might be wrong.

```
<?php
namespace App\Serializer\Encoder;

use Symfony\Component\Serializer\Encoder\DecoderInterface;
use Symfony\Component\Serializer\Encoder\EncoderInterface;
use Symfony\Component\Yaml\Yaml;

class MyFormatEncoder implements EncoderInterface, DecoderInterface
{
    const FORMAT = 'myformat';

    public function encode($data, $format, array $context = [])
    {
        return Yaml::dump($data);
    }

    public function supportsEncoding($format)
    {
        return 'yaml' === $format;
    }

    public function decode($data, $format, array $context = [])
    {
        return Yaml::parse($data);
    }

[...]

PR #1797, opened 2023-08-30
docs(content-negociation): document how to declare an encoder for custom format

nikophil:
fixes https://github.com/api-platform/docs/issues/994

### DIFF
--- a/core/content-negotiation.md
+++ b/core/content-negotiation.md
@@ -196,6 +196,19 @@ api_platform:
         myformat: ['application/vnd.myformat']
 ```
 
+You will also need to declare an encoder which supports the new format:
+
+```yaml
+services:
+    app.api-platform.myformat.encoder:
+      class: ApiPlatform\Serializer\JsonEncoder
+      arguments:
+          $format: 'myformat'
+      # The following lines are only needed if autoconfigure is disabled
+      # tags:
+      #     - { name: 'serializer.encoder' }
+```
+
 API Platform will automatically call the serializer with your defined format name as `format` parameter during the deserialization process (`myformat` in the example).
 It will then return the result to the client with the requested MIME type using its built-in responder.
 For non-standard formats, [a vendor, vanity or unregistered MIME type should be used](https://en.wikipedia.org/wiki/Media_type#Vendor_tree).


core/content-negotiation.md
```diff
@@ -200,6 +200,19 @@ API Platform will automatically call the serializer with your defined format nam
 It will then return the result to the client with the requested MIME type using its built-in responder.
 For non-standard formats, [a vendor, vanity or unregistered MIME type should be used](https://en.wikipedia.org/wiki/Media_type#Vendor_tree).
 
+You will also need to declare an encoder which supports the new format:
```
alanpoulain, line 203: Maybe add it before?

nikophil: you mean before the previous paragraph?

alanpoulain: Yes.

nikophil: sorry, I totally forgotten about this! :sweat_smile: 

fixed!

core/content-negotiation.md
```diff
@@ -196,6 +196,19 @@ api_platform:
         myformat: ['application/vnd.myformat']
 ```
 
+You will also need to declare an encoder which supports the new format:
+
+```yaml
```
dunglas, line 201: Isn't it possible to use DI attributes instead?

core/content-negotiation.md
```diff
@@ -196,6 +196,19 @@ api_platform:
...
+services:
+    app.api-platform.myformat.encoder:
+      class: ApiPlatform\Serializer\JsonEncoder
+      argumments:
```
dunglas, line 205:
```suggestion
      arguments:
```

dunglas (changes requested):

core/content-negotiation.md
nikophil: :facepalm:

core/content-negotiation.md
nikophil: Not sure to understand how? we're suggesting here the user to declare a new service based on a class that comes from the vendors

dunglas: Ok I missed the namespace 😅. LGTM then.

dunglas (approved):

vinceAmstoutz:
Can I merge it? @soyuka @alanpoulain @dunglas

dunglas:
Can we create an encoder and register it using attributes instead of YAML config?

vinceAmstoutz:
> Can we create an encoder and register it using attributes instead of YAML config?

Yes @dunglas, can you please document this @nikophil?

nikophil:
This comment has already been made, but this cannot be done 😁
https://github.com/api-platform/docs/pull/1797#discussion_r1327951521

PR #1802, opened 2023-09-17
varnish is not provided by the default distribution anymore

dbu:
as discussed in https://github.com/api-platform/api-platform/issues/2133 varnish is not provided in the default setup, therefore the documentation is confusing.

on a quick search i did not find any instructions on how to set up the varnish side to work with the invalidation requests. i guess it would be found in old branches of api-platform that still contained varnish. if there is documentation somewhere, i'd be happy to link to it because as is, these instructions are not very user friendly when having to figure out completely on your own how to configure the caching proxy.


core/performance.md
```diff
@@ -20,11 +20,7 @@ cache. This ensures that the content served will always be fresh, because the ca
 most specific cases such as the invalidation of collections when a document is added or removed or for relationships and
 inverse relations is built-in.
 
-Integration with Varnish and Doctrine ORM is shipped with the core library, and [Varnish](https://varnish-cache.org/) is
```
dunglas, line 23: Maybe can we keep this part of the sentence "Integration with Varnish and Doctrine ORM is shipped with the Core library.", which is still true.

We may also mention that we now support Souin too.

dunglas (approved):

core/performance.md
dbu: ah, indeed, i deleted a bit too much. i added that line back.

i don't know souin, maybe better somebody who knows about it adds some information.

dbu:
@dunglas is the varnish side of the configuration documented somewhere? e.g. how to handle the tag invalidation requests?

soyuka:
thanks @dbu ! I guess the only docs is at https://github.com/api-platform/core/pull/5273/files#diff-5d7f4ac044cc872666abe48c4b0bfd771f274fbf8117db065fc8029eebd82841

I need to add this though.

PR #1805, opened 2023-09-24
Caddy proxy deployment

SiMoeBoe:
Adding a new deployment documentation for users, who want to run the api platform distrubution in a setup, that uses a caddy proxy in front of the api platform app.

This documentation was created during solving the problem of this discussion: https://github.com/api-platform/api-platform/discussions/2509


SiMoeBoe:
Forked from wrong branch

PR #1808, opened 2023-09-27
Document how to get introspection to work on any route

Jayfrown:
The previously documented solution would only trigger introspection on the `/` route. Users directly navigating to a resource without being authenticated (eg. navigating to `/#/books` and then being redirected to the authentication page) would not hit the `<Route path="/">`, and thus wouldn't trigger introspection after authenticating.

If `redirectToLogin` is `true`, the `<Resource>`es shouldn't be defined in the application because otherwise they will register the specific routes, e.g. `/#/books`, meaning the `<Route path="/:any">` wouldn't be selected, and introspection also would not be triggered.


soyuka:
thanks!

PR #1822, opened 2023-10-20
Update mentions of OpenAPI docs.json to docs.jsonopenapi

carlobeltrame:
Starting with 3.2, and since https://github.com/api-platform/core/pull/5808 and https://github.com/api-platform/api-platform/pull/2521, we cannot use /docs.json anymore


dunglas (approved):

dunglas:
Thanks!

PR #1829, opened 2023-11-03
fix: wrong parameter for docker composer up

### LINKED ISSUE #1828: Wrong parameter for docker compose up
On this page https://api-platform.com/docs/distribution/ we can see 

> Then, start Docker Compose in detached mode: `docker compose up --pull --wait` but `--pull` is not a parameter of this command

This line should be `docker compose up --wait` as wrote just under this line.

Hidoyatmz:
Close #1828


dunglas:
Good catch, thanks!

PR #1830, opened 2023-11-10
Update docs for new location of PathSegmentNameGenerator

paullallier:
This functionality has moved (possibly in 3.1, so that might be the right target?), and I ended up with a deprecation warning following the docs.


paullallier:
Sorry - I didn't realise renaming my branch would kill the PR.

PR #1844, opened 2023-11-23
docs(extra/releases): update Maintenance stable branch number

monitaure:
Based on the example in the latest versions of Api Platform _(and the example just above)_, the _"stable"_ version should be `3.2`.


soyuka:
thanks!

PR #1854, opened 2023-12-08
correct launch.json path mapping to /app

brettins:
Docs show `/srv/api` for `launch.json`, however the docker container now contains the app in `/app`.


dunglas:
Thanks!

PR #1862, opened 2023-12-23
Improves file upload testing documentation

cheike569:
This change will improve the File Upload Testing Documentation, as this has been hidden in the "File Upload" Section when it should be under the Testing section. 
It was not clearly pointed out how to only pass form fields with the `extra` option instead of the `body` (like one would assume) 
Also, it was not clearly pointed out that Symfony's `FormDataPart` Class won't work (like described at https://symfony.com/doc/current/http_client.html).


vinceAmstoutz:
Thank you so much for your PR and the time you invested in contributing to this project! We truly appreciate your effort and involvement.

PR #1866, opened 2023-12-30
Update subresources.md

meddouribadis:
Fixed a problem displaying a piece of code.

### DIFF
--- a/core/subresources.md
+++ b/core/subresources.md
@@ -108,6 +108,8 @@ Now to create a new way of retrieving an Answer we will declare another resource
 To make things work, API Platform needs information about how to retrieve the `Answer` belonging to
 the `Question`, this is done by configuring the `uriVariables`:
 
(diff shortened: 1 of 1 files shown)


core/subresources.md
```diff
@@ -108,7 +108,6 @@ Now to create a new way of retrieving an Answer we will declare another resource
 To make things work, API Platform needs information about how to retrieve the `Answer` belonging to
 the `Question`, this is done by configuring the `uriVariables`:
 
-<code-selector>
```
tvanhateren, line 111: The code-selector tag is used to create the buttons to switch between the PHP, XML, and YAML code.  
Instead of removing it we should add a new line after the tag. That fixes the rendering, and has a working code selector.

meddouribadis: Hey, thanks for your message. I made the proposed fixes.

PR #1869, opened 2024-01-09
Update filters.md

fouteox:
```php
abstract class AbstractFilter implements FilterInterface :

public function __construct(protected ManagerRegistry $managerRegistry, LoggerInterface $logger = null, protected ?array $properties = null, protected ?NameConverterInterface $nameConverter = null)
{
    $this->logger = $logger ?? new NullLogger();
}
```

The tild should not be in second position, it should be the log as shown in the AbstractFilter constructor


soyuka:
thanks

PR #1883, opened 2024-01-19
fix(doc) : Add slash to the documentation for fix swagger

SalvadorCardona:
If the / of auth is missing in Api Swager the Url should be localhostauth

If you add / the url should localhost/auth

Before : 

image

After :

image

### DIFF
--- a/core/jwt.md
+++ b/core/jwt.md
@@ -85,7 +85,7 @@ security:
             stateless: true
             provider: users
             json_login:
-                check_path: auth # The name in routes.yaml is enough for mapping
+                check_path: /auth # The name in routes.yaml is enough for mapping
                 username_path: email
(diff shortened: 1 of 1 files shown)


core/jwt.md
```diff
@@ -85,7 +85,7 @@ security:
             stateless: true
             provider: users
             json_login:
-                check_path: auth # The name in routes.yaml is enough for mapping
+                check_path: /auth # The name in routes.yaml is enough for mapping
```
soyuka, line 88: imo this is a name of a route declared inside symfony routing no?

SalvadorCardona: Maybe it's the problem that I updated my symfony to version 7 but the name of my route is auth

image

PR #1886, opened 2024-01-31
added missing hard return

dgoosens:
which caused the layout to be messed up

<https://api-platform.com/docs/core/subresources/>

image


soyuka:
thanks!

Issue #1780, opened 2023-07-12
[2.7] Broken internal links


alexislefebvre:
Here: https://api-platform.com/docs/v2.7/distribution/#plugging-the-persistence-system

If we try to open the “[data persisters](https://api-platform.com/docs/core/data-persisters)” link, we are redirected to the URL https://api-platform.com/docs/core/data-persisters that can't be found.

While 2.7 is not supported anymore, it would be nice to be able to still see its documentation.

PR #1892, opened 2024-02-08
fix(distribution): fix broken links for state providers and state processors

davy-beauzil:
#1780 

fix(distribution): fix broken links for state providers and state processors


vinceAmstoutz:
Not applicable anymore

PR #1896, opened 2024-02-13
Added missing line break after <code-selector> tag

Volyz:
Added a missing line break after the <code-selector> tag that resulted in a broken page


Volyz:
There is already another PR correcting the same problem

PR #1900, opened 2024-02-15
Update path configuration to show an exemple in case of multiple path

Simperfit:
The current wording is suggesting that if we add one path if should add to the current project, which is not the case

### DIFF
--- a/core/configuration.md
+++ b/core/configuration.md
@@ -92,8 +92,8 @@ api_platform:
             partial_parameter_name: 'partial'
 
     mapping:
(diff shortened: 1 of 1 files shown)


alanpoulain (approved):

vinceAmstoutz:
> I don't think it's the case, sorry: https://github.com/api-platform/core/blob/a8d0c8471060039bdd0b8d118dcf7f0e97709276/src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php#L385C50-L385C74

@Simperfit can you provide a reproducer of the issue you mention? Otherwise, we will close this PR as @alanpoulain explained

vinceAmstoutz:
Do you plan to keep going?

PR #1905, opened 2024-02-22
Use Symfony Response const

louismariegaborit:


louismariegaborit:
Can I homogenize the documentation by replacing const with direct value ?

soyuka:
Agree we don't want this patch or only for symfony docs but we don't have this yet. Thanks!

PR #1914, opened 2024-03-08
feat(performance): add Souin setup to the documentation

darkweak:
Add Souin setup in the performance documentation.


dunglas (approved):
LGTM! Thanks for improving this part of the docs.

PR #1912, opened 2024-02-28
typo on subresources.md

Altherius:
missing line break was causing formatting error on https://api-platform.com/docs/core/subresources/

PR #1916, opened 2024-03-11
Fix code example in subresource

Fan2Shrek:
While reading [subresources docs](https://api-platform.com/docs/v3.0/core/subresources/)

I noticed a weird '```php' so I tried to fix it :)


Fan2Shrek:
Closing due to #1912 my bad

PR #1917, opened 2024-03-12
Update getting-started.md

jjsmclaughlin:
Currently it is necessary to provide valid 'theme' and 'darkTheme' props to <HydraAdmin> or it fails with no useful error message.

- https://stackoverflow.com/questions/77840771/error-while-generate-an-admin-interface-using-api-platform-admin-and-react-admin

- https://github.com/marmelab/react-admin/issues/9687


jjsmclaughlin:
Fixed in 3.4.6.

PR #1919, opened 2024-03-14
fix: add newline for proper rendering of codeblock

darthf1:
Current:

https://api-platform.com/docs/core/subresources/

image


soyuka:
thanks!

PR #1930, opened 2024-03-29
Removed ResetDatabase trait

SebSept:
Because the tutorial make use of DAMADoctrineTestBundle to reset the database automatically before each test, the ResetDatabase trait is not needed (it may even cause trouble (I guess)).

### DIFF
--- a/distribution/testing.md
+++ b/distribution/testing.md
@@ -200,12 +200,11 @@ use ApiPlatform\Symfony\Bundle\Test\ApiTestCase;
 use App\Entity\Book;
 use App\Factory\BookFactory;
 use Zenstruck\Foundry\Test\Factories;
-use Zenstruck\Foundry\Test\ResetDatabase;
 
 class BooksTest extends ApiTestCase
 {
     // This trait provided by Foundry will take care of refreshing the database content to a known state before each test
-    use ResetDatabase, Factories;
+    use Factories;
 
     public function testGetCollection(): void
     {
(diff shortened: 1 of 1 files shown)


dunglas:
Shouldn't we remove DMADoctrineTestBundle instead for the sake of simplicity?

SebSept:
I have no real experience using that bundle and no idea of the speed improvements provided.

It's just 2 actions, a `composer req` command and adding a line in phpunit.xml.
So, not complicated. No more than a minute.

As mentioned on the [Symfony documentation](https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#database-reset), installing DamaDoctrineTestBundle improve the speed and should be encouraged.

So my humble opinion of new comer is to keep DamaDoctrineTestBundle.

dunglas:
Sounds reasonable! @kbond do you have an opinion about this?

kbond:
We've made sure this trait works seamlessly with dama. In fact, it's required for the following scenarios:
- cold run of your tests (database doesn't exist yet)
- using [global state](https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#global-state)
- debugging can be a bit easier since you can run a single test w/o dama easily

This isn't to say it's required but if there is a compatibility issue, I'd like to fix.

SebSept:
So DamaDoctrineTestBundle and Foundry ResetDatabase can look like a useless redundancy but they are not.
Just to mention, Dama ensure database remains untouched after a test runs, while Foundry's DatabaseReset() maker sur that the database is in a certain state.

So the trait should remain in the code snippet.

Maybe we could add a word in the doc to clarify things...

dunglas:
We can close then. Right?

SebSept:
Yes. 
Do you think it worth adding a word in these page ?

kbond:
> Just to mention, Dama ensure database remains untouched after a test runs, while Foundry's DatabaseReset() maker sur that the database is in a certain state.

Technically, dama does the same.

PR #1932, opened 2024-03-29
docs: parameters documentation

soyuka:


GwendolenLynch:
Thanks so much for your help lately, and for this `Parameter` work. 

This all reads well for me, with two comments.

It might be worth adding a note/example specifically relative to JSON:API's `include` parameter. For example, when testing the beta I was using this to have the parameter show up in Swagger UI, et al:
```php
#[GetCollection(
    parameters: ['include' => new QueryParameter(description: 'Included relationship paths, comma separated')],
)]
#[Get(
    parameters: ['include' => new QueryParameter(description: 'Included relationship paths, comma separated')],
)]
```

The other thing is with documentation the very long constructors are really hard to read. Can they be one property per-line to better communicate to the developer what they need to "pull in", or not?

soyuka:
@GwendolenLynch you should be able to create a parameter with default values and use it as attribute right? 

```
class IncludeParameter extends QueryParameter {
    public function __construct(...) {
        parent::__construct(description: 'my very long descr');
    }
}
```

GwendolenLynch:
Ha! I didn't even think of trying something like that yet. I am liking this new feature more and more than I already did!

PR #1933, opened 2024-03-29
Update nextjs.md : Document a peer dependency bug

Ak1r0:
Unfetch issue with his v5.0.0


soyuka:
We should remove unfetch alltogether as it's not needed anymore. You can maybe fix this on api-platform/api-platform, thanks!

PR #1939, opened 2024-04-24
docs(kubernetes) : Adding instruction for the messenger queue with Fr…

matthieu-delisle:
FrankenPHP image Kubernetes instruction incomplete

The readinessProbe and livenessProbe suggested in the documentation failed because the /bin/ps wasn't installed on the Docker Image.

### DIFF
--- a/deployment/kubernetes.md
+++ b/deployment/kubernetes.md
@@ -227,6 +227,8 @@ The `readinessProbe` and the `livenessProble` can not use the default `docker-he
 
 First, make sure to install the `/bin/ps` binary, otherwise the `readinessProbe` and `livenessProbe` will fail:
 
+<!-- markdownlint-disable no-hard-tabs -->
+
 ```patch
 # api/Dockerfile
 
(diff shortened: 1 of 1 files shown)


deployment/kubernetes.md
```diff
@@ -225,6 +225,8 @@ commandArgs: ['messenger:consume', 'async', '--memory-limit=100M']
 
 The `readinessProbe` and the `livenessProble` can not use the default `docker-healthcheck` but should test if the command is running.
 
+If you are using the FrankenPHP image make sure to install the `/bin/ps` binary (just add `procps` in the `apt-get install script`), otherwise the `readinessProbe` and `livenessProbe` will fail.
```
dunglas, line 228: This will be necessary as long as Debian (even without FrankenPHP). So I suggest the following:

```suggestion
Finally, make sure to install the `/bin/ps` binary), otherwise the `readinessProbe` and `livenessProbe` will fail:

```patch
# api/Dockerfile

RUN apt-get update && apt-get install --no-install-recommends -y \
+	procps \
```
```

dunglas (changes requested):
Good catch!

matthieu-delisle:
I just removed a ")" in your suggestion, I think everything is ok now.

dunglas:
Thank you very much @matthieu-delisle

PR #1940, opened 2024-04-29
fix: Use NormalizerInterface

maxhelias:
`ContextAwareNormalizerInterface` is deprecated since Sf 6.1 and remove in Sf 7


dunglas:
Thanks Max!!

PR #1942, opened 2024-04-30
Add enable_link_security

KDederichs:
Add missing enable_link_security config option


KDederichs:
Missing info added for #1692

KDederichs:
@soyuka Could you maybe merge/release this as well? Otherwise people will wonder why the Subresource security is not working.

PR #1943, opened 2024-05-01
Note requirement to explicitly enable multipart as an input format

paullallier:
V3.3 has become more strict about input format checking (intentionally or otherwise) and users need to be more rigorous in allowing multipart when using it for file uploads.

Prior to v3.3, I suspect people should have been defining this, but it wasn't being checked, so many won't have done.

Not sure my "note" format is in the preferred document style - possibly we don't want to mention the version for example?  Feel free to edit or request changes.

### DIFF
--- a/core/file-upload.md
+++ b/core/file-upload.md
@@ -79,6 +79,7 @@ use Vich\UploaderBundle\Mapping\Annotation as Vich;
         new Get(),
         new GetCollection(),
         new Post(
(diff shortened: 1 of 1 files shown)


paullallier:
I'm not sure if the linter is failing because I've malformed my PR or because there's a configuration issue with the CI process?

soyuka:
Thanks @paullallier not sure what's wrong with the linter

PR #1948, opened 2024-05-06
Update filters.md: place nested openapi properties on same level as o…

mikec655:
…ther properties

Place nested openapi properties on same level as other properties as nesting this properties has no effect. Only non-nested properties seem to work.

Place nested openapi properties on same level as other properties in getDescription() method on a custom filter as nesting this properties has no effect. Only non-nested properties seem to work.


soyuka:
this doesn't work on api platform

PR #1951, opened 2024-05-21
Update state-providers.md

gonzaloalonsod:
Return JsonResponse


gonzaloalonsod:
> This change doesn't look good to me.
> 
> The provider must return an object that will be handled by the serializer, not a HTTP response.

"Returns an empty DTO, does not serialize the object. Should it be explicit when serializing? Do you have an example?"

dunglas:
This shouldn't return an empty object. The serialization is done automatically, no need to do it explicitly. This example should work.

PR #1959, opened 2024-07-03
Update readinessProbe and livenessProbe in kubernetes.md

nightio:
The previous probe reported the container as ready during entrypoint.

During container start, the process (ps output) might looks like `{docker-entrypoi} /bin/sh /usr/local/bin/docker-entrypoint bin/console messenger:consume something`


dunglas (approved):

dunglas:
Thanks!

PR #1961, opened 2024-07-10
Fix path to SerializerWorkflow.png image

pierredup:
The image is broken all the way down to the [v2.1 docs](https://api-platform.com/docs/v2.1/core/serialization/), but not sure if it should be fixed all the way back to that branch (might make merging up to the latest version a bit difficult)


dunglas:
Thanks

PR #1963, opened 2024-07-26
doc: Add configuration doc for keep_legacy_inflector

jonnyeom:


jonnyeom:
Current ConfigurationTree line for reference
```php
    // src/Symfony/Bundle/DependencyInjection/Configuration.php
    ...

    ->booleanNode('keep_legacy_inflector')->defaultTrue()->info('Keep doctrine/inflector instead of symfony/string to generate plurals for routes.')->end()
```

dunglas (approved):

dunglas:
Thanks!

PR #1966, opened 2024-08-13
Fix: Reduce proselint warnings

pieterocp:
This addresses all the easily addressable proselint-able issues (on the basis that if there is a tool linting, getting it down to zero _or as close to_ makes any increase jump out). I've gone with a marked version over main as this is a bug fix? Not sure.

The three remaining are a bit on the fiddly side:

```console
./admin/index.md:5:87: garner.phrasal_adjectives.ly No hyphen is necessary in phrasal adjectives with an adverb ending in -ly, unless the -ly adverb is part of a longer phrase
./deployment/docker-compose.md:44:29: consistency.spacing Inconsistent spacing after period (1 vs. 2 spaces).
./core/messenger.md:111:61: garner.phrasal_adjectives.ly No hyphen is necessary in phrasal adjectives with an adverb ending in -ly, unless the -ly adverb is part of a longer phrase
```

First one is completely fine to the best of my knowledge, or the removal of `fully-featured` might work as an alternative?

Second is the whitespace in the dns section, I do not know enough about DNS records and their creation to know if the removal of a single whitespace to appease a prose linter would still make the guide still work, so I have not touched it.

Third is a link to symfony docs, where the it's the section title which would need "fixing", but this is a url, should a prose tool be reading URLs - maybe they [will stop validating URLs](https://github.com/amperser/proselint/issues/759) over this.


core/bootstrap.md
```diff
@@ -559,7 +559,7 @@ $list->insert($itemNormalizer, -895);
 // $list->insert($uuidDenormalizer, -895); //Todo ramsey uuid support ?
 $list->insert($openApiNormalizer, -780);
 
-// TODO: JSON-API support
+// Todo: JSON-API support
```
dunglas, line 562: Can you revert this one please? This syntax in "normal"

dunglas (changes requested):
Sorry for the late review. Could you also rebase please?

core/bootstrap.md
pieterocp: Happy to do so, and I wholly agree this is normal syntax, just proselint doesn't agree, removed the commit entirely + rebased.

pieterocp:
@dunglas I've rebased and made the changes you've suggested. A re-review would be appreciated.

dunglas:
Thanks

PR #1969, opened 2024-08-19
feat(performance): update Souin setup & enable Cache-Control public in docs

laurent-bientz:
Hello,

This pull request updates Souin setup in docs, see https://github.com/dunglas/frankenphp/issues/976#issuecomment-2296050761

### DIFF
--- a/core/performance.md
+++ b/core/performance.md
@@ -78,6 +78,22 @@ api_platform:
             urls: [ 'http://caddy/souin-api/souin' ]
             purger: api_platform.http_cache.purger.souin
 ```
+Don't forget to set your `Cache-Control` directive to enable caching on your `ApiResource`. This can be achieved using the `cacheHeaders` attributes:
+
+```php
+use ApiPlatform\Metadata\ApiResource;
+
+#[ApiResource(
+    cacheHeaders: [
+        'public' => true,
+        'max_age' => 60, 
+    ]
+)]
+class Book
+{
+    // ...
+}
+```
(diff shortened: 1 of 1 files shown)


core/performance.md
```diff
@@ -67,6 +75,8 @@ Setup the HTTP cache invalidation in your API Platform project
 ```yaml
 api_platform:
     http_cache:
+        # To make all responses public by default.
+        public: true
```
dunglas, line 79: This a dangerous default, this should be done on a per-operation basis instead, except for entirely public APIs. Could you add an example on a per-resource basis instead (like this: https://api-platform.com/docs/core/performance/#setting-custom-http-cache-headers)?

We should also remove this default from the "Varnish" snippet.

dunglas (changes requested):
Thank you very much for improving this, this is much appreciated! I left a comment regarding `public`, then we'll be ready to merge.

dunglas:
By the way, the config can now be simplified:

Before:


```caddyfile
{
    order cache before rewrite
    ...
    cache {
        api {
            souin
        }
    }
}
```

After:


```caddyfile
{
    cache
    # ...
}

# ...
```

core/performance.md
laurent-bientz: `public: true` removed (also in Varnish) + example added

laurent-bientz:
@dunglas I followed your recommendations, tell me if everything is ok

dunglas:
Thank you!!

PR #1976, opened 2024-09-10
Added XDEBUG_SESSION=1 to console instruction enabling xdebug

temple:
'Debug' concept is widely understood as the way xdebug step-debugging runs, so there is an explicit need to use XDEBUG_SESSION in order to enable step-debugging. Otherwise it seems not to work, cause providing `XDEBUG_MODE=debug` is not enough if using the api-platform distribution on its release v3.3.7 (which is the latest currently)


dunglas:
Thanks

PR #1985, opened 2024-09-19
Use latest Souin/cache-handler version in the documentation

darkweak:
Use latest Souin/cache-handler version


dunglas (approved):

dunglas:
Thanks Sylvain!

PR #2010, opened 2024-09-27
[Laravel > index] Remove failing assets part after api plaform install

vinceAmstoutz:
After installing Laravel via `laravel new` and adding API Platform by running `composer require api-platform/laravel`.

Everything works correctly except for the following command in this tutorial (Error on command run: “There is no command defined in the ‘api-platform’ namespace.")

image
image

Tell me if I'm wrong, but the asset will already be cleaned up automatically using the composer scripts and this is already done after each installation, for example after API Platform installation, as can be seen here :

image

That's why I'm proposing in this PR to delete this failed command.

### DIFF
--- a/laravel/index.md
+++ b/laravel/index.md
@@ -47,12 +47,6 @@ In your Laravel project, install the API Platform integration for Laravel:
 composer require api-platform/laravel
 ```
 
-After installing API Platform, publish its assets and config:
-
-```console
-php artisan api-platform:install
-```
-
 If it's not already done, start the built-in web server:
 
 ```console


vinceAmstoutz:
what to do with this part that also fails? https://github.com/api-platform/docs/blob/f4bcae2d4e027d53621420201f8a7654790f9c1e/laravel/index.md?plain=1#L66-L72

Replace with `php artisan vendor:publish --tag=laravel-assets --ansi --force` or remove?

dunglas:
I think it's because we need to tag a new release or core. cc @soyuka

vinceAmstoutz:
> I think it's because we need to tag a new release or core. cc @soyuka

@soyuka @dunglas Would you like me to create an issue for this?

dunglas:
Antoine published a new release that fixes the issue yesterday. Thanks!

PR #2011, opened 2024-09-27
Document spec-version option usage

Minetorpia:
This is currently an undocumented feature as fas as I know. I think it would make sense to include it in the documentation, because for some it might be very useful.


dunglas (approved):

dunglas:
Thanks!

PR #2017, opened 2024-09-30
patch(filters) fix openapi property in custom filter documentation

gabrielius914:
I noticed that starting from api-platform:4.0 the `openapi` property expects `ApiPlatform\OpenApi\Model\Parameter` object instead of the array in the filter description.


dunglas:
Thanks!

PR #2028, opened 2024-10-03
Change composer require package name

Roshyo:
If you try to require with `API` in uppercase, it returns this error
image

Whereas lowercase `api` it goes fine


dunglas (approved):

dunglas:
Good catch, thanks!

Issue #1837, opened 2023-11-17
Github actions lint should fail 


soyuka:
Our linter on github action doesn't fail and stop the job.

PR #2035, opened 2024-10-09
Bump super-linter from v4 to v7 and fix all files

vinceAmstoutz:
Solves #1837 by upgrading super-linter from v4 to v7.1.0, upgrade other CI images & fixing files


dunglas:
Thank you very much!!

PR #2049, opened 2024-10-28
Create LICENSE file

vinceAmstoutz:
I took the license file from the other API Platform repos.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,397 @@
+Licence Creative Commons Attribution 4.0 International (CC BY 4.0)
+
+Copyright (c) 2015-present [Kévin Dunglas]
+
+=======================================================================
+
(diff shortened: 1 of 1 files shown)


dunglas (changes requested):
Unlike the code, the docs are under [CC by 3.0](https://creativecommons.org/licenses/by/3.0/), not under MIT.
By the way, we could upgrade to CC 4.0. If we do so, we must update the license on the website too.

vinceAmstoutz:
> Unlike the code, the docs are under [CC by 3.0](https://creativecommons.org/licenses/by/3.0/), not under MIT. By the way, we could upgrade to CC 4.0. If we do so, we must update the license on the website too.

Here's what I've done with the CC-BY-4.0 template. If it's okay for you Kévin, I'll also make a PR for the website.

dunglas:
On for me! Thanks!!

PR #2051, opened 2024-10-28
fix: Dockerfile for custom FrankenPHP builds

Deuchnord:
See also: https://github.com/dunglas/frankenphp/pull/1121

### DIFF
--- a/core/performance.md
+++ b/core/performance.md
@@ -32,7 +32,7 @@ The integration using the cache handler is quite simple. You just have to update
 # Versions
 -FROM dunglas/frankenphp:1-php8.3 AS frankenphp_upstream
 
-+FROM dunglas/frankenphp:latest-builder AS builder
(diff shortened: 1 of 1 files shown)


core/performance.md
```diff
@@ -32,7 +32,7 @@ The integration using the cache handler is quite simple. You just have to update
 # Versions
 -FROM dunglas/frankenphp:1-php8.3 AS frankenphp_upstream
 
-+FROM dunglas/frankenphp:latest-builder AS builder
++FROM dunglas/frankenphp:latest AS builder
```
dunglas, line 35:
```suggestion
+FROM dunglas/frankenphp:builder-php8.3 AS builder
```

(A builder image is mandatory)

dunglas (changes requested):

dunglas (approved):

PR #2064, opened 2024-11-18
Fix MultipartDecoder to support more denormalization types

cedriclombardot:

PR #2065, opened 2024-11-19
fix(file-upload): always JSON encode field values

alanpoulain:
Alternative to https://github.com/api-platform/docs/pull/2064.

### DIFF
--- a/core/file-upload.md
+++ b/core/file-upload.md
@@ -331,6 +331,34 @@ The file and the resource fields will be posted to the resource endpoint.
 
 This example will use a custom `multipart/form-data` decoder to deserialize the resource instead of a custom controller.
 
+> [!WARNING]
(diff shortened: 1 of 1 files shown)


core/file-upload.md
```diff
@@ -331,6 +331,34 @@ The file and the resource fields will be posted to the resource endpoint.
 
 This example will use a custom `multipart/form-data` decoder to deserialize the resource instead of a custom controller.
 
+> [!WARNING]
+> Make sure to encode the fields in JSON before sending them.
+
+For instance, you could do something like this:
+```js
+async function uploadBook(file) {
```
cedriclombardot, line 339: This will not working with swagger interface

alanpoulain: Maybe, but the implementation is much simpler, isn't it?

PR #2094, opened 2024-12-06
docs(file-upload): move from core to symfony and update

vinceAmstoutz:
Documentation moved from `/core` to `/symfony`, updated and invalid links fixed


vinceAmstoutz:
Documentation moved from `core/` to `symfony`, confirmed by @dunglas via message

PR #2095, opened 2024-12-06
docs(jwt): support for laravel and reorganize

vinceAmstoutz:
- Adds Laravel documentation for JWT
- Move Symfony JWT documentation from `core/` to `symfony`


laravel/jwt.md
```diff
@@ -0,0 +1,115 @@
+# JWT Authentication with Laravel
+
+To implement JWT authentication in a Laravel application, the recommended package is [tymondesigns/jwt-auth](https://github.com/tymondesigns/jwt-auth),
```
dunglas, line 3: Maybe say for both Laravel and Symfony that we recommend using JWT-based standards such as OIDC instead of ad-hoc solutions like these?

laravel/jwt.md
```diff
@@ -0,0 +1,115 @@
...
+which provides a robust solution for handling JSON Web Tokens.
+
+For comprehensive details on authentication, refer to our [Laravel Authentication documentation](https://api-platform.com/docs/laravel/#authentication).
```
dunglas, line 6: Please use relative links everywhere. The website will do the conversion automatically.

laravel/jwt.md
vinceAmstoutz: Done

PR #2096, opened 2024-12-06
Replace const with property in state provider examples

GromNaN:
While trying the exact example, I found an [issue](https://3v4l.org/mStY0):

> New expressions are not supported in this context


dunglas (approved):

dunglas:
Good catch!

PR #2109, opened 2024-12-14
graphql.md: add "MediaObject Normalization" chapter

timble-one:
graphql-normalizer needs to be injected


vinceAmstoutz (commented):
@71m024 Tanks for your contribution!

Can you please add more explanation on the “Why”?

Also, can you add an injection example for Laravel? As you can see [here with an example of injection](https://api-platform.com/docs/core/graphql/#modify-the-extracted-types), we now document both Symfony and Laravel (since `4.0`).

timble-one:
I think the linting Problem is not caused by this PR.

vinceAmstoutz (changes requested):
Should be great for me after theses changes @timble-one

vinceAmstoutz (approved):

PR #2111, opened 2024-12-27
Keep original headers, so the PUT/POST operations aren't bloked

xrogers:
Without this, some headers could be dropped and POST/PUT operations will not work anymore.

### DIFF
--- a/admin/authentication-support.md
+++ b/admin/authentication-support.md
@@ -23,9 +23,16 @@ import { parseHydraDocumentation } from "@api-platform/api-doc-parser";
 import authProvider from "utils/authProvider";
 import { ENTRYPOINT } from "config/entrypoint";
 
-const getHeaders = () => localStorage.getItem("token") ? {
-  Authorization: `Bearer ${localStorage.getItem("token")}`,
-} : {};
+const getHeaders = (defaultHeaders = {}) => {
+    const authHeaders = localStorage.getItem("auth")
+        ? { Authorization: `Bearer ${JSON.parse(localStorage.getItem('auth'))['token']}` }
+        : {};
+
+    return {
+        ...defaultHeaders,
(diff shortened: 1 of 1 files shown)


xrogers:
This is not correct. The provided example breaks in the current version POST actions. Try it out. It's not related to any specific use case, just basic functionality.

J3m5:
Hello @xrogers, could you provide a reproduction, please?  

In its current state, I can't see how adding a default parameter can change the return value of the `getHeaders` function.  

The `options` object with the `headers` property, which is passed to the [`parseHydraDocumentation`](https://github.com/api-platform/api-doc-parser/blob/main/src/hydra/parseHydraDocumentation.ts#L195) function, has [this signature](https://github.com/api-platform/api-doc-parser/blob/main/src/hydra/types.ts#L2). As you can see, it doesn't specify any parameter, and nothing is passed to it [when it is called](https://github.com/api-platform/api-doc-parser/blob/3000d61c7fb48eacce67e9c533f318b92354b7e2/src/hydra/fetchJsonLd.ts#L59).

xrogers:
HI @J3m5 
I'm quite swamped to prepare the example, but the issue relates to code from this change: https://github.com/api-platform/admin/commit/66dca9c2fa5b4841c350f62e575a0f9ad7249a61#diff-a3ad30593f2f6c73e4e5b94fbdec633ffec199309aa092ebdf98ad4ed10a95bcR501 for example
As you can see, content type header is modified in case of `PATCH`. 
With the code from docs this header would be over-ridden/cleared.

xrogers:
Also see my last commit, somehow it is not visible in this PR: https://github.com/xrogers/docs/commit/30a97820f3431a7b366801787cbb87f80ad96431

PR #2120, opened 2025-01-08
Update index.md

fabienVernieres:
Fix ambiguous choices

### DIFF
--- a/symfony/index.md
+++ b/symfony/index.md
@@ -460,7 +460,7 @@ bin/console make:entity --api-resource
 ```
 
 Doctrine's [attributes](https://www.doctrine-project.org/projects/doctrine-orm/en/current/reference/attributes-reference.html) map these entities to tables in the database.
-Mapping through [attributes](https://www.doctrine-project.org/projects/doctrine-orm/en/current/reference/attributes-reference.html) is also supported, if you prefer those.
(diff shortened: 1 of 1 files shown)


symfony/index.md
```diff
@@ -460,7 +460,7 @@ bin/console make:entity --api-resource
 ```
 
 Doctrine's [attributes](https://www.doctrine-project.org/projects/doctrine-orm/en/current/reference/attributes-reference.html) map these entities to tables in the database.
-Mapping through [attributes](https://www.doctrine-project.org/projects/doctrine-orm/en/current/reference/attributes-reference.html) is also supported, if you prefer those.
+Mapping through [annotations](https://www.doctrine-project.org/projects/doctrine-annotations/en/2.0/index.html)  is still supported for backward compatibility, but they are considered deprecated and attributes are now the recommended approach.
```
vinceAmstoutz, line 463:
```suggestion
Mapping through [annotations](https://www.doctrine-project.org/projects/doctrine-annotations/en/current/index.html)  is still supported for backward compatibility, but they are considered deprecated and attributes are now the recommended approach.
```

vinceAmstoutz (changes requested):

symfony/index.md
fabienVernieres: Done ty

PR #2123, opened 2025-01-09
fix: path references in YAML config snippets

phansys:
AFAIK, the documentation is built around the paths from [`api-platform/api-platform`](https://github.com/api-platform/api-platform/tree/v2.6.0/api/config).

* Added missing `api/` directory references where missing;
* Replaced `app/` with `api/`.

### DIFF
--- a/admin/getting-started.md
+++ b/admin/getting-started.md
@@ -42,7 +42,7 @@ export default () => (
 Be sure to make your API send proper [CORS HTTP headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) to allow
 the admin's domain to access it.
 
-To do so, if you use the API Platform Distribution, update the value of the `CORS_ALLOW_ORIGIN` parameter in `api/.env` (it will be set to `^https?://localhost:?[0-9]*$`
+To do so, if you use the API Platform Distribution, update the value of the `CORS_ALLOW_ORIGIN` parameter in `.env` (it will be set to `^https?://localhost:?[0-9]*$`
 by default).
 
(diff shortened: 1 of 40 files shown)


alanpoulain (approved):
Seems OK to me, @dunglas do you agree?

soyuka:
`core` is the symfony-only version in a sense, I'd say that the path is fine without the `api` prefix.

phansys:
>core is the symfony-only version in a sense, I'd say that the path is fine without the api prefix.


I agree, but I also think we should make explicit clarifications in places where the distribution version is referenced, as there are some examples where the removal of the `api/` prefix may lead to misleading results.

I mean cases like these:

* https://github.com/api-platform/docs/blob/24dfb34e0d8af7bd56962d2bfb6b6ea2d520a6f9/deployment/heroku.md?plain=1#L24
* https://github.com/api-platform/docs/blob/24dfb34e0d8af7bd56962d2bfb6b6ea2d520a6f9/distribution/index.md?plain=1#L103
* https://github.com/api-platform/docs/blob/24dfb34e0d8af7bd56962d2bfb6b6ea2d520a6f9/distribution/index.md?plain=1#L205
* https://github.com/api-platform/docs/blob/24dfb34e0d8af7bd56962d2bfb6b6ea2d520a6f9/distribution/debugging.md?plain=1#L14
* https://github.com/api-platform/docs/blob/24dfb34e0d8af7bd56962d2bfb6b6ea2d520a6f9/extra/troubleshooting.md?plain=1#L53

IIUC, the only places where the `api/` prefix is right is where the documentation is covering the setup for the `api-platform/api-platform` distribution. Other examples for use with Symfony or Laravel SHOULD NOT use the `api/` prefix.

Do you agree? I'll be waiting for your feedback in order to confirm what updates to make.

PR #2135, opened 2025-02-11
Use property type

GromNaN:
I'm not sure if it was intentional or not to use untyped properties. Providing examples using the modern syntax is useful for the readers.


soyuka:
thanks @GromNaN !

PR #2136, opened 2025-02-18
docs(identifier): update identifier doc to use php attribute to tag service for uri transformer

asalisaf:

### DIFF
--- a/core/identifiers.md
+++ b/core/identifiers.md
@@ -164,6 +164,60 @@ services:
 
 </code-selector>
 
+You can also use `#[AutoconfigureTag('api_platform.uri_variables.transformer')]` to tag the service.
+
+```php
+<?php
+
+namespace App\Identifier;
+
+use ApiPlatform\Api\UriVariableTransformerInterface;
+use ApiPlatform\Exception\InvalidUriVariableException;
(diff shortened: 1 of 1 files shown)


core/identifiers.md
```diff
@@ -164,6 +164,60 @@ services:
...
+namespace App\Identifier;
+
+use ApiPlatform\Api\UriVariableTransformerInterface;
+use ApiPlatform\Exception\InvalidUriVariableException;
+use App\Uuid;
+use Symfony\Component\DependencyInjection\Attribute\AutoconfigureTag;
+use Ramsey\Uuid\Exception\InvalidUuidStringException;
+
+#[AutoconfigureTag('api_platform.uri_variables.transformer')]
+final class UuidUriVariableTransformer implements UriVariableTransformerInterface
```
soyuka, line 181: this interface isn't tagged as auto-configurable? Maybe that we should just add it as a bug fix...

soyuka: Indeed this should be automatic already: 
https://github.com/api-platform/core/blob/main/src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php#L168-L169

PR #2138, opened 2025-02-21
docs: update links to use Laravel section references

sneycampos:
This pull request add the laravel prefix in the laravel section since some anchor links are pointing to the root (symfony) documentation

(ps: i don't know if this fix the links since i think this is automatically generated to site using the gh workflow)

For example in this image, this link: https://api-platform.com/docs/#exposing-a-model should be `/docs/laravel#exposing-a-model`
image

### DIFF
--- a/laravel/index.md
+++ b/laravel/index.md
@@ -7,16 +7,16 @@ using Laravel!
 
 With API Platform, you can:
 
-- [expose your Eloquent](laravel#exposing-a-model) models in minutes as:
+- [expose your Eloquent](./#exposing-a-model) models in minutes as:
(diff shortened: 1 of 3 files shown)


sneycampos:
> Maybe it should be `./#`?

Maybe. Changed to `./#`

dunglas:
Shouldn't we fix the docs generator instead of the markdown (which looks good to me)?

vinceAmstoutz:
> Shouldn't we fix the docs generator instead of the markdown (which looks good to me)?

That's exactly what I wanted to say 😄

soyuka:
The problem is at https://github.com/api-platform/docs-website/blob/main/themes/api-platform/layouts/_default/_markup/render-link.html in this menu its working fine:

20250314_22h53m27s_grim

PR #2140, opened 2025-03-07
docs: add swagger_ui_extra_configuration.persistAuthorization

caspla:
Added persistAuthorization config option. 
I was looking for this feature in the docs, but couldn't find it. After i digged deeper, i found it in the swagger ui documentation:
https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/#authorization

I can imagine this is a often wanted feature, promoting it in the API Platform docs may help other developers to find it easier.


caspla:
i have no idea why the markdown check fails

PR #2147, opened 2025-03-12
docs(releases): update information releases

Saleen0x0f:

### DIFF
--- a/extra/releases.md
+++ b/extra/releases.md
@@ -9,17 +9,19 @@ For example:
 - version 3.1 has been released on 23 January 2023;
 - version 3.2 has been released on 12 October 2023;
 - version 3.3 has been released on 9 April 2024 (we were a little late, it should have been published in March);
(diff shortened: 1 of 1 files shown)


soyuka:
thanks @Saleen0x0f !

lazka:
This dropped security support for 3.4 (by removing the "2 minor branches" part which would be 4.0 and 3.4) and reduced the supported versions from 3 to 2, was that on purpose?

Just curious so I can update https://endoflife.date/api-platform :)

soyuka:
Actually I didn't noticed, and I'm not not supposed to maintain 3.4 anymore. Although for security patches I think it helps but there's no guarantee. We couldn't leverage the funds for an LTS version therefore we only support the last 2 minor versions.

lazka:
"not not supposed" or "not supposed" ;)

soyuka:
Then everything is fine :).

lazka:
I assume you meant "not supposed" then.

PR #2154, opened 2025-04-10
doc(errors): explain how to link an Error with its ApiResource

JacquesDurand:
This pull request aims to document the following pull request: https://github.com/api-platform/core/pull/6332


ytilotti:
Hi @JacquesDurand. Quick question, but was it intended that errors should overwrite the default errors?

Issue #2058, opened 2024-11-07
Issue with documentation in part with VichUploaderBundle


D3xime:
Hello everyone, 
i have seen an issue in ApiPlatform documentation about integration with VichUploaderBundle. The issue is when i follow the documentation step by step i got an error saying multipart can't be denormalize. The issue seems to appear in 3.3 to 4.0. I think, some part of the documentation like the creation of a dedicated controller are missing. 
```
<?php
// api/src/Entity/MediaObject.php

namespace App\Entity;

use ApiPlatform\Metadata\ApiProperty;
use ApiPlatform\Metadata\ApiResource;
use ApiPlatform\Metadata\Get;
use ApiPlatform\Metadata\GetCollection;
use ApiPlatform\Metadata\Post;
use ApiPlatform\OpenApi\Model;
use App\State\SaveMediaObject;
use Doctrine\ORM\Mapping as ORM;
use Symfony\Component\HttpFoundation\File\File;
use Symfony\Component\Serializer\Annotation\Groups;
use Symfony\Component\Validator\Constraints as Assert;
use Vich\UploaderBundle\Mapping\Annotation as Vich;

#[Vich\Uploadable]
#[ORM\Entity]
#[ApiResource(
    normalizationContext: ['groups' => ['media_object:read']],
    types: ['https://schema.org/MediaObject'],
    outputFormats: ['jsonld' => ['application/ld+json']],
    operations: [
        new Get(),
        new GetCollection(),
        new Post(
            inputFormats: ['multipart' => ['multipart/form-data']],
            openapi: new Model\Operation(
                requestBody: new Model\RequestBody(
                    content: new \ArrayObject([
                        'multipart/form-data' => [
                            'schema' => [
                                'type' => 'object',
                                'properties' => [
                                    'file' => [
                                        'type' => 'string',
                                        'format' => 'binary'
                                    ]
                                ]
                            ]
                        ]
                    ])
                )
            )
        )
    ]
)]
class MediaObject
{
    #[ORM\Id, ORM\Column, ORM\GeneratedValue]
    private ?int $id = null;

    #[ApiProperty(types: ['https://schema.org/contentUrl'], writable: false)]
    #[Groups(['media_object:read'])]
    public ?string $contentUrl = null;

    #[Vich\UploadableField(mapping: 'media_object', fileNameProperty: 'filePath')]
    #[Assert\NotNull]
    public ?File $file = null;

    #[ApiProperty(writable: false)]
[...]

PR #2156, opened 2025-04-14
Patched documentation problem with VichUploaderBundle

D3xime:
This PR is aimed to patch documentation problem in case with VichUploaderBundle

See issue for context : https://github.com/api-platform/docs/issues/2058 

fix: patch indentation for mutiples markdown Files. Change documentation for file upload to indicate well the way to use ApiPlatform with VIchUploaderBundle

### DIFF
--- a/symfony/file-upload.md
+++ b/symfony/file-upload.md
@@ -84,6 +84,8 @@ use Vich\UploaderBundle\Mapping\Annotation as Vich;
         new GetCollection(),
         new Post(
             inputFormats: ['multipart' => ['multipart/form-data']],
+            controller: CreateMediaObjectAction::class, 
+            deserialize: false, 
             openapi: new Model\Operation(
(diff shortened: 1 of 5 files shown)


symfony/file-upload.md
```diff
@@ -84,6 +84,8 @@ use Vich\UploaderBundle\Mapping\Annotation as Vich;
         new GetCollection(),
         new Post(
             inputFormats: ['multipart' => ['multipart/form-data']],
+            controller: CreateMediaObjectAction::class,
```
soyuka, line 87: we do not want to recommend using a controller

D3xime: What can we do to create a media with API platform then ? Nobody told me how to do it without using a controller. Is there a way to do this ?

D3xime: Noted. I will update the PR when I have time to do this. Thanks for the example.

PR #2158, opened 2025-04-29
Update Laravel Blade documentation view override location

jerodev:
The location of the blade files to override the default documentation template should be under `resources/views/vendor/api-platform`.


soyuka:
thanks!

PR #2160, opened 2025-05-06
Update file-upload.md

PaddyS:
Added part about serialization when using a file upload with the User entity, which gets serialized by Symfony for the session.

Ignoring this will result in an PHP exception when trying to upload a file to the User entity via the Symfony form builder.

To be fair, this is not exactly an API platform issue, since the upload via API works flawlessly even without this change.
I just happened to face this issue in this context, and found several other public posts about people using API platform facing the same issue.

It's totally fine if you don't want to have that part in your documentation though!


soyuka:
Indeed this looks like it should be in Symfony's documentation, what do you think?

PaddyS:
Yeah, unfortunately I did not find a fitting Symfony documentation that actually takes care of VichUploaderBundle, besides mentioning it in a sidenote.

I could imagine it here otherwise, in VichUploaderBundle's own documentation, maybe here:
https://github.com/dustin10/VichUploaderBundle/blob/master/docs/known_issues.md

PR #2161, opened 2025-05-07
Fix all occurrences of the old ApiPlatform/Api namespace

Dodger77:
The 4.1 documentation has some occurrences of the old `ApiPlatform/Api` namespace which may lead to problems for users implementing or upgrading API platform.


soyuka:
thanks!

PR #2167, opened 2025-06-08
docs(core/dtos): improve by adding other usages

vinceAmstoutz:
Completes #2076.

Also fixes a coding style issue introduced by this commit https://github.com/api-platform/docs/commit/e751b7e61e405f61585d1d4184191904476d5ddb.

### DIFF
--- a/core/dto.md
+++ b/core/dto.md
@@ -213,3 +213,29 @@ final class BookRepresentationProcessor implements ProcessorInterface
     }
 }
 ```
+
+## DTO without response
+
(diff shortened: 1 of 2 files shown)


vinceAmstoutz:
> I struggled to find "simple ways" to deal with DTOs in APIP a few months back, so any additional info about it is really welcomed. Last time I checked, I wanted to created routes with basic filtering/sorting combined with DTO and I was surprised Doctrine Collection Provider wasn't mentioned as well. Maybe something to add? Thanks!

That's what I've tried to explain @soyuka

PR #2174, opened 2025-06-19
Fix an error on pagination_client_enabled param in pagination.md

jrvoisin:

### DIFF
--- a/core/pagination.md
+++ b/core/pagination.md
@@ -203,7 +203,7 @@ To configure this feature globally, use the following configuration:
 # api/config/packages/api_platform.yaml
 api_platform:
   defaults:
-    pagination_client_enabled: false
+    pagination_client_enabled: true
   collection:
     pagination:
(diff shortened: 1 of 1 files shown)


vinceAmstoutz:
I disagree with your PR @jrvoisin. Disabling client-side paging globally requires `pagination_client_enabled: false`. The opposite will enable it and that's not what we want when we need to disable it.

jrvoisin:
I think there's a misunderstanding @vinceAmstoutz, because this part of documentation precisely shows how to configure API Platform to let the client enable or disable the pagination with the query parameter `pagination`.
If `pagination_client_enabled` is set to false (the default value), it will not be possible to enable or disable pagination from the client.
When we want to be able to use the query parameter as below, `pagination_client_enabled` as to be set to true : 

> The pagination can now be enabled or disabled by adding a query parameter named pagination:
> 
> GET /books?pagination=false: disabled
> GET /books?pagination=true: enabled

PR #2178, opened 2025-07-16
Update state-processors.md

brettins:
State processor clarification for graphql operations.


brettins:
The issue was simply declaring a mutation a DeleteMutation

Issue #2134, opened 2025-02-01
Documentation states that exposing a model to OpenAPI without any routes can use non-existing controllers


derTuca:
In [this](https://api-platform.com/docs/core/operations/#expose-a-model-without-any-routes) section, the following is written:
`This way, we expose a route that will do… nothing. Note that the controller does not even need to exist.`

However, if a non-existing controller is used, Symfony returns the following error:

```
Operation "_api_/entity_name{._format}_get" is defining an unknown service as controller "App\Entity\NonExistingControllerForOpenApi". Make sure it is properly registered in the dependency injection container in . (which is being imported from "/app/config/routes/api_platform.yaml"). Make sure there is a loader supporting the "api_platform" type.
```

Creating an empty AbstractController and referencing that fixes the issue.

soyuka:
Indeed it's an error in our documentation but as you don't need controllers with API Platform we should probably just remove that.

PR #2197, opened 2025-09-08
fix(operations): remove legacy doc about controllers

vinceAmstoutz:
Should fix https://github.com/api-platform/docs/issues/2134, as purposed [here](https://github.com/api-platform/docs/issues/2134#issuecomment-2652959105).


vincentchalamon (approved):
LGTM

PR #2201, opened 2025-09-10
docs: add https port overwrite example (#2200)

### LINKED ISSUE #2200: Alternate ports example is insufficient
> Be sure that the ports 80, 443, and 5432 of the host are not already in use. The usual offenders are Apache, NGINX, and Postgres. If they are running, stop them and run docker compose up --wait again.
> 
> Alternatively, run the following command to start the web server on port 8080 with HTTPS disabled:

https://api-platform.com/docs/symfony/#introduction:~:text=Be%20sure%20that,with%20HTTPS%20disabled%3A

Yes this does disable https, but port 443 is still used. The command will fail, if 443 is already in use.

I propose to also overwrite the https port

woldtwerk:
Fixes #2200 (alternative to #2189)

### DIFF
--- a/symfony/index.md
+++ b/symfony/index.md
@@ -91,7 +91,7 @@ docker compose up --wait
 > Alternatively, run the following command to start the web server on port `8080` with HTTPS disabled:
 >
 > ```console
-> SERVER_NAME=localhost:80 HTTP_PORT=8080 TRUSTED_HOSTS=localhost docker compose up --wait
+> SERVER_NAME=localhost:80 HTTP_PORT=8080 HTTPS_PORT=3001 HTTP3_PORT=3001 TRUSTED_HOSTS=localhost docker compose up --wait
 > ```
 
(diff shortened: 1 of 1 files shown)


symfony/index.md
```diff
@@ -91,7 +91,7 @@ docker compose up --wait
 > Alternatively, run the following command to start the web server on port `8080` with HTTPS disabled:
 >
 > ```console
-> SERVER_NAME=localhost:80 HTTP_PORT=8080 TRUSTED_HOSTS=localhost docker compose up --wait
+> SERVER_NAME=localhost:80 HTTP_PORT=8080 HTTPS_PORT=3001 HTTP3_PORT=3001 TRUSTED_HOSTS=localhost docker compose up --wait
```
soyuka: AFAIK changing SERVER_NAME should be sufficient, I think its too much informations, do you know what entity these env variables target?

PR #2204, opened 2025-09-13
Fix enable pagination client in API Platform configuration

pjehan:
The pagination_client_enabled should be true instead of false to enable client side pagination.

### DIFF
--- a/core/pagination.md
+++ b/core/pagination.md
@@ -203,7 +203,7 @@ To configure this feature globally, use the following configuration:
 # api/config/packages/api_platform.yaml
 api_platform:
   defaults:
-    pagination_client_enabled: false
+    pagination_client_enabled: true
   collection:
     pagination:
       enabled_parameter_name: pagination # optional


vinceAmstoutz:
@pjehan As you can see [here](https://github.com/pjehan/docs/blob/bb9afb452b98f714f05d47ce836b44928e42e26c/core/pagination.md?plain=1# L198-L206), the goal is to disable client-side pagination globally. That's why you'll find `pagination_client_enabled: false` (which disables it).

pjehan:
@vinceAmstoutz The documentation on this point is confusing.

According to the [configuration docs](https://api-platform.com/docs/core/configuration/), `pagination_client_enabled` defaults to `false`. This means clients cannot use `?pagination=false` to disable pagination, even though pagination itself is enabled by default.
Setting `pagination_client_enabled=true` allows clients to control pagination via the query parameter. I confirmed this behavior - when set to `true`, the pagination query option becomes available and clients can then disable pagination for specific requests.

The key distinction is:

- pagination_client_enabled=false: Pagination is always on, no client control
- pagination_client_enabled=true: Clients can choose to disable pagination with ?pagination=false

I just tried it and if the parameter is set to true, the pagination query option appear and then I can choose to actually disable the pagination for this endpoint:
image

PR #2211, opened 2025-09-19
docs(state): advertising Linked Data Platform headers Allow & Accept-Post

LaurentHuzard:
Add section about new header for Linked Data Platform compliance.


LaurentHuzard:
I added this documentation about headers for LDP, but I’m not sure if it’s in the right place.

PR #2215, opened 2025-09-30
Remove invalid tip about #[GetCollection] alias

rvanlaak:
Removed tip about #[GetCollection] alias in operations documentation, as #[Get] does not have a collection field anymore.


soyuka:
thanks :)

PR #2203, opened 2025-09-12
docs: add new search filters syntax and the new make filter command

vinceAmstoutz:
Adds documentation for https://github.com/api-platform/core/pull/7121, https://github.com/api-platform/core/pull/7364 and https://github.com/api-platform/core/pull/7355.

Also fixes https://github.com/api-platform/docs/issues/2202

PR #2227, opened 2025-11-04
docs(filters): document enhanced QueryParameter syntax on old filters

vinceAmstoutz:
Completes #2203

And here to avoid these issues: 
-  https://github.com/api-platform/core/issues/7483
- https://github.com/api-platform/core/issues/7493
- https://github.com/api-platform/core/issues/7495

Main changes:

- Added examples for Date, Boolean, Numeric, Range, Exists, and Order filters using the `QueryParameter` approach. Included notes on configuring custom strategies, handling null values, and enabling filters for nested properties.
- Add other warnings to promote the `QueryParameter` attribute over `ApiFilter`, as the latter is deprecated and scheduled for removal in API Platform 5.0. 
- Fix several typos

### DIFF
--- a/core/doctrine-filters.md
+++ b/core/doctrine-filters.md
@@ -6,8 +6,7 @@ For further documentation on filters (including for Eloquent and Elasticsearch),
 
 > [!WARNING]
 > For maximum flexibility and to ensure future compatibility, it is strongly recommended to configure your filters via
-> the parameters attribute using `QueryParameter`. The legacy method using the `ApiFilter` attribute is **deprecated** and
-> will be **removed** in version **5.0**.
+> the parameters attribute using `QueryParameter`. The legacy method using the `ApiFilter` attribute is not recommended.
 
 The modern way to declare filters is to associate them directly with an operation's parameters. This allows for more
 precise control over the exposed properties.
@@ -54,11 +53,6 @@ class Book {
 }
 ```
 
-**Further Reading**
-
-- Consult the documentation on [Per-Parameter Filters (Recommended Method)](../core/filters.md#2-per-parameter-filters-recommended).
-- If you are working with a legacy codebase, you can refer to the [documentation for the old syntax (deprecated)](../core/filters.md#1-legacy-filters-searchfilter-etc---not-recommended).
-
 ## Basic Knowledge
 
 Filters are services (see the section on [custom filters](../core/filters.md#creating-custom-filters)), and they can be linked
@@ -69,23 +63,6 @@ to a Resource in two ways:
 For example, having a filter service declaration in `services.yaml`:
 
 ```yaml
-# api/config/services.yaml
-services:
-  # ...
-  offer.date_filter:
-    parent: 'api_platform.doctrine.orm.date_filter'
-    arguments: [{ dateProperty: ~ }]
-    tags: ['api_platform.filter']
-    # The following are mandatory only if a _defaults section is defined with inverted values.
-    # You may want to isolate filters in a dedicated file to avoid adding the following lines.
-    autowire: false
-    autoconfigure: false
-    public: false
-```
-
-Alternatively, you can choose to use a dedicated file to gather filters:
-
-```yaml
 # api/config/filters.yaml
 services:
   offer.date_filter:
@@ -94,10 +71,13 @@ services:
     tags: ['api_platform.filter']
 ```
 
+> [!WARNING]
+> Its discouraged to use a filter with properties in the dependency injection as it may conflict with how
+> `QueryParameter` works. We recommend to use a per-parameter filter or to use the :property placeholder with a defined 
+> `filterContext` specifying your strategy for a given set of parameters.
+
 We're linking the filter `offer.date_filter` with the resource like this:
 
-<code-selector>
-
 ```php
 <?php
 // api/src/Entity/Offer.php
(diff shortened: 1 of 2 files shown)


ttskch:
@vinceAmstoutz 

Thank you for improving the documentation. Please allow me to share my insights.

In v4.2.3, when setting filters using `QueryParameter`, declaring the key using the associative array index in the `parameters` argument is not possible, resulting in a 500 error with the message "A Parameter should have a key." Instead, you must use the `key` argument of `QueryParameter`.

```diff
- parameters: [
-     'search[:property]' => new QueryParameter(/* ... */)
+ parameters: [
+     new QueryParameter(key: 'search[:property]', /* ... */)
```

Also, directly passing a filter instance to the `filter` argument still mostly does not work correctly. Defining the filter service in `services.yaml` and passing the service ID works correctly in most cases (excluding #7493, #7494, and #7495).

For example, I have observed the following issues in my environment.

> `new QueryParameter(key: 'date', filter: new DateFilter())`

- `date[<after|before|strictly_after|strictly_before>]` is not output in OpenAPI or Hydra's `search.template`, only `date` is output.
- Searching works.

> `new QueryParameter(key: 'id', filter: new RangeFilter())`

- `id[<lt|gt|lte|gte|between>]` is not output in OpenAPI or Hydra's `search.template`, only `id` is output.
- Searching works.

> `new QueryParameter(key: ':property', filter: new PartialSearchFilter(), properties: ['title', 'note'])`

- The query parameter `:property` is output in OpenAPI instead of `title` or `note`.
- `{?title,note}` is correctly output in Hydra's `search.template`.
- Searching does not work even when requesting with `?title={query}` or `?note={query}`.

> `new QueryParameter(key: 'title', filter: new PartialSearchFilter(), property: 'title')`

- Both OpenAPI and Hydra output correctly, and searching works.
- However, in the case of `BooleanFilter`, `NumericFilter`, `RangeFilter`, etc., if the `key` is the property name itself, it works even if the `property` argument is omitted, but in the above case, omitting `property` causes an error because `$property` becomes `null` [here](https://github.com/api-platform/core/blob/v4.2.3/src/Doctrine/Orm/Filter/PartialSearchFilter.php).

> `new QueryParameter(key: 'comments.content', filter: new PartialSearchFilter(), property: 'comments.content')`

- Nested properties are not parsed, and an error such as `Error: Class App\\Entity\\Foo has no field or association named comments.content` occurs at the DB layer.

> `new QueryParameter(key: 'exists[:property]', filter: new ExistsFilter(), properties: ['content', 'comments'])`

- In OpenAPI... - The query parameter `exists[:property]` is output instead of `exists[content]` or `exists[comments]`.
- Hydra's `search.template` correctly outputs `{?exists[content],exists[comments]}`.
- Searching does not work even when requesting with `?exists[content]=true` or `?exists[comments]=false`.
- Defining the following two parameters instead makes it work completely:

```php
new QueryParameter(key: 'exists[content]', filter: new ExistsFilter(), property: 'content'),
new QueryParameter(key: 'exists[comments]', filter: new ExistsFilter(), property: 'comments'),
```

> `new QueryParameter(key: 'order[:property]', filter: new OrderFilter(), properties: ['id', 'date'])`

- The query parameter `order[:property]` is output in OpenAPI instead of `order[id]` or `order[date]`.
- Hydra's `search.template` correctly outputs `{?order[id],order[date]}`.
- Searching does not work even when requesting with `?order[id]=asc` or `?order[date]=desc`.
- Defining the following two parameters instead makes it work completely:

```php
new QueryParameter(key: 'order[id]', filter: new OrderFilter(), property: 'id'),
new QueryParameter(key: 'order[date]', filter: new OrderFilter(), property: 'date'),
```

core/doctrine-filters.md
```diff
@@ -21,7 +21,7 @@ Here is the recommended approach to apply a `PartialSearchFilter` only to the ti
 #[ApiResource(operations: [
     new GetCollection(
         parameters: [
-            // This WILL restrict to only title and author properties
+            // This WILL restricts to only title and author properties
```
aegypius:
```suggestion
            // This WILL restrict to only title and author properties
```

restrict is a verb in this context

core/doctrine-filters.md
```diff
@@ -807,7 +1008,7 @@ services:
     arguments: [{ transportFees: ~ }]
     tags: ['api_platform.filter']
     # The following are mandatory only if a _defaults section is defined with inverted values.
-    # You may want to isolate filters in a dedicated file to avoid adding the following lines (by adding them in the defaults section)
+    # You may want to isolate filters in a dedicated file to avoid adding the following lines (by adding them in the default section)
```
aegypius:
```suggestion
    # You may want to isolate filters in a dedicated file to avoid adding the following lines (by adding them in the "defaults" section)
```

core/doctrine-filters.md
```diff
@@ -875,13 +1132,13 @@ services:
       $orderParameterName: order
     tags: ['api_platform.filter']
     # The following are mandatory only if a _defaults section is defined with inverted values.
-    # You may want to isolate filters in a dedicated file to avoid adding the following lines (by adding them in the defaults section)
+    # You may want to isolate filters in a dedicated file to avoid adding the following lines (by adding them in the default section)
```
aegypius:
```suggestion
    # You may want to isolate filters in a dedicated file to avoid adding the following lines (by adding them in the "defaults" section)
```

core/doctrine-filters.md
```diff
@@ -987,13 +1349,13 @@ services:
       ]
     tags: ['api_platform.filter']
     # The following are mandatory only if a _defaults section is defined with inverted values.
-    # You may want to isolate filters in a dedicated file to avoid adding the following lines (by adding them in the defaults section)
+    # You may want to isolate filters in a dedicated file to avoid adding the following lines (by adding them in the default section)
```
aegypius:
```suggestion
    # You may want to isolate filters in a dedicated file to avoid adding the following lines (by adding them in the "defaults" section)
```

core/doctrine-filters.md
```diff
@@ -1057,7 +1428,7 @@ services:
     arguments: [{ product.releaseDate: ~ }]
     tags: ['api_platform.filter']
     # The following are mandatory only if a _defaults section is defined with inverted values.
-    # You may want to isolate filters in a dedicated file to avoid adding the following lines (by adding them in the defaults section)
+    # You may want to isolate filters in a dedicated file to avoid adding the following lines (by adding them in the default section)
```
aegypius:
```suggestion
    # You may want to isolate filters in a dedicated file to avoid adding the following lines (by adding them in the "defaults" section)
```

core/doctrine-filters.md
```diff
@@ -615,12 +698,40 @@ Syntax: `?property=<true|false|1|0>`
...
+class Offer
+{
+    // ...
+}
+```
+
+> [!TIP]
+> For other syntaxes, for e.g., if you want to new syntax with the ApiResource attribute take a look [here](#introduction).
+
+### Boolean Filter using the ApiFilter Attribute Syntax (not recommended)
```
soyuka:
```suggestion
### Boolean Filter using the ApiFilter Attribute Syntax (using the DateFilter service)
```

soyuka:
Many thanks for the work on this! I'll fix linting issues in another PR!

PR #2262, opened 2026-03-08
Document property_info not loaded error

lacatoire:
Fixes https://github.com/api-platform/docs/issues/2249

Adds a troubleshooting section for the `property_info` service not found error that occurs when upgrading Symfony.


soyuka:
mhh but this probably means we need to fix something

PR #2272, opened 2026-03-25
Update Kubernetes deployment guide to match current Helm chart

### LINKED ISSUE #1450: Kubernetes deployment guide mentions nginx instead of caddy
As per title, in docs/deployment/kubernetes.md the documentation speaks of nginx but the helm charts are all focusing on caddy, could that be a mistake?

lacatoire:
Fixes #1450

Update the Kubernetes deployment documentation to align with the current Helm chart structure (FrankenPHP/Caddy, bitnami postgresql auth, correct value paths).

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -11,4 +11,3 @@ MD004:
   style: dash
 MD007:
   indent: 4
-MD030: false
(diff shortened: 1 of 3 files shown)


.markdownlint.yaml
```diff
@@ -11,3 +11,4 @@ MD004:
   style: dash
 MD007:
   indent: 4
+MD030: false
```
soyuka:
```suggestion
```

PR #2275, opened 2026-03-29
Review dto docs

soyuka:

### DIFF
--- a/core/dto.md
+++ b/core/dto.md
@@ -123,8 +123,9 @@ final class Book
 
 ### Implementation Details: The Object Mapper Magic
 
-Automated mapping relies on two internal classes: `ApiPlatform\State\Provider\ObjectMapperProvider`
-and `ApiPlatform\State\Processor\ObjectMapperProcessor`.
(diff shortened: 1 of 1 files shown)


core/dto.md
```diff
@@ -260,6 +329,21 @@ In your Book resource, configure the operations to use these classes via input a
...
+### Recommendations on input and output
+
+**Declare your operations on the Resource class** — it represents the JSON contract of your API.
+Avoid using `output:` on write operations (`Post`, `Put`, `Patch`, `Delete`). The
+`ObjectMapperOutputProcessor` already maps the persisted Entity back to the Resource class
+automatically. Adding an explicit `output:` on writes creates confusion and can lead to subtle
+bugs (see [#7745](https://github.com/api-platform/core/issues/7745)).
+
+The main legitimate use case for `output:` is on `GetCollection` when you need a lighter
+representation with fewer fields than the main Resource. Even then, consider whether serialization
```
Cafeine42, line 341: Wouldn’t that be problematic with JSON streaming, since it’s not compatible with serialization groups?

From my perspective, one of the main benefits of using DTOs is precisely to explicitly define and control the exposed properties, without relying on serialization groups.

PR #2276, opened 2026-03-30
feat: extract ApiTestCase in its own api-platform/test package

Maxcastel:
https://github.com/api-platform/core/pull/7887


soyuka:
thanks!

PR #2289, opened 2026-06-06
docs(serialization): document element-type requirement for readableLink on collection getters

soyuka:
## Why

When `readableLink: false` is applied to a getter returning a bare `array`, Symfony PropertyInfo resolves the type as `array<mixed>`. Without a known element class, API Platform cannot identify the items as API resources and falls back to embedding them — the opposite of the intended behavior. This is a configuration issue, not a framework bug (see api-platform/core#8179).

## What this PR adds

A new **"Collection Getters and readableLink"** subsection under the "Force IRI with relations of the same type" section in `core/serialization.md`, documenting three remedies:

- **PHPDoc `@return Car[]`** — zero-dependency, works with all API Platform versions
- **`ApiProperty(nativeType: Type::list(Type::object(Car::class)))`** — explicit type via `symfony/type-info` (API Platform >= 4.2)
- **Typed Doctrine collection** (`Collection<int, Car>`) — return a typed `Collection` instead of a plain `array`

## Related

- Core issue: api-platform/core#8179
- A companion fix PR on branch `fix/readable-link-array-collection-8179-v3` will be opened against `api-platform/core` separately.

### DIFF
--- a/core/openapi.md
+++ b/core/openapi.md
@@ -924,6 +924,32 @@ return [
 > **must** be set according to the
 > [OpenID Connect specification](https://openid.net/specs/openid-connect-core-1_0.html).
 
+## Sending Credentials with Swagger UI Requests
+
+When your API is deployed behind a proxy that uses cookie-based authentication (e.g. Cloudflare
+Access), Swagger UI's requests may be rejected because the authentication cookie is not forwarded by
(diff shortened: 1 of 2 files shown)


soyuka:
Closing — overkill for what is ultimately a typing requirement. The closed-issue thread on api-platform/core#8179 carries the explanation.

PR #2295, opened 2026-06-11
docs: document running a processor on read operations (write: true)

soyuka:
## Gap

Declaring a `processor:` on a `GetCollection` (or `GET`) operation silently does nothing — the processor is never invoked. This trips users up (see api-platform/core#8089) and is currently undocumented.

## Why

`WriteProcessor` only dispatches to a custom processor when the operation's `write` flag is truthy. For safe HTTP methods (`GET`/`GetCollection`) `write` defaults to `false`, so the processor is skipped. Setting `write: true` opts the operation in.

## What

New section in `core/state-processors.md`: **Running a Processor on Read (GET) Operations**.

- Explains the `write` flag default and `WriteProcessor` dispatch.
- Shows PHP attribute + YAML examples (the issue used YAML).
- CQRS framing: `processor` = command/query handler, `provider` = read/query side — with `write: true` a processor serves as the handler for a read operation, completing the CQRS picture already referenced at the top of the page.

Closes api-platform/core#8089

Note: `canWrite()` / the `write` flag is semantically "run the processor", not strictly "mutate state" — worth a rename later, but documenting the current behavior first.

### DIFF
--- a/core/state-processors.md
+++ b/core/state-processors.md
@@ -155,6 +155,80 @@ use App\State\BlogPostProcessor;
 class BlogPost {}
 ```
 
+## Running a Processor on Read (GET) Operations
+
+By default, a custom `processor` only runs on operations that mutate state: `POST`, `PUT`, `PATCH`,
+and `DELETE`. The internal `WriteProcessor` is responsible for dispatching to your processor, and it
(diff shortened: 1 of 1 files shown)


soyuka:
we eventually need to rename or alias `write` to `process`

PR #2320, opened 2026-08-27
Update parameter security documentation to reflect that it throws AccessDeniedException

leightonthomas:
In practice the `security` expression seems to `403` when returning `false` due to the logic in [`ApiPlatform\State\Provider\SecurityParameterProvider`](https://github.com/api-platform/core/blob/4.3/src/State/Provider/SecurityParameterProvider.php#L114), rather than ignoring the parameter as currently stated in the docs


soyuka:
thanks

PR #2321, opened 2026-09-02
docs(dto): clarify object mapper output on write operations

soyuka:
## Summary

`core/dto.md` implied that an `output` class configured on a write operation is used as the
Object Mapper target after persistence. It is not: `ObjectMapperOutputProcessor` always maps the
persisted entity back to the resource class. This is intentional, a resource instance is required
downstream for IRI generation and the `Location`/`Content-Location` headers, and the same change
was declined in api-platform/core#7611, api-platform/core#7940 and api-platform/core#8420.

This PR:

- Names the 4.3 classes (`ObjectMapperInputProcessor`, `ObjectMapperOutputProcessor`) in
  "How it works internally" and states that write mapping always targets the resource class.
- Adds a note after the input/output wiring example: `output` with the Object Mapper applies to
  `Get` and `GetCollection`.
- Adds "Returning a Different Representation From a Write Operation": a second `#[ApiResource]`
  class sharing the entity through `stateOptions`, `input:` for the request body, and
  `itemUriTemplate` on `Post` to reuse the main resource's item IRI.

Refs api-platform/core#7940, api-platform/core#8420.

## Checks

- [x] `python3 -m proselint check --config .proselintrc.json core/dto.md`
- [x] `npx markdownlint-cli2 core/dto.md`
- [x] `npx prettier --check core/dto.md --prose-wrap always`

### DIFF
--- a/core/dto.md
+++ b/core/dto.md
@@ -123,8 +123,10 @@ final class Book
 
 ### Implementation Details: The Object Mapper Magic
 
-Automated mapping relies on two internal classes: `ApiPlatform\State\Provider\ObjectMapperProvider`
-and `ApiPlatform\State\Processor\ObjectMapperProcessor`.
+Automated mapping relies on three internal classes:
+`ApiPlatform\State\Provider\ObjectMapperProvider`,
+`ApiPlatform\State\Processor\ObjectMapperInputProcessor`, and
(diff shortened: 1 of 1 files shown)


dylan-rumble:
Before merging this, please take a look at: https://github.com/api-platform/core/pull/8497

